PR #2, opened 2020-12-31
Quick fix for checking if the function 'WebDevIconsGetFileTypeSymbol' exists

jarviliam:
When I installed this plugin without any Webdevicons plugins already installed, I received an error on opening any file due to the WebDevIcons function not existing.
 It turns out that the return from checking if the 'WebDevIconsGetFileTypeSymbol'  function exists returns a 0 if it does not exist, so I just updated accordingly.


hoob3rt:
Nice catch, thanks!

PR #3, opened 2020-12-31
PoC to resolve #1

### LINKED ISSUE #1: Scrolling using vim-smoothie lags when lualine.nvim is enabled
* `nvim --version`: 

```
NVIM v0.5.0-dev+972-g84faeb07d
Build type: RelWithDebInfo
LuaJIT 2.0.5
```
* latest versions of both plugins
* terminal: `kitty 0.19.3` (also occurs in xterm)

Scrolling using vim-smoothie while lualine.nvim is enabled produces lag, especially towards the end of the scrolling motion.  This does not occur while using vim-airline (though lualine.nvim is the first statusline plugin I've ever used, but I had to test with another one).

eric-hansen:
General proof of concept to demonstrate using libuv/vim.loop.  The issue isn't so much being exec'ed on each refresh but that systemlist is kind of performance intensive.  Especially when you only need 1 single bit of information (the branch name in this case).

Git 2.22+ supports `git branch --show-current` where as if this fails it falls back to the pre-2.22 method which works as far back as 1.7 at least.

This is untested (I do not use Lualine currently) but hope this can at least help guide in a proper direction.

### DIFF
--- a/lua/lualine/components/branch.lua
+++ b/lua/lualine/components/branch.lua
@@ -1,11 +1,19 @@
+local utils = require 'lualine.utils'
+
 local function Branch()
-  local branch = vim.fn.systemlist(
-  'cd '..vim.fn.expand('%:p:h:S')..' 2>/dev/null && git status --porcelain -b 2>/dev/null')[1]
+  local branch
+
+	local cb = function (err, data)
+		if err == 0 and data then branch = data:match "^%*(.-)%s*$" end
+	end
+
+	utils.asyncCall('git', {'branch', '--show-current'}, cb)
+
   if not branch or #branch == 0 then
-    return ''
+		utils.asyncCall('git', {'rev-parse', '--abbrev-ref', 'HEAD'}, cb)
+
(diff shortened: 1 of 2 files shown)


hoob3rt:
Thank you for that! I will review it soon and check if it solves the issue

eric-hansen:
Let me know your results.  I plan on installing lualine soon myself just haven't sat down and got things set up yet.

hoob3rt:
I implemented async here https://github.com/hoob3rt/lualine.nvim/tree/async 
Let me know what you think

eric-hansen:
I would error check the git command so if an error happens you can fallback to the earlier way that's in my pr

hoob3rt:
> I would error check the git command so if an error happens you can fallback to the earlier way that's in my pr

That would be nice but git `2.22` is the oldest maintained version of git. If git developers don't want to maintain older versions then neither do I.

On top of that `git rev-parse --abbrev-ref HEAD` breaks on git repos without remotes.

PR #5, opened 2020-12-31
Modified buffer component

samirettali:
A really simple component that shows a `+` if the buffer is modified. I was using it in my old statusline, hope it's useful.

### DIFF
--- a/lua/lualine/components/modified.lua
+++ b/lua/lualine/components/modified.lua
@@ -1,0 +1,9 @@
+local function Modified()
+   if (vim.bo.modified) then
+      return "+"
+   else
+      return ""
+   end
+end
+
+return Modified


phenax:
Can we make that character configurable? I'd like to have a more prominent character there.

hoob3rt:
Is this not alredy handled in `filaname.lua` with `%m` ? I agree this should be a separate component but your solution is very specific and doesn't handle readonly files etc

samirettali:
Hey, sorry for the late response, I was a little busy. I was not using the `filename` so I didn't notice it, what would be a good decision for you?
I think we can modify the `filename` component to only show the filename and add a `modified` component and a `readonly` component maybe?
We could add two default values `modified` and `readonly` components and two options to change the string they show.

hoob3rt:
> Hey, sorry for the late response, I was a little busy. I was not using the `filename` component so I didn't notice it, what would be a good decision for you?
> I think we can modify the `filename` component to only show the filename and add a `modified` component and a `readonly` component maybe?

I think `modified` and `readonly` components should be merged into one to avoid config confusion/hell. This could be done like this as an early poc
```lua
local function test()
  local data = [[%m %r]]
  return data
end

return test
```
But a better solution would be to use `vim.bo.modified` and `vim.bo.readonly` to get state.

> We could add two default values `modified` and `readonly` components and two options to change the string they show.

This cannot be currently done. I will be working on a universal way of passing arguments/options to components soon.

hoob3rt:
this is resolved with #60

PR #6, opened 2020-12-31
highlight theme if table

mhartington:
Moves the `highlight.createHighlightGroups` for a theme so that inline custom themes can be parsed as well. Otherwise the readme is incorrect

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -64,8 +64,8 @@ local function StatusLine(isFocused)
   end
   if type(M.theme) == 'string' then
     M.theme = utils.setTheme(M.theme)
-    highlight.createHighlightGroups(M.theme)
   end
+  highlight.createHighlightGroups(M.theme)
   local status = ''
   if sections.lualine_a ~= nil then
     status = status .. highlight.formatHighlight(isFocused, 'lualine_a')
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -65,7 +65,6 @@ local function StatusLine(isFocused)
     sections = M.inactiveSections
   end
   if not theme_set  then
-    print('only called once')
     if type(M.theme) == 'string' then
       M.theme = utils.setTheme(M.theme)
     end
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -17,6 +17,8 @@ M.sections = {
   lualine_diagnostics = {  }
 }
 
+local theme_set = false
+
 M.inactiveSections = {
   lualine_a = {  },
   lualine_b = {  },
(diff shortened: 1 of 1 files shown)


lua/lualine.lua
```diff
@@ -64,8 +64,8 @@ local function StatusLine(isFocused)
   end
   if type(M.theme) == 'string' then
     M.theme = utils.setTheme(M.theme)
-    highlight.createHighlightGroups(M.theme)
   end
+  highlight.createHighlightGroups(M.theme)
```
hoob3rt, line 68: This is a really cheap fix to this problem. This `M.Statusline` runs every time a change happens in the statusline. Your solution makes lualine run all highlight commands run on every change. A better solution might be to copy the whole if block to the `M.status()` function, see if that resolves this issue.

mhartington: Tried that, but the highlights will get called before the sections are drawn. Though, M.Statusline only appears to be called on Win/Buf Enter/Leave, so it's doesn't seem overkill to have the highlights applied then.

hoob3rt: Please check if https://github.com/hoob3rt/lualine.nvim/tree/theme-highlight-fix resolves this issue.

mhartington: This still would not resolve things if the theme is not a module in this repo.  If a third party theme is passed, it would not be highlighted.

hoob3rt: Oh you are right, but I think a better solution is needed than just spamming a highlight all the, let me know if you come up with something, I will try to tackle this tomorrow.

mhartington:
Used some local state to track the theme changes and only apply the highlighting once.

alaric:
> Used some local state to track the theme changes and only apply the highlighting once.

This approach does prevent you changing the theme live in an instance, which is quite nice to try them out now there's multiple (or if you're defining your own). Maybe there could be a specific function to both set `M.theme` and reset `theme_set`?

lua/lualine.lua
```diff
@@ -62,10 +64,14 @@ local function StatusLine(isFocused)
   if not isFocused then
     sections = M.inactiveSections
   end
-  if type(M.theme) == 'string' then
-    M.theme = utils.setTheme(M.theme)
+  if not theme_set  then
```
hoob3rt, line 67: A solution for fixing this bug while keeping functionality of live theme changing might be something like this
```lua
local theme_set = {}
M.theme = 'gruvbox'
--
--
if theme_set ~= M.theme then
  if type(M.theme) == 'string' then
        M.theme = utils.setTheme(M.theme)
   end
   highlight.createHighlightGroups(M.theme)
   theme_set = M.theme
```
Let me know what you think about this solution.

hoob3rt:
This is resolved with https://github.com/hoob3rt/lualine.nvim/commit/3ab706671898746c268f59ca7f02a4919072690f

Have fun setting your colorscheme!

PR #7, opened 2020-12-31
Use table.concat instead of string concatentation

elihunter173:
Per your request on [Reddit](https://www.reddit.com/r/neovim/comments/kn5wo4/lualinenvim_a_blazing_fast_and_easy_to_configure/ghj5ajp?utm_source=share&utm_medium=web2x&context=3) :)

I also went through and switched other similar cases to use `table.concat`. There were 3, listed below. I am not 100% confident that I translated `drawSection()` correctly.
* `StatusLine()` in `lualine.lua`
* `highlight()` in `lualine/highlight.lua`
* `drawSection()` in `lualine/utils.lua`

### DIFF
--- a/lua/lualine/utils.lua
+++ b/lua/lualine/utils.lua
@@ -5,29 +5,21 @@ function M.setTheme(theme)
 end
 
 function M.drawSection(section, separator)
-  local status = ''
-  for index, statusFunction in pairs(section) do
+  local status = {}
+  for _, statusFunction in pairs(section) do
     local localstatus = statusFunction()
-    if localstatus:len() > 0 then
-      if separator:len() > 0 then
-        if index > 1 then
-          status = status .. separator .. ' '
-        end
-        status = status .. localstatus
-        status = status .. ' '
-      else
-        status = status .. localstatus
-        status = status .. ' '
-      end
+    if #localstatus > 0 then
+      table.insert(status, localstatus)
     end
   end
-  if status:len() > 0 then
-    if separator:len() > 0 and table.maxn(section) > 1 then
-      return ' ' .. status .. ' '
-    end
-    return ' ' .. status
+  if #status == 0 then
+    return ''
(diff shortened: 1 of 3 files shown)


lua/lualine/utils.lua
```diff
@@ -5,29 +5,23 @@ function M.setTheme(theme)
...
-  if status:len() > 0 then
-    if separator:len() > 0 and table.maxn(section) > 1 then
-      return ' ' .. status .. ' '
-    end
-    return ' ' .. status
+
+  local extra = ''
+  if #status ~= 0 and table.maxn(section) > 1 then
+    extra = ' '
+  end
```
elihunter173, line 19: I included `extra` because of lines `25-27` of the old version. I am not 100% sure why it's necessary because I haven't read the rest of the code, but it would be nice to remove if it's unnecessary

hoob3rt: What you refer to as `extra` is simply padding for components, your code is close to perfect. I think this code is easier to understand and handles all cases
```lua
function M.drawSection(section, separator)
  local status = {}
  for _, statusFunction in pairs(section) do
    local localstatus = statusFunction()
    if #localstatus > 0 then
      table.insert(status, localstatus)
    end
  end
  if #status == 0 then
    return ''
  end
  local sep = ' '
  if #separator > 0 then
    sep = ' ' .. separator .. ' '
  end
  local padding = ' '
  return  padding .. table.concat(status, sep) .. padding
end

```

hoob3rt: Otherwise everything looks good and it's ready to merge

elihunter173: Ah, that makes sense. It uses your code now with `padding` as a variable replaced with just `' '`

hoob3rt: Great! Thanks for that!

PR #8, opened 2021-01-01
Porting onedark theme

teddylear:
Please let me know if there is anything else that needs to be done here. Used [onedark's airline theme in it's repo](https://github.com/joshdick/onedark.vim/blob/master/autoload/airline/themes/onedark.vim) as reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ lualine.theme = 'gruvbox'
 
 * gruvbox
 * dracula
+* onedark
 
 Please create a pr if you managed to port a popular theme before me
 </details>
--- a/lua/lualine/themes/onedark.lua
+++ b/lua/lualine/themes/onedark.lua
@@ -1,0 +1,101 @@
+local M = {  }
+
+local Colors = {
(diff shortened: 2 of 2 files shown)


hoob3rt:
pic-window-210102-1652-18
Are you sure you have not mistaken `fg` with `bg`? This looks weird...

hoob3rt:
Also you have not set `M.command` as said in https://github.com/hoob3rt/lualine.nvim/issues/9

Please set it to something, `lualine`'s highlight should fall back to `lualine_normal` but it is currently not supported. Even
```lua
M.command = M.normal
```
will do fine as long as the highlights make sense with ported colorscheme.

teddylear:
@hoob3rt Thanks for the review! I addressed both comments with another commit, please let me know if there is anything else I should address.

hoob3rt:
@teddylear looks good now. Please add onedark to `README.md` and it should be ready to merge.

teddylear:
@hoob3rt Added theme to list on README.

PR #17, opened 2021-01-03
Gotham theme

kxmndev:
Colorscheme based on (gotham)[https://github.com/whatyouhide/vim-gotham] with little adaptations to not be so brighter and trouble coding focus.

### DIFF
--- a/lua/lualine/themes/gotham.lua
+++ b/lua/lualine/themes/gotham.lua
@@ -1,0 +1,52 @@
+local M = {  }
+
+local P = {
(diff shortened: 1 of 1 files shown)


hoob3rt:
I personally think statusline should be a slightly different color than `Normal`'s `bg` otherwise `lualine_c` and `lualine_x` blend in with code inside vim. 
In addition to that, I think that `lualine_inactive` should have slightly different colors than other highlights.

hoob3rt:
inactive

PR #23, opened 2021-01-04
Added powerline theme

shadmansaleh:
#### Checkout if it feels like powerline
continuing #22

I think mirroring theme on left(A,B,C) and right(X,Y,Z) is limiting.
I kind of like the left and right structure in lightline.
can we have something like that in lualine ?

Also I feel like the colours in normal mode should be default . if a cirtain modes values aren't defined lualine should fallback to normal modes colors.

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -15,6 +15,14 @@ All available themes are only best effort ports by myself/ other users. If you f
 ![command](https://user-images.githubusercontent.com/41551030/103467919-2ba8b380-4d54-11eb-8585-6c667fd5082e.png)
 ![replace](https://user-images.githubusercontent.com/41551030/103467925-32372b00-4d54-11eb-88d6-6d39c46854d8.png)
 
+### powerline
+![normal](https://user-images.githubusercontent.com/13149513/103506288-8fcb9480-4e86-11eb-97f8-3768a34188fd.jpg)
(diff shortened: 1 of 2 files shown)


hoob3rt:
> #### Checkout if it feels like powerline

Thank you for porting another theme!


> I think mirroring theme on left(A,B,C) and right(X,Y,Z) is limiting.
> I kind of like the left and right structure in lightline.
> can we have something like that in lualine ?

I will open a new issue to discuss this.

> Also I feel like the colours in normal mode should be default . if a cirtain modes values aren't defined lualine should fallback to normal modes colors.

This is something I am already aware of, I will try to implement this asap. https://github.com/hoob3rt/lualine.nvim/issues/13

THEMES.md
```diff
@@ -15,6 +15,14 @@ All available themes are only best effort ports by myself/ other users. If you f
 ![command](https://user-images.githubusercontent.com/41551030/103467919-2ba8b380-4d54-11eb-8585-6c667fd5082e.png)
 ![replace](https://user-images.githubusercontent.com/41551030/103467925-32372b00-4d54-11eb-88d6-6d39c46854d8.png)
 
+### powerline
```
hoob3rt, line 18: Please swap `powerline` with `onedark` to maintain alphabetical order

hoob3rt:
All good now, thanks.

PR #26, opened 2021-01-04
Generates colorscheme automatically

kxmndev:
Generates theme automatically using syntax highlight of current colorscheme.
The colors became perfect 95% of time... as it's only code doing the work, is good while have no official port or need to be fast, and being someone who uses too much colorschemes ;)

### DIFF
--- a/lua/lualine/themes/auto.lua
+++ b/lua/lualine/themes/auto.lua
@@ -1,0 +1,58 @@
+function _getHi( scope, syntaxlist )
+	local color
+	for i , syn in ipairs( syntaxlist  ) do
+		color = vim.api.nvim_command_output(
+			[[ execute 'hi ]] .. syn .. [[']]
+		)
+		if color:match( [[gui=reverse]] ) then
+			if scope == 'guifg' then scope = 'guibg' else scope = 'guifg' end
+		end
+		color = color:match( scope .. [[=(#?%w+)]]) or nil
+		if color then break end
+	end
+	return color or '#000000'
+end
+
+local _ = {
+	normal  = _getHi( 'guibg', {'PmenuSel', 'PmenuThumb', 'TabLineSel' } ),
+	insert  = _getHi( 'guifg', {'String', 'MoreMsg' } ),
+	replace = _getHi( 'guifg', {'Number', 'Type' } ),
+	visual  = _getHi( 'guifg', {'Special', 'Boolean', 'Constant' } ),
+	command = _getHi( 'guifg', {'Identifier' } ),
+	back1   = _getHi( 'guibg', {'Normal', 'StatusLineNC' } ),
+	fore    = _getHi( 'guifg', {'Normal', 'StatusLine' } ),
+	back2   = _getHi( 'guibg', {'StatusLine' } ),
+}
+
+local M = {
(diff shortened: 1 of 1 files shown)


hoob3rt:
This sounds insanely good! I will test it asap.

hoob3rt:
It would be great to integrate this with https://github.com/hoob3rt/lualine.nvim/pull/27 to make `cterm` colors automatic too. 

Also it would be nice to have automatic `cterm` colors generated for themes which only have `gui` colors. Do you have an idea on how to implement it?

hoob3rt:
I tested it, the idea and direction is nice but there has to be some more testing done. For my on most themes I installed the statusline is 50% black, not optimal.

kxmndev:
Yes, at first I thought be a default for who wants to match the themes color as use a non conformant theme for the status line is very annoying...

It's a funny thing that I really wanna swim to... but, in website migration, I'm busy this days.

What I think... there are some simple algorythms for color calculation. 

1. Simply adjust RGB to be 5% lighter in darker themes in each channel (or 5% darker in lighter themes). Not something too complicated as a state of art, but simple calculations to avoid hurt performance. This could powers terminal colors too.

2. The second step, and a bit more delicate, is the control of contrast. We could use the idea found in https://24ways.org/2010/calculating-color-contrast/ in Lua. So we could inspect each ":hi" syntax  of the basic types (Identifier, Number, String, Boolean, NonText, LineNr, StatusLine etc.) matching its contrast to surroundings, to be different of page background and with contrast with its text.

3. As the complexity increases, may be useful, once well done, memoize the result after first run of code.

The process to create ctermbg e ctermfg e guisp are simple, a question of more itens in the table specification. Except that how to detect color contrast in a 16 colors terminal? Consider that some people rearrange (like me) the dark shade of red to be red and the light shade of red to be orange... the light green to be lemon green (#ccff00) or yellow to be more reddish than pure yellow. May be the color generated in such terminals could be a simpler greyish... the sufficient to be util, since the less equipped  terminals are just used on communications, where the editor must be fast and unobstrusive to make the needed urgent changes, since the systems tend to be developed using a well equiped environment.

I hope sincerely can help. This moment ideas and comment is the best I can do.

hoob3rt:
First of all thank you for your reply and for your idea. I think creating themes automatically is a neat feature and should be implemented somewhere down the line. 

> It's a funny thing that I really wanna swim to... but, in website migration, I'm busy this days.

I don't have too much time right now either due to finals... I will try to work on it asap but there are a few things in the project that are higher priority to me than automatic theming so you might find time to do it before I do.

> What I think... there are some simple algorythms for color calculation.

We already have a non trivial way of setting perfect cterm colors thanks to @shadmansaleh so that's out of the window :)

But I think your POC is great, we can handle cterm colors the same way you tried to handle gui colors instead of using mentioned algorithm, this should work better in terms of performance.

> As the complexity increases, may be useful, once well done, memoize the result after first run of code.

This might be overkill, the auto group would only run on startup and I don't think it would affect performance too much. Needs to be tested when the `auto` theme works well tho.

> I hope sincerely can help. This moment ideas and comment is the best I can do.

Yes it sure does! Good luck on your website :)

PR #27, opened 2021-01-04
Added notermguicolor support and special effects

shadmansaleh:
### Changes
- Added notermgui support
- Added calculating cterm values from rgb hax codes
- updated nord theme as POC
- Exposed special parameters for highlight(like bold,italic) to themes
  with sp like to bold "B" in normal mode do normal.b.sp = 'bold'
- Checked notermguicolor support in README.md
- Changed CONTRIBUTING.md to reflect current notermguicolor support

It's backwords compatable so old themes will work .


hoob3rt:
Good job! This looks like an optimal solution.

CONTRIBUTING.md
```diff
@@ -18,30 +18,34 @@
 <summary><b>theme example</b></summary>
 
 To create a custom theme you need to define a colorscheme for each of vim's modes. Each mode has a `fg` and `bg` field for every lualine section.
+You can add special effects with `sp`.
```
hoob3rt, line 21: I would personally change `sp` to `gui` since this is what is used to create vim's `highlight` commands.

lua/lualine/highlight.lua
```diff
@@ -1,20 +1,34 @@
 local M = {  }
 
 local function highlight (name, foreground, background, special)
```
hoob3rt, line 3: here as well, `special` to `gui`, this is a mistake on my part

CONTRIBUTING.md
shadmansaleh: ok I'll change that

shadmansaleh: both cterm and gui create those effects .  
Thats why I thought of not going with gui.

shadmansaleh:
Done

CONTRIBUTING.md
hoob3rt: Yea the naming convention for highlights is weird but I think it's better to stick with it to avoid further confusion

shadmansaleh:
I've added a function to calculate cterm values from rgb colors . Now even the themes that don't provide those values will have notermgui support :) .
Though the color seems abit washed out see if you can find any impovement.

CONTRIBUTING.md
```diff
@@ -18,30 +18,34 @@
...
+
+  lightgray = {'#504945', 7},
+  inactivegray = {'#7c6f64', 7},
 }
 
 gruvbox.normal = {
   a = {
     bg = colors.gray,
     fg = colors.black,
+    gui = "italic",
```
hoob3rt, line 48: `a` should be bold by default so the example should also provide `gui="bold"`

lua/lualine/highlight.lua
```diff
@@ -1,24 +1,58 @@
...
+      return math.floor(num)
+    else
+      return math.ceil(num)
+    end
+  end
+
+  local r = tonumber(color:sub(2,3), 16)
+  local g = tonumber(color:sub(4,5), 16)
+  local b = tonumber(color:sub(6,7), 16)
+  return round(36 * round(r * 5/256) +
```
hoob3rt, line 15: Could you explain how this works? Maybe then we will be able to figure out a better solution. This creates really weird colors for `gruvbox` theme

lua/lualine/highlight.lua
```diff
@@ -1,24 +1,58 @@
...
+  local g = tonumber(color:sub(4,5), 16)
+  local b = tonumber(color:sub(6,7), 16)
+  return round(36 * round(r * 5/256) + 
+               6 * round(g * 5/256) +
+               round(b * 5 / 256) + 16) 
+end
+
+local function highlight (name, foreground, background, gui)
+  local command = {}
+  if type(foreground) == 'table' and type(background) == 'table' then
```
hoob3rt, line 22: This is a mess, here is how to do it better
```lua
local function highlight (name, foreground, background, gui)
  local command = {
    'highlight', name,
    'ctermfg=' .. (foreground[2] or get_cterm_color(foreground)),
    'ctermbg=' .. (background[2] or get_cterm_color(background)),
    'cterm=' .. (gui or 'none'),
    'guifg=' .. (foreground[1] or foreground),
    'guibg=' .. (background[1] or background),
    'gui=' .. (gui or 'none'),
  }
  return table.concat(command, ' ')
end
```

lua/lualine/themes/nord.lua
```diff
@@ -1,13 +1,13 @@
...
 
 local colors = {
-  nord1  = "#3B4252", 
-  nord3  = "#4C566A", 
-  nord5  = "#E5E9F0", 
-  nord6  = "#ECEFF4", 
-  nord7  = "#8FBCBB", 
-  nord8  = "#88C0D0", 
-  nord13 = "#EBCB8B",
+  nord1  = {"#3B4252", 0},
```
hoob3rt, line 4: I understand this is a poc but please change it to look like an actual nord theme, your poc looks less nordy than the automatically generated one

hoob3rt (changes requested):

lua/lualine/highlight.lua
shadmansaleh: checkout https://stackoverflow.com/questions/27159322/rgb-values-of-the-colors-in-the-ansi-extended-colors-index-17-255#27165165

The issue is high contrast colors are in first 16 numbers . for 8bit and extended 8bit colors 16-232 are separated as rgb colors after that there are various shades of gray . What I'm doing here is pushing 6byte rgb in that 16-232 range . But we are missing out on those first 16 and later grays . I think thats whats causing them . I'm not sure how to incorparate those colors in this . I might try to add first 16 but I have no idea about later grays . See if you can find a better algorithm

CONTRIBUTING.md
shadmansaleh: `a` will be bold even if not specified . That's why I thoght of giving something different. To show how to change it . Showing `a` set to bold might send a wrong message that they have to set it too.

lua/lualine/highlight.lua
shadmansaleh: Agreed

lua/lualine/themes/nord.lua
shadmansaleh: I'm sure thats because those are 8-colors . The automated genarater uses 256 colors . I donn't know too much about 256 colors. I can just add the function  genarted values in the theme.

shadmansaleh: >your poc looks less nordy than the automatically generated one

Are you talking about the theme or the theme without no termgui support

shadmansaleh:
256_colors_bg

Thats 256 colors palate. Theres nothing like the sheds of gray used in gruvbox in those 256 colors . There are limitations when that few colors are used. That's why they get rendered like different gray .
(after all RGB has 16,777,216 colors that we are trying to fit in 256 colors)

lua/lualine/themes/nord.lua
hoob3rt: I'm talking about `notermguicolors`. I think we should provide sane `cterm` colors for all currently available themes with this change. I will work on them on friday.

lua/lualine/highlight.lua
hoob3rt: tmux has a way of handling grays [here](https://github.com/tmux/tmux/blob/dae2868d1227b95fd076fb4a5efa6256c7245943/colour.c#L57). 
You can look at it closely and let me know if you think their implementation would be an overkill for our usecase or not.

CONTRIBUTING.md
hoob3rt: What I'm afraid of is people copying this without too much thinking while creating their themes. Maybe make this a comment and imply that it's optional?
```lua
-- optional gui parameter
-- gui = "italic", 
```

shadmansaleh: Ok i'll comment it

lua/lualine/highlight.lua
shadmansaleh: I actualy came across that . But didn't realy check what they were doing . I'll check it out when I'm free and let you know

lua/lualine/themes/nord.lua
shadmansaleh: >I think we should provide sane cterm colors for all currently available themes

I agree . I'll work on then when I get time

lua/lualine/highlight.lua
hoob3rt: Great, I will check it out when I'm free as well and we will be able to proceed further

CONTRIBUTING.md
shadmansaleh: Maybe we should remove default bold on `a` element for consistency . Since we will edit all the themes we will just add theme in the theme files . What do do you think?

hoob3rt: Yeah, let's do it. 
We should also change the recently added comment to
```lua
-- optional gui parameter
gui = "bold",
```

shadmansaleh:
Themes checklist

- [x] nord
[x] - powerline
[ ] - one dark
[ ] - gruvbox
[ ] - dracula
[ ] - forest night

shadmansaleh:
Bad tap naver mind . See if it fells more nordy now

CONTRIBUTING.md
```diff
@@ -18,30 +18,34 @@
...
-  blue = '#83a598',
-  yellow = '#fe8019',
-
-  gray = '#a89984',
-  darkgray = '#3c3836',
-
-  lightgray = '#504945',
-  inactivegray = '#7c6f64',
+ -- color format { termguicolor, notermguicolor}
+  black = {"#282828", 0},
```
hoob3rt, line 29: These termguicolors need to be changed after we port `gruvbox` theme to support notermguicolors

hoob3rt (commented):
Everything looks good now, amazing work!

shadmansaleh:
I've rewritten the cterm autogenaration . Check it out . It works really well . I donn't think we even need to add cterm values to themes any more.

CONTRIBUTING.md
```diff
@@ -18,30 +18,34 @@
...
-  green = '#b8bb26',
-  blue = '#83a598',
-  yellow = '#fe8019',
-
-  gray = '#a89984',
-  darkgray = '#3c3836',
-
-  lightgray = '#504945',
-  inactivegray = '#7c6f64',
+ -- color format { termguicolor, notermguicolor}
```
hoob3rt, line 28: as you said, cterm colors work really well now, we can remove them from here and let them be calculated by your function

lua/lualine/utils.lua
```diff
@@ -18,4 +18,303 @@ function M.draw_section(section, separator)
...
+      local db = math.abs(color1[3] - color2[3]) / (color1[3]+1) * 100
+      return (dr + dg + db)
+    end
+
+    local r = tonumber(hex_color:sub(2,3), 16)
+    local g = tonumber(hex_color:sub(4,5), 16)
+    local b = tonumber(hex_color:sub(6,7), 16)
+
+    -- check which color deviates lest from haxcolor
+    local lest_deviation = {0, 10000}
```
hoob3rt, line 306: This algorithm could probably be improved in time complexity, I will take a look at it later.

hoob3rt (commented):
You are right, automatically generated cterm colors look really good now, we can safely remove them from all theme definitions.

CONTRIBUTING.md
shadmansaleh: Maybe leave them as they are . Allowing theme authors to provide them will make it easier for them to fix if the algorithm choses a wrong color.
Also it's better if we donn't have to calculate it in runtime. Plus genarating those values from function is pretty easy . I've added cterm values to those last 4 themes in a few minutes .
We might just add instruction to do that when docs are written.

shadmansaleh:
I've updated CONTRIBUTING.md . You can try improving the algorithm .Apart from that I think it's complete.

CONTRIBUTING.md
hoob3rt: I think you are right.

hoob3rt:
> I've updated CONTRIBUTING.md . You can try improving the algorithm .Apart from that I think it's complete.

I agree. I will look into the algorithm and `CONTRIBUTING.md` later today.

hoob3rt:
I updated theme formatting to make them more compact, I think it makes them clearer to read while also reducing code size slightly.

lua/lualine/utils.lua
hoob3rt, line 306: lest deviation? I think you mean `least` or `smallest` but then again I don't know these math terms in english

shadmansaleh: Well I'm not a native english speaker either . What it means is "The color in the table that is closest to hex_color in terms of rgb values"

hoob3rt:
```lua
    -- check which cterm color is closest to hex colors in terms of rgb values
    local closes_cterm_color = nil
    local min_distance = 10000
    for _, color in ipairs(color_table) do
      local current_distance = get_color_deviation(color[2], {r,g,b})
      if current_distance < min_distance then
        min_distance = current_distance
        closes_cterm_color = color[1]
      end
    end
    return closes_cterm_color
```
This should be more readable and avoid using math terms neither of us know in english

shadmansaleh: fixed

shadmansaleh:
I've changed implementation to a function . no more tables and meta tables

hoob3rt (approved):
Nice, I'm glad we got this resolved. Thank you for your contribution :)

shadmansaleh:
Wellcome

Issue #25, opened 2021-01-04
Change theme structure


hoob3rt:
> I think mirroring theme on left(A,B,C) and right(X,Y,Z) is limiting.
> I kind of like the left and right structure in lightline.
> can we have something like that in lualine ?

I personally can't see how lualine's theme structure is limiting compared to `lightline`. Could you elaborate on what exactly are you missing?

_Originally posted by @shadmansaleh in https://github.com/hoob3rt/lualine.nvim/pull/23#issue-548058407_

samirettali:
I think he is referring to the fact that he can't change, for example, the background color of block `a` to be red and of block `z` to be green, they must be either both red or both green.

I modified my local configuration to be able to set individual block colors and not in a mirrored way, I'd be happy to open a PR if needed.

shadmansaleh:
I'm voting to close this issue . Since it was started on my comment . I donn't thunk it's nessary any more.

Issue #13, opened 2021-01-02
Lualine's highlight should fall back to `lualine_normal` if current mode's highlight is not set


hoob3rt:
`lualine`'s highlight should fall back to `lualine_normal` but it is currently not supported. 

_Originally posted by @hoob3rt in https://github.com/hoob3rt/lualine.nvim/issues/8#issuecomment-753490011_

hoob3rt:
resolved with https://github.com/hoob3rt/lualine.nvim/commit/6da4406a5a5996038bcb09e46a2824c57b0b1856

PR #28, opened 2021-01-04
Themes now fallback to normal mode theme

shadmansaleh:
A solution to #13 

Since I'm setting defaults here, I though we could make element `(x,y,z)` default to `(c,b,a)`. And separate them in internal code . So if a theme author specifies `(x,y,z)` it statusline will be themed acording to that . otherwise left and right side will be symetrical by default .

So I've also added a implementation for   #25

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -13,10 +13,13 @@ end
 local function apply_defaults_to_theme(theme)
   local modes = {'insert', 'visual', 'replace', 'command', 'terminal', 'inactive'}
   for _, mode in ipairs(modes) do
-    theme[mode] = (theme[mode] or theme['normal'])
-    theme[mode]['a'] = (theme[mode]['a'] or theme['normal']['a'])
-    theme[mode]['b'] = (theme[mode]['b'] or theme['normal']['b'])
-    theme[mode]['c'] = (theme[mode]['c'] or theme['normal']['c'])
+    if not theme[mode] then
+      theme[mode] = theme['normal']
+    else
+      for section_name, section in pairs(theme['normal']) do
+        theme[mode][section_name] = (theme[mode][section_name] or section)
+      end
+    end
   end
   return theme
 end
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -10,7 +10,33 @@ local function highlight (name, foreground, background, special)
   return table.concat(command, ' ')
 end
 
+local function apply_defaults_to_theme(theme)
+  local modes = {'insert', 'visual', 'replace', 'command', 'terminal', 'inactive'}
+  -- normal mode cann't have a fallback for now
+  local sections = {'a', 'b', 'c'}
+  for _, mode in ipairs(modes) do
+    repeat
+      if theme[mode] == nil then
+        theme[mode] = theme['normal']
+        break
+      end
+      for _, section in ipairs(sections) do
+        repeat
+          if theme[mode][section] == nil then
+            theme[mode][section] = theme['normal'][section]
+            break
+          end
+          break
+        until true
+      end
+      break
+    until true
+  end
+  return theme
+end
+
 function M.create_highlight_groups(theme)
+  apply_defaults_to_theme(theme)
   for mode, sections in pairs(theme) do
     for section, colorscheme in pairs(sections) do
       local special = nil
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -12,25 +12,11 @@ end
 
 local function apply_defaults_to_theme(theme)
   local modes = {'insert', 'visual', 'replace', 'command', 'terminal', 'inactive'}
-  -- normal mode cann't have a fallback for now
-  local sections = {'a', 'b', 'c'}
   for _, mode in ipairs(modes) do
-    repeat
-      if theme[mode] == nil then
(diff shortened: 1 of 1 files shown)


lua/lualine/highlight.lua
```diff
@@ -10,7 +10,41 @@ local function highlight (name, foreground, background, special)
...
+  -- normal mode cann't have a fallback for now
+  local sections = {'a', 'b', 'c', 'x', 'y', 'z'}
+  for _, mode in ipairs(modes) do
+    repeat
+      if theme[mode] == nil then
+        theme[mode] = theme['normal']
+        break
+      end
+      for _, section in ipairs(sections) do
+        repeat
```
hoob3rt, line 24: I don't understand what is `repeat` and `until true` supposed to do in this situation, could you elaborate? `luackeck` marks `until true` as unreachable code...

lua/lualine/highlight.lua
```diff
@@ -10,7 +10,41 @@ local function highlight (name, foreground, background, special)
...
+          if theme[mode][section] == nil then
+            if section == 'x' then
+              theme[mode][section] = theme[mode]['c']
+            elseif section == 'y' then
+              theme[mode][section] = theme[mode]['b']
+            elseif section == 'z' then
+              theme[mode][section] = theme[mode]['a']
+            else
+              theme[mode][section] = theme['normal'][section]
```
hoob3rt, line 33: Will this yield the same results? I think it's a lot cleaner and still has room for improvements.
```lua
  for _, mode in ipairs(modes) do
    repeat
      theme[mode] = (theme[mode] or theme['normal'])
      theme[mode]['x'] = (theme[mode]['x'] or theme[mode]['c'])
      theme[mode]['y'] = (theme[mode]['y'] or theme[mode]['b'])
      theme[mode]['z'] = (theme[mode]['z'] or theme[mode]['a'])
      for _, section in ipairs(sections) do
        repeat
          theme[mode][section] = theme['normal'][section]
          break
        until true
      end
      break
    until true
  end
```

hoob3rt (changes requested):

lua/lualine/highlight.lua
shadmansaleh: (repeat ... break.. utill true) it's basically a continue statement(lua doesn't have continue). I didn't like the structure either . Second repeat loop was unnessary .

lua/lualine/highlight.lua
shadmansaleh: nice . I like it . I'll update it .

shadmansaleh:
I've updated it . Now it's alot saner .

hoob3rt:
While we are at it I think it would be nice to have an option to mirror theme or not, something like `M.mirror_theme = true` in `lualine.lua`. 
I'm not sure if this is a good idea though. If this change is made there would have to be changes made to existing themes and an additional enforcement for newer theme contributions to at least set `lualine_z` as well.

Let me know what you think.

shadmansaleh:
>I'm not sure if this is a good idea though. If this change is made there would have to be changes made to existing themes and an additional enforcement for newer theme contributions to at least set lualine_z as well.

Sorry i didn't get what you ment. In this pull `x,y,z` gets mirrored if not provided . It means if I want I can mirror the `z` with a in normal mode . and have a different color on `z` and `a` in insert mode . Since `x,y,z` aren't given in current themes they will work as they are now . (Themeing `x,y,z` is optional)

hoob3rt:
> > I'm not sure if this is a good idea though. If this change is made there would have to be changes made to existing themes and an additional enforcement for newer theme contributions to at least set lualine_z as well.
> 
> Sorry i didn't get what you ment. In this pull `x,y,z` gets mirrored if not provided . It means if I want I can mirror the `z` with a in normal mode . and have a different color on `z` and `a` in insert mode . Since `x,y,z` aren't given in current themes they will work as they are now . (Themeing `x,y,z` is optional)

That is correct. What I meant is if this is to be merged then all themes should have `lualine_z` defined and give user a choice to choose between mirroring or not by an additonal setting. Otherwise some themes might define mirroring and some not which would be a mess.

shadmansaleh:
>Otherwise some themes might define mirroring and some not which would be a mess.

Even if we enforce `lualine_z` some themes might define `y/z` and some might not.

shadmansaleh:
I think themes should be responsible of there looks .

shadmansaleh:
I've removed #25 related patches from this pr .
Now this can be merged.
We can discuss how to solve #25 in #31

lua/lualine/highlight.lua
```diff
@@ -10,7 +10,19 @@ local function highlight (name, foreground, background, special)
   return table.concat(command, ' ')
 end
 
+local function apply_defaults_to_theme(theme)
+  local modes = {'insert', 'visual', 'replace', 'command', 'terminal', 'inactive'}
+  for _, mode in ipairs(modes) do
+    theme[mode] = (theme[mode] or theme['normal'])
+    theme[mode]['a'] = (theme[mode]['a'] or theme['normal']['a'])
```
hoob3rt, line 17: This could be further cleaned up, especially when there is a possibility that there will be more section themes
Something like this could work
```lua
for section_name, section in pairs(theme['normal']) do
  theme[mode][section_name] = (theme[mode][section_name] or section)
end
```

hoob3rt (approved):
All good now, thanks

PR #29, opened 2021-01-04
Add component for coc.nvim

smarquez1:
Added support for `coc.nvim`  based from my local configuration.

Some thoughts:
* I created many functions to display the status. We can separate these into 2 or more components (such as coc_status, coc_diagnostics).
* We can mimic VScode behaviour, which always shows the Error and Warning count and only shows Info and Hints if there's any.
* We can add icons

Thanks for the great plugin! Suggestions are very welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ lualine.inactive_sections = {
   * progress
 * plugin
   * signify
+  * coc
 
 </details>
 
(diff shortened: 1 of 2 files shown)


hoob3rt:
Your code looks good. However I am planning on providing a diagnostic api in lualine to view diagnostics from multiple sources if wanted in the statusline. Your implementation will be merged when this is implemented.

hoob3rt:
implemented at master with #95

PR #31, opened 2021-01-05
Asymatric themeing

shadmansaleh:
Continuing #28
Allows `x,y,z` to be indivisualy themed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,7 @@
 
 To create a custom theme you need to define a colorscheme for each of vim's modes. Each mode has a `fg` and `bg` field for every lualine section.
 This is really easy in lua. Here is and example of a gruvbox theme.
+By default `x,y,z` mirror `c,b,a` . But you can set `x,y,z` to specifically theme theme them.
 
 ```lua
 local gruvbox = {  }
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -88,15 +88,15 @@ local function statusline(is_focused)
   end
   table.insert(status, "%=")
   if sections.lualine_x then
-    table.insert(status, highlight.format_highlight(is_focused, 'lualine_c'))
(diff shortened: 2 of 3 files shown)


hoob3rt:
Why did you close this pr?

shadmansaleh:
While working on #55 I've realized this wonn't work .
If we are to create asymatric themes we will have to keep  a middle part that is always present . So this cann't work .

Even curently the statusline depends on user keeping every component active . Try commenting out `c` in your config or all `x,y,z` or `a,b,c` you will find some amazing results. best one is only have `a` available . The whole statusline takes up a . I've tried to deal with it on #55 .

hoob3rt:
I understand what the issue is, I have not thought of it before. Please open a new issue with this. 

I think we should try to tackle it in a new pr BEFORE we do #55. Also, #55 should handle only what it states, I had no idea it fixes any theming issues right now :)

shadmansaleh:
>I think we should try to tackle it in a new pr BEFORE we do #55. Also, #55 should handle only what it states, I had no idea it fixes any theming issues right now :)
 
I basicaly had to rewrite most of statusline function for implementing #55 as it required new highlights for each separator . As I've changed the interpretation of themes there to make looks more consistent . (Simply I've made so `c` sections highlighting is always present. An `a,b,c` compartment always render on leftaligned and `x,y,z` rightaligned )

I think if we want to change themeing we can do that on top of #55 .

shadmansaleh:
Those bugs are mainly there because of curent implementation of statusline function doesn't account for execption . Specialy on section_c it '%=' if c is skiped it's not added . Even if those bugs are fixed in this implementation . After #55 merges they will be replaced by new implementation of statusline function.
So I think fixing them here is unnessery.

hoob3rt:
Yeah sure. I will take a look at #55 asap and hopefully we will get it fixed :)

Issue #37, opened 2021-01-07
Git branch persists even after buffer change


coxackie:
Suppose I open a buffer and the file is at a git folder. The branch is displayed, which is good.

 Now, I switch to another buffer with a file at another folder that is not (and neither its parents are) git-controlled. The git branch from the previous buffer will persist on the status line for the new buffer. Even if I close the previous buffer, the git branch persists.

 Not sure what the issue is; maybe something about async?

hoob3rt:
Nice catch, this is also an issue with the `signify` component but I think these are two different bugs, I will try to fix this tomorrow.

hoob3rt:
Should be  fixed :)

coxackie:
thanks, looks good, indeed. If there is any issue, I will report.

PR #48, opened 2021-01-08
Fix branch not changing

shadmansaleh:
#37 
git branch wasn't changing because git command was being executed in current directory rather then file directory .

I donn't know what the async code actually doing . So there might be a better way to solve it.

### DIFF
--- a/lua/lualine/components/branch.lua
+++ b/lua/lualine/components/branch.lua
@@ -13,7 +13,10 @@ local get_git_branch = async:new({
 
 local timer = vim.loop.new_timer()
 timer:start(0, 1000, vim.schedule_wrap(function()
+  local cur_dir = vim.fn.getcwd()
+  vim.api.nvim_set_current_dir(vim.fn.expand("%:p:h"))
   get_git_branch:start()
+  vim.api.nvim_set_current_dir(cur_dir)
 end))
 
 local function branch()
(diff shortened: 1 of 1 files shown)


hoob3rt:
What you did is also done by `set autochdir`. But I think this is not an optimal solution, although it fixes the issue of branch not changing, we don't want to set global vim variables for users.

That's not the only issue tho, even with your patch if you open a file in git directory and switch to a file outside of a git directory the branch persists.

There might be a fix with `on_stderr` from `async`, you can try to have fun with it.

shadmansaleh:
>That's not the only issue tho, even with your patch if you open a file in git directory and switch to a file outside of a git directory the branch persists.

Oh I didn't notice that before . I've added a stderr detection now branch gets removed when not in a git-repo.

lua/lualine/components/branch.lua
```diff
@@ -8,12 +8,22 @@ local get_git_branch = async:new({
...
+        git_branch = ''
+      end
+    end
+  end,
 })
 
 local timer = vim.loop.new_timer()
 timer:start(0, 1000, vim.schedule_wrap(function()
+  local cur_dir = vim.fn.getcwd()
+  vim.api.nvim_set_current_dir(vim.fn.expand("%:p:h"))
```
hoob3rt, line 24: This should be error checked, right now it produces errors in `fzf` in floating mode, probably due to temp buffers or sth.
pic-window-210108-1015-58

hoob3rt: Okay I resolved this, should be working now.

shadmansaleh:
Is there any other exception you can think of ?

hoob3rt:
> Is there any other exception you can think of ?

I changed string matching to pcall, will handle all exceptions now. I think this is done, let me know what you think

shadmansaleh:
I think pcall is the way to go.

Issue #51, opened 2021-01-08
[suggestion] use icons for fileformat


coxackie:
Instead of the boring `unix`, `dos` and `mac` for fileformat, one may use devicons (since they are also used for file types). More precisely, from `https://www.nerdfonts.com/cheat-sheet`, check out `nf-dev-linux`, `nf-dev-windows` and `nf-dev-apple`. Just eye-candy, but, hey, why not? (I do not think both icon and text should be provided. Icons if the possibility is there, text for fallback.)

shadmansaleh:
Seems like a nice idea

coxackie:
I think it will make it look prettier, and also will relieve some space. I am debating on whether `nf-pl-current_line` and `nf-ple-current_column` would be good to use in `location`. They do look good, but maybe they will add up on space...

hoob3rt:
This is related to #16 and #45. Hopefully in the near future you will be able to set
```lua
lualine.icons = true
lualine.powerline_fonts = true
lualine.sections.lualie_a = { 'filetype' icons false } -- something along those lines
```

hoob3rt:
resolved with #45

PR #52, opened 2021-01-08
Added icons for fileformat

shadmansaleh:
Implementing #51


hoob3rt:
I think this should be integrated with #60. Close this pr if you agree.

PR #53, opened 2021-01-08
Added other sources for component configuration

shadmansaleh:
When a string is passed first it's checked whether it's a component
if it's not a component use it as a vim function

This allows configs like
    ```lualine_b = { 'FugitiveHead', 'coc#status' }```
to work.

Added support for vim variables and lua strings to.
If a plugin exposes a variable for statusline like coc_status that can that can be used easily .
```lualine_b = { 'coc_status' } ```

Or if you just happen to want to display a constant string in you statusline you can do that too.

Also If component name is missspaled in config it will apear as a constant component and not crash lualine. It will probably help identify problems .

### CODE REVIEW

**Comment by hoob3rt** [comment]
> When a string is passed first it's checked whether it's a component
> if it's not a component use it as a vim function

That's a nice idea!

> Or if you just happen to want to display a constant string in you statusline you can do that too.

This is bad imo. Say someone creates config and passes `filefornat` instead of `fileformat`. A pretty hard config bug to catch. Now their statusline is not working properly. I think displaying custom string in statusline should be limited to lua functions just like it is now.

I think there should be an error prompt when a component fails while keeping the rest of lualine working. (replacing the broken component with '')

File: lua/lualine.lua [hunk 1]
```diff
@@ -34,7 +34,19 @@ local function load_components()
     for _, section in pairs(sections) do
       for index, component in pairs(section) do
         if type(component) == 'string' then
-          section[index] = require('lualine.components.' .. component)
+          local ok,loaded_component = pcall(require, 'lualine.components.' .. component)
+          if not ok then
+            loaded_component = function()
+              if vim.fn.exists('*'..component) ~= 0 then
+                return vim.fn[component]() 
+              elseif vim.g[component] then
```
hoob3rt, line 42: This currently only supports global variables.

shadmansaleh:
>I think there should be an error prompt when a component fails while keeping the rest of lualine working. (replacing the broken component with '')

The problem with that is other plugins might be loaded before or after lualine . Say a component depends on Vim-fugitive . if fugitive is loaded after lualine . lualine will keep producing errors until it's loaded . It doesn't have to be a plugin . A function written later in the file will act same way.

> Say someone creates config and passes filefornat instead of fileformat

come on if you see filefonat on your statusline won't you know whats wtong with it ?

lua/lualine.lua
shadmansaleh: We could add b: w: t: . Does v: make any sence ?

hoob3rt: Well... some variables are both related to window and global. There should probably be a correct order to checking them.

As of `vim.v` I have no idea how it works, always displays empty table no matter what I do.

shadmansaleh: I think we should stick with globals only .

shadmansaleh: >There should probably be a correct order to checking them.

If we are talking about prder then it clearly should be b > w > t .
But what to do when they are not available.

>As of vim.v I have no idea how it works, always displays empty table no matter what I do.

try `:help v:var`
vim.v / vim.g these tables are lazy_loading in lua . they don't store a key before it's used first time

shadmansaleh:
If we do not display strings then we will have to silently ignore errors . Also if we add b: w: t: variables we will have to silently ingnore errors . Because they will won't be available in most cases. That's why I think we should stick to g: vars . 

I think it's most likely other plugins will interface with statusline with functions or global variables . 

Unless you can tell me what's the difference between not available and naver will be available , displaying errors aren't a option as it will create hundreds of errors on startup.

shadmansaleh:
Is there any limitation on how many times statusline is updated ?

shadmansaleh:
Check it out . Now strings are nolonger shown in statusline.
If the nothing matches after first 50 iterations a warning message is shown.

hoob3rt:
> That's why I think we should stick to g: vars.

I agree.

> Is there any limitation on how many times statusline is updated every second?

I should actually research this closely, I really don't know. But my guess would be it's connected to `updatetime` variable.

>  Now strings are nolonger shown in statusline.

That's great, I think having strings there would only provide confusion.

>  If the nothing matches after first 10 seconds a warning message is shown.

This is nice but I'm not sure I like this. We should merge this without the `timer` functionality and discuss it further elsewhere.

shadmansaleh:
> This is nice but I'm not sure I like this. We should merge this without the timer functionality and discuss it further elsewhere.

So we shouldn't report errors?

hoob3rt:
> > This is nice but I'm not sure I like this. We should merge this without the timer functionality and discuss it further elsewhere.
> 
> So we shouldn't report errors?

Yeah for now...

Btw please remove all trailing whitespaces when you consider something ready to merge. I think this one is ready, will test this and merge when I have some free time

shadmansaleh:
>Btw please remove all trailing whitespaces when you consider something ready to merge

ok

hoob3rt:
Are you sure vim variables are working now? They are not working for me. 
Minimal example with [gruvbox](https://github.com/gruvbox-community/gruvbox.git)

```vim
:echo g:gruvbox_bold
```
returns 1
```vim
:lua print(vim.g.gruvbox_bold)
```
returns 1

```lua
lualine.lualine_a = {  'gruvbox_bold' }
```

This is really weird because I can use all functions exposed to `:call`. 

Btw I can't think of a global vim variable I would like to have in my statusline. `coc_status` will be useless when I create the diagnostics api. Otherwise `g` variables are usually plugin's defines. Maybe we should scrap variables and just support lua & vim functions? This should give all users really easy access to hook into lualine with their custom components.

shadmansaleh:
Fixed it . It's because it's an int . function calls that return anything other then strting was also crashing.

>Btw I can't think of a global vim variable I would like to have in my statusline. coc_status will be useless when I create the diagnostics api. Otherwise g variables are usually plugin's defines. Maybe we should scrap variables and just support lua & vim functions? This should give all users really easy access to hook into lualine with their custom components.

I think we should keep them. I think there are other plugins that provide interface like `coc_status` .

shadmansaleh:
Now users are required to mention g: for a string to be treated as a
veriable
so setup has to be 'g:coc_status' insted of 'coc_status'

That should clear things up.

hoob3rt:
> Now users are required to mention g: for a string to be treated as a
> veriable

This I can agree on. Without it there might be too much confusion. I still think there is little use for it but it might be useful for some :)

> That should clear things up.

It does! I will report back when I have time to look at the code and test it properly.

shadmansaleh:
@hoob3rt if you have time can we get this pr merged I think it will be a really nice addition to lualine

hoob3rt:
> @hoob3rt if you have time can we get this pr merged I think it will be a really nice addition to lualine

Yeah that's next on my list. I will try to finish this asap

shadmansaleh:
I've added v:, t:, w:, b:, o, go:, vo:, to:, wo:, bo:  veriables and options as component source too .

argument for having them 
https://github.com/neoclide/coc-git
uses b: veriables along with g: veriables for statusline . Some w: specific veriables might b usefull for some plugins too .

As for options if user wants to display value of some option they can do that . Since the scope have to be explicitly mentioned I donn't thingk it will create confution , I beleave vim users know what are v:, t:, w:, b:, o, go:, vo:, to:, wo:, bo:
It was just 1LOC so why not...

Also now errors when veriables donn't exist are captured so it wonn't cause errors when veoables aren't available.

@hoob3rt  can you tell me when you will be available ??

lua/lualine.lua
```diff
@@ -34,7 +34,35 @@ local function load_components()
     for _, section in pairs(sections) do
       for index, component in pairs(section) do
         if type(component) == 'string' then
-          section[index] = require('lualine.components.' .. component)
+          local ok,loaded_component = pcall(require, 'lualine.components.' .. component)
+          if not ok then
+            -- vim veriable component
+            -- accepts g:, v:, t:, w:, b:, o, go:, vo:, to:, wo:, bo:
+            if component:find('[gvtwb]?o?:') == 1 then
```
hoob3rt, line 41: I would split this to handle both `g:something` and `vim.bo.something` to make it more distinct on when the vim syntax and lua syntax is used.

shadmansaleh: vim.bo.something aren't currently handled . Actualy lua veriables csnn't be used as a component for now . Are you saying to implement just vim.b/vim.g kind of lua veriable and not other lua globals wonn't that be inconsistent?

shadmansaleh:
@hoob3rt  have your exams ended ? If it has and you have sometime can we get these PRs merged  this week . I wonn't be available from next week . It's been like a month . At lest give me a date on when you will be available......

hoob3rt:
This pr tomorrow

shadmansaleh:
No nitpicks huh 🤔

hoob3rt:
Yea, this lgtm, nothing to nitpick really. Everything works, is well documented and implemented nicely IMO.

shadmansaleh:
I've been using these fetures of these prs for a mounth . So I'm pretty confident all of them work as intened . You can merge #91 it's just a small fix  in docs . Can we merge anyother feature prs today?

hoob3rt:
Merge? I'm not sure. Will I look into them and submit my review? Most likely

shadmansaleh:
>Merge? I'm not sure. Will I look into them and submit my review? Most likely

Ok . Then I think we should get through #55 first . Also #91 needs to get merged it's like 2 lines so wouldn't take time. also #92 is really small 10 lines You let me know what you think of that idea.

PR #54, opened 2021-01-08
Bug fix Theme not rendered properly

shadmansaleh:
I intruduced it with #47 
Every mode was being rendered as normal mode.


hoob3rt:
Well, my bad for not testing this pr properly, thanks for fixing it

Issue #21, opened 2021-01-04
[suggestion] Implement optional powerline aesthetic with < and > characters


Monklite:
In other status bars such as airline or lightline, there are options to implement powerline-like aesthetics like [this](https://i.redd.it/a2hrqr8ocbl31.png) using < and > keys. I would love to see this added, shouldn't be too difficult

coxackie:
I think this could be partly done by allowing to specify different left and right separators. One could then use glyphs from [here](https://github.com/ryanoasis/powerline-extra-symbols#glyphs) to do this.

PR #55, opened 2021-01-09
Added Powerline like > < separators

shadmansaleh:
solution for #21

It works, doesn't look bad .
Need to do some cleanup on the code.

Demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,17 +86,19 @@ All available themes are listed in [THEMES.md](./THEMES.md)
 Please create a pr if you managed to port a popular theme before me, [here is how to do it](./CONTRIBUTING.md).
 
 ### Changing separator in section
-Lualine defines a separator between components in given section, the default
-separator is `|`. You can change the separator this way:
+Lualine defines two kinds of seperators. One is for sections and other is for components. Default section seperators are '', '' and component separators are '', ''.
+They require powerline patched fonts. But you can easily change yours to something else like below
 
 ```lua
-lualine.separator = '|'
+lualine.section_separators = {'', ''}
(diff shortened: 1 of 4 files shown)


shadmansaleh:
I think it's quite stable now

shadmansaleh:
Does anyone know of any unicode symbles that can be used for this without patched fonts?

hoob3rt:
I think we should keep separators from https://github.com/hoob3rt/lualine.nvim/pull/55/commits/ef3dff46b492b2d5198f2c81159416433e7fe8e7 as default. They work with nerd fonts well while the current defaults crash on nerd fonts and display funny emojis with emoji theme installed

pic-window-210110-0036-00

hoob3rt:
> pic-window-210110-0036-00

Also there is a new visible bug, there should be no separator after `filaneme` or before `utf-8` here.

shadmansaleh:
>I think we should keep separators from ef3dff4 as default. They work with nerd fonts well while the current defaults crash on nerd fonts and display funny emojis with emoji theme installed

Funny naver thought they would look like this . Though it will be a small fix.

shadmansaleh:
> Also there is a new visible bug, there should be no separator after filaneme or before utf-8 here.

Why not ?

shadmansaleh:
changed to nerd font separators

shadmansaleh:
>Also there is a new visible bug, there should be no separator after filaneme or before utf-8 here.

I've realized they aren't found in powerline . I've removed them

shadmansaleh:
Closed in favor of #94.

PR #56, opened 2021-01-09
Added Completion Mode Label

jarviliam:
I noticed that when Neovim's autocomplete suggestion pops up while in Insertion Mode, the status would change to "ic".
So I just added basic labels for the remaining Complete-related modes.

https://neovim.io/doc/user/eval.html#mode()


hoob3rt:
Nice! It definitely looks better now:)

PR #57, opened 2021-01-10
Wider range of mode support

shadmansaleh:
basically gave all of them nice names


lua/lualine/components/mode.lua
```diff
@@ -2,20 +2,34 @@ local function mode()
   local mode_map = {
     ['__'] = '------',
```
hoob3rt: funny... I don't recall what this is supposed to do... nothing I guess? pretty sure it's safe to delete

shadmansaleh: I could't understand why it was there either.

Issue #50, opened 2021-01-08
Signify component opens in non git directories


hoob3rt:
Nice catch, this is also an issue with the `signify` component but I think these are two different bugs, I will try to fix this tomorrow.

_Originally posted by @hoob3rt in https://github.com/hoob3rt/lualine.nvim/issues/37#issuecomment-756176317_

PR #58, opened 2021-01-10
Fixed signify component apearing on non repo buffer

shadmansaleh:
fixing #50
[added, modified, removed] = [-1, -1, -1] was a error code that was
being ignored

I think we should use sy#repo#get_stats_decorated()
insted of sy#repo#get_stats()

### DIFF
--- a/lua/lualine/components/signify.lua
+++ b/lua/lualine/components/signify.lua
@@ -1,13 +1,6 @@
 local function signify()
-   if vim.fn.exists('*sy#repo#get_stats') == 0 then return '' end
-   local added, modified, removed = unpack(vim.fn['sy#repo#get_stats']())
-   if added == -1 then return '' end
-   local data = {
-    '+', added,
-    '~', modified,
-    '-', removed
-   }
-  return table.concat(data, ' ')
+   local ok, result = pcall(vim.fn['sy#repo#get_stats_decorated'])
+   if ok then return result else return '' end
 end
 
 return signify
--- a/lua/lualine/components/signify.lua
+++ b/lua/lualine/components/signify.lua
@@ -3,9 +3,9 @@ local function signify()
    local added, modified, removed = unpack(vim.fn['sy#repo#get_stats']())
    if added == -1 then return '' end
    local data = {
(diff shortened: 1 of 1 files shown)


lua/lualine/components/signify.lua
```diff
@@ -1,9 +1,7 @@
 local function signify()
    if vim.fn.exists('*sy#repo#get_stats') == 0 then return '' end
    local added, modified, removed = unpack(vim.fn['sy#repo#get_stats']())
-   if added == -1 then added = 0 end
-   if modified == -1 then modified = 0 end
-   if removed == -1 then removed = 0 end
+   if added == -1 then return '' end
```
hoob3rt, line 4: This indeed fixes the issue but could you explain why?

hoob3rt (commented):
> I think we should use sy#repo#get_stats_decorated()

This function sounds like something we are looking for but it does not seem to work on my system. Can you confirm you can get it working on yours?

lua/lualine/components/signify.lua
shadmansaleh: This it the get_stats function
```vim
function! sy#repo#get_stats(...) abort 
  let sy = getbufvar(a:0 ? a:1 : bufnr(''), 'sy')
  return empty(sy) ? [-1, -1, -1] : sy.stats
endfunction
```

shadmansaleh:
See if it works on your device

hoob3rt:
It does work but I don't like what `sy#repo#get_stats_decorated` returns. I think my implementation looks better and it's more consistent on what it shows. before 9ca50c8d7f4a06f16bb76ef0369b9f54aa38b733 it worked well, I think we should stick with that. 

Also it would be nice to change the formatting of the result. Currently on master it returns
```
+ 0 - 0 ~ 0
```

I think it would be nicer without the additional spaces (this is how it works on airline)
```
+0 -0 ~0
```

shadmansaleh:
Fixed

hoob3rt:
Yep, this works now. I had to add a comma you missed though :)

PR #59, opened 2021-01-10
Putting autocomands in a  group

shadmansaleh:
Wraping autocmd with augroup so multiple autocommands aren't created due to multiple calls to lualine.status() .

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -109,9 +109,12 @@ end
 local function exec_autocommands()
   _G.set_lualine_theme = set_lualine_theme
   _G.set_active_statusline = statusline
+  vim.cmd([[augroup LuaLine]])
+  vim.cmd([[autocmd!]])
   vim.cmd([[autocmd WinEnter,BufEnter * setlocal statusline=%!v:lua.set_active_statusline(1)]])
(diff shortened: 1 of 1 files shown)


hoob3rt:
It's a good idea but why is https://github.com/hoob3rt/lualine.nvim/pull/59/commits/93137d10dc9d3babfa8d1c8bdbc8595322dc08d0 in this pr?

shadmansaleh:
>It's a good idea but why is 93137d1 in this pr?

My bad I probably branched of wrong branch . I'll rebase it on master . I think it should solve it.

shadmansaleh:
fixed

lua/lualine.lua
```diff
@@ -109,9 +109,12 @@ end
 local function exec_autocommands()
   _G.set_lualine_theme = set_lualine_theme
   _G.set_active_statusline = statusline
+  vim.cmd([[augroup LuaLine]])
```
hoob3rt, line 112: any reason why you are using PascalCase instead of snake_case here?
I think it should simply be `lualine`

shadmansaleh: My bad . Theres no reson . I have a habit of using pascal cases for augroups . I'll fix it.

shadmansaleh: fixed

Issue #45, opened 2021-01-07
additional setting for components


hoob3rt:
as mentioned in https://github.com/hoob3rt/lualine.nvim/issues/16#issuecomment-753699274 there should be additional settings for lualine's components. Something like

* icon
* padding
* uppercase/lowercase

Let me know if you wish to have anything else supported.

deevus:
Full path for `filename` component would be appreciated. I've found myself missing that.

bobicino:
For full file path, it would be great if it had an option to use `pathshorten` if the window size get lower than a defined value.
This is what I have been using so far:
```
local function filepath()
  local path = vim.fn.expand('%')
  if vim.fn.empty(path) == 1 then return '' end
  local squeeze_width  = vim.fn.winwidth(0) / 2
  if squeeze_width <= 42 then
    path = vim.fn.pathshorten(path)
  end
  return path
end
lualine.sections.lualine_c = { filepath }
```

shadmansaleh:
>For full file path, it would be great if it had an option to use pathshorten if the window size get lower than a defined value.-

good idea i'll add it to #60. Maybe we should see the lenght of file path too for shorting it .
By the way should we have relative filepath or complete filepath from root for full filename . What do you guys think?

bobicino:
I have always used relative path. I find it more convenient since you are, in most cases, working in a project folder (maybe even set with `vim-rooter`).
But since we are discussing about options, it would be cool having both.. the only difference would be `expand('%:p')` instead of `expand('%')`.
Also looking for the string lenght is a great idea! 😊

shadmansaleh:
>But since we are discussing about options, it would be cool having both.. the only difference would be expand('%:p') instead of expand('%')

I've added a relative option that can be set to false to have full_path from root otherwise it will show relative path . Also added path shortning.

mellery451:
I tried making sense of the comments/links here, but can't quite follow. What's the best way currently to display a root-relative filepath instead of just filename?

PR #60, opened 2021-01-10
Adding component specific options

shadmansaleh:
### Add basic framework for component specific options

Addressing #45

- M.opt configs are global for all components
- component.args are component specific . They are passed to init
  function of component
- componemts.others can be interpreted in draw_section function

filename, filetype and padding are given as POC .
Added no_modified option for filename . [To not show modified status]
icons option for filetype show or not show filetype

padding is implemented in draw_section function

config example:
```lua
lualine.sections = {
  lualine_c = { {'filename', args = {no_modified = true}, padding = 2}},
  lualine_x = { 'encoding', 'fileformat', {'filetype', args = {icons=true} } },
}
```
We can discuss what options to implement and how to implement here . I'll later add component specific highlighting if I can


hoob3rt:
I will try to take a look at this on thursday. But your example looks great!!! This is the exact functionality I had in mind.

shadmansaleh:
Now args table is not required for configuring . The entire configuration table is sent to component.init() and draw sections function.

Now configuration should be 
```lua
lualine.sections = {
  lualine_c = { {'filename', no_modified = true, padding = 2}},
  lualine_x = { 'encoding', 'fileformat', {'filetype', icons=true } },
}
```

shadmansaleh:
Is there any way to get values of guifg,guibg,ctermbg,ctermfg from a highlight group name. If I run :highlight group_name it prints those values . But is there any function that returns them?

hoob3rt:
> If I run :highlight group_name it prints those values . But is there any function that returns them?

```lua
local color = vim.api.nvim_command_output([[execute 'hi lualine_a']])
```

shadmansaleh:
Now components can color themselves . I've added signify component as a POC.

shadmansaleh:
### Currently available options

#### Global options
- icons (Default: true)
- powerline_fonts (true)

### Customizations
#### Global 
- padding (0)
- left_padding (0)
- right_padding (0)

#### Component specific
- filename
  - modify (true)
- fileformat
  - icons (true)
- filetype
  - icons (true)
- signify
  - colored (true)

hoob3rt:
> ### Currently available options
> #### Global options
> 
>     * icons (Default: true) [suported by branch,fileformat,filetype]
> 
>     * powerline_fonts (true) [suported by branch, fileformat]

I think this should be one option, if we are supporting powerline icons then they should be used everywhere where it's possible. 

>     * ~signify~ (Branched of to #61)
>       
>       * ~colored (true)~

Yeah thanks for splitting these two pr's :)

shadmansaleh:
>I think this should be one option, if we are supporting powerline icons then they should be used everywhere where it's possible.

do you mean to remove powerline_icons or icons?

hoob3rt:
> > I think this should be one option, if we are supporting powerline icons then they should be used everywhere where it's possible.
> 
> do you mean to remove powerline_icons or icons?

Yeah imo there should be only icons option which uses powerline icons for everything except filetype while also using plugin(if available) to display filetype icons

btw I think we should document these icons better to make sure they are indeed from nerdfonts

replace
```lua
local icon =  ''
```
with
```lua
local icon = '' -- e725
```

lua/lualine/components/branch.lua
```diff
@@ -29,22 +29,22 @@ timer:start(0, 1000, vim.schedule_wrap(function()
...
-  local ok,devicons = pcall(require,'nvim-web-devicons')
-  if ok then
-    local icon = devicons.get_icon('git')
-    if icon ~= nil then
-      return icon .. ' ' .. git_branch
+local function branch(args)
+  local icon = '' -- e0a0
+  if args.icons_enabled == false then
+    icon = nil
+  elseif args.icons_enabled == nil and args.opt.icons_enabled == false then
```
hoob3rt, line 36: This is bad imo, why have both `args` and `args.opt`? 

There should only be one parameter here `options` or `opts`. The options should be parsed when loading the component. Something like
```
component.options = M.options
for name, value in (component options array)
  component.options.name = value
```

This will make components more clear imo. I will look into this pr more when this is updated.

shadmansaleh: Theres args.opt.icons_enabled and args.icons_enabled . Brecause args.opt is global options. and args is component specific options .

Users can set ```lualine.opt.icons_enabled = false```
like users set ```lualine.seperator = '|'```
to not have icons in any components insted of having to set it on every component . And then can override that global option in components. 
Like if users wants icons in filetype and nowhere else they can do something like
```lualine.section_c = { {'filetype', icons_enabled = true}}```

Also not all options are processed in compinents . global options like padding is handled when components are drawn in util.lua

Comment (hoob3rt) [reply lua/lualine/components/branch.lua]: > Theres args.opt.icons_enabled and args.icons_enabled . Brecause args.opt is global options. and args is component specific options .

Well, what exactly is the benefit of having them both separately? I cannot see one. They should be parsed when loading the component for sure.

> Users can set `lualine.opt.icons_enabled = false`
> like users set `lualine.seperator = '|'`
> to not have icons in any components insted of having to set it on every component . And then can override that global option in components.
> Also not all options are processed in compinents . global options like padding is handled when components are drawn in util.lua

This is inconsistent as hell. First of all why are some options in `opt` and others are not? Secondly I don't see a reason why padding cannot be a component option, it surely can.

Imo all options should be in the `opt` object ( I would personally rename it to options though).

This way the user can set everything through `lualine.options`

The component loading function should be aware (probably hard coded) which of these global options to pass to components (icons should be passed for example) and update them if the given component has an overriding option

Comment (shadmansaleh) [reply lua/lualine/components/branch.lua]: I'm probably not making my self clear.
Here's my lualine conf.

```lua
vim.cmd("packadd lualine.nvim")
vim.cmd("packadd nvim-web-devicons")
local lualine = require('lualine')

vim.o.showmode = false
lualine.separator = '|'
lualine.opt.icons_enabled = true

lualine.sections = {
  lualine_a = { {'mode', lower=false} },
  lualine_b = { {'branch', icons_enabled=true}, {'signify', colored=true}}, 
  lualine_c = { {'filename', modified=true, padding = 1, full_name=true, relative=true}},
  lualine_x = { 'encoding', {'fileformat', icons_enabled=true}, {'filetype', icons_enabled=true }},
  lualine_y = { 'progress' },
  lualine_z = { {'location', icons_enabled=false}  },
  lualine_diagnostics = {  }
}

lualine.status()
```

opt table is not something user will set in components. 
Also most of the options i've set here are in default values so they aren't required to be mentioned. I've put them for testing purposes.

>Well, what exactly is the benefit of having them both separately? 

They allow user to set an option globally . The options that effect more than one component like `icons_enabled`

> Secondly I don't see a reason why padding cannot be a component option, it surely can.

Options like padding, lower, upper are same for all components no point in implementing the samething on every component.

>This way the user can set everything through lualine.options

That will be like current opt table . In that case users wonn't be able to set different value for an option for different components.

>The component loading function should be aware (probably hard coded) which of these global options to pass to components (icons should be passed for example) and update them if the given component has an overriding option

please elaborate . I couldn't understand this part.

If you haven't already please look at the entire thing when you have time .

hoob3rt: I understand that everything works well now, I have tested it on my system. I'm talking about an implementation... I tried my best to explain it, I will try to clean this up myself soon and you will be able to see what I meant :)

shadmansaleh: Alright .

shadmansaleh:
Have you been talking about this ?

hoob3rt:
> Have you been talking about this [844daa4](https://github.com/hoob3rt/lualine.nvim/commit/844daa4677f301b55b8c7c28001d62651128c995) ?

I've seen this and this is a step in the right direction. I have been tinkering with this pr a lot and have a general idea on how to make it cleaner. I will commit the changes here (tomorrow hopefully) and we will be able to proceed further

shadmansaleh:
How about taking a modifier function as a global option ?
It will make it easy  to change formats of components .

It can be like this in utils.draw_section()
```lua
 for _, component in pairs(section) do
    local localstatus = component[1]()
    if #localstatus > 0 then
      if args.modifier then localstatus = args.modifier(localstatus)
      localstatus = set_case(localstatus, component)
      localstatus = set_padding(localstatus, component)
      table.insert(status, localstatus)
    end
  end
```

shadmansaleh:
I think upper and lower optuon can be removed as  the same affect can be achived by setting `format=string.upper` and `format=string.lower`

pianocomposer321:
Any update on the status of this PR?

hoob3rt:
> Any update on the status of this PR?

I will resolve all available prs and implement diagnostics next week. Exams right now 😡

pianocomposer321:
> > Any update on the status of this PR?
> 
> I will resolve all available prs and implement diagnostics next week. Exams right now 😡

Just checking back, wondering if any progress has been made.

lua/lualine/components/location.lua
```diff
@@ -1,6 +1,30 @@
-local function location()
-  local data = [[%3l:%-2c]]
-  return data
+local function location(args)
+  -- set when user wants to set a custom icon
+  local icons_enabled = args.icons_enabled
+
+  local icon_line_no = '' -- e0a1
+  local icon_col_no  = ''  -- e0a3
```
hoob3rt, line 6: I don't like this, messes up the consistency of all components.

hoob3rt, line 6: I don't like this, messes up with the consistency of all components, there should be only one icon per component imo. If user wants further customization they can create their own lua function

lua/lualine.lua
```diff
@@ -8,6 +8,10 @@ local theme_set = {}
 
 M.separator = '|'
 
+M.opt = {
+  icons_enabled = true,
```
hoob3rt, line 12: if we are going down the options route then all options should be placed in `opt` table (theme and separator as well)

lua/lualine.lua
```diff
@@ -65,17 +69,54 @@ local function load_special_components(component)
...
+local function component_loader(component)
+  if type(component[1]) == 'function' then return component end
+  -- apply default args
+  for opt_name, opt_val in pairs(M.opt) do
+    if component[opt_name] == nil then
+      component[opt_name] = opt_val
+    end
+  end
+  -- set custom highlights
+  if component.color then
```
hoob3rt, line 81: `color` is misleading. You are not really changing one color. Should be called `theme` imo. With this change there should also be an additional check here
```lua
if component.theme and component.theme ~= M.options.theme
```
I really like this idea overall

lua/lualine.lua
```diff
@@ -65,17 +69,54 @@ local function load_special_components(component)
...
         end
+        component.self = {}
+        -- used to provide default for bg and fg for custom highlights
+        component.self.theme = theme_set
+        -- setting highlight because utils need highlight but cann't require
+        -- it as it creates circular dependency . It's a workaround for now
+        -- Should look for a better solution.
+        component.self.highlight = highlight
+        -- convert lualine_section -> section before setting section
+        component.self.section = section_name:sub(9,10)
```
hoob3rt, line 117: I don't quite get what's going on here.

hoob3rt, line 117: I don't really get what's going on here. This comment doesn't say much to me unfortunately

lua/lualine/components/branch.lua
```diff
@@ -59,31 +59,29 @@ local function watch_head()
...
 
--- returns the git_branch value to be shown on statusline
-local function branch()
-  if not git_branch or #git_branch == 0 then return '' end
-  local ok,devicons = pcall(require,'nvim-web-devicons')
-  if ok then
-    local icon = devicons.get_icon('git')
-    if icon ~= nil then
-      return icon .. ' ' .. git_branch
+local function branch(args)
```
hoob3rt, line 62: This is an issue in all components. If it's called `options` in `lualine.lua` then why use `args` here?

lua/lualine/components/branch.lua
```diff
@@ -59,31 +59,29 @@ local function watch_head()
...
 end
 
 -- run watch head on load so branch is present when component is loaded
 watch_head()
 
--- TODO Don't export as a global function
-_G.lualine_branch_update = watch_head
 -- update branch state of BufEnter as different Buffer may be on different repos
-vim.cmd[[autocmd BufEnter * call v:lua.lualine_branch_update()]]
+vim.cmd[[autocmd BufEnter * lua require'lualine.components.branch'.lualine_branch_update()]]
```
hoob3rt, line 82: I don't know what's the performance difference but calling a global seems more optimal than using `require` on every buffer change

lua/lualine/components/filename.lua
```diff
@@ -1,6 +1,31 @@
...
+      data = vim.fn.expand('%:p')
+    end
+    if data == '' then
+      data = '[No Name]'
+    elseif vim.fn.winwidth(0) <= 84 or #data > 40 then
+      data = vim.fn.pathshorten(data)
+    end
+
+    if modified then
+      if vim.bo.modified then data = data .. "[+]"
```
hoob3rt, line 24: This is redundant. By this I mean `[+]` etc. Vim already has a char for that, you can check master for reference. Also this variable shouldn't be called modified but `fileinfo` or something like that

lua/lualine/components/filename.lua
```diff
@@ -1,6 +1,31 @@
-local function filename()
-  local data = [[%t %m]]
-  return data
+local function filename(args)
+  -- setting defaults
+  local modified, full_name, relative = true, false, true
```
hoob3rt, line 3: `full_name` and `relative` don't describe well what's happening here. My first thought was that `full_name` is the same as `full_path`, which is exactly opposite of what `relative` means.Should probably be named `shorten` and `full_path` respectively.

hoob3rt (commented):
Okay, sorry it took so long. Last week I wasted a bunch of time trying to do some smart shinenigans with oop in lua and ultimately found it ridiculous and not worth the time. We shall continue with your implementation, I've got time now so hopefully this will get resolved by the end of this week. 

I will mention one thing here and not in review comments. There is a lot of the same code in components. `args.icons_enabled`, `args.icon`.  No need to have it imo. We could create a general component function(this is what I was trying to achieve in oop). Here is what I have in mind, I will provide fileformat component as an example
```lua
local function fileformat(options)
  local icon_linux  = "" -- e712
  local icon_windos = "" -- e70f
  local icon_mac    = "" -- e711
  return function()
    local format = vim.bo.fileformat
    local icon = ''
    if options.icons_enabled then
      if     format == 'unix' then icon = icon_linux
      elseif format == 'dos'  then icon = icon_windos
      elseif format == 'mac'  then icon = icon_mac end
    end
    return component_creator(format, icon, options)
  end
end
```
If we go down this route we should also export `padding`, `themes` etc to component_creator. Lmk what you think

lua/lualine/components/location.lua
shadmansaleh: How about splitting them in 2  components? what common icon might line no or column no have?

lua/lualine.lua
shadmansaleh: That's a good idea .

lua/lualine.lua
shadmansaleh: >You are not really changing one color. Should be called theme imo.

We actualy might be changing just one color if user provides only fg or bg . I think theme is to broad an idea for this also probably will create confusion with M.theme . They are not the same theme . This option doesn't let you pick colors from other themes that many might think.

>```if component.theme and component.theme ~= M.options.theme```

I'm realy confused by this . What's it suppose to do?

lua/lualine.lua
shadmansaleh: Basicaly we are already requireing utils from highlight but I needed highlight in a function in utils . It's just a hacky workaround . Will probably have to reorganize some functionsbto properly resolve it . I thought it was out of scope of this pr  . By the way what do you think of moving logic in itils.lua related to noterguicolors to separate file?

lua/lualine/components/filename.lua
shadmansaleh: ok

lua/lualine/components/filename.lua
shadmansaleh: >By this I mean [+] etc. Vim already has a char for that, you can check master for reference.

I couldn't really understand what you ment by this . if you're talking about '%m' the only reson I've implemented it in lua so genaral upper and lower works.

> Also this variable shouldn't be called modified but fileinfo or something like that

How about file_status insted?

lua/lualine/components/branch.lua
shadmansaleh: As far as I know lua require only runs code once and caches the output . So there shouldn't be any performace drop.

checkout `:lua print(vim.inspect(package.loaded))`

lua/lualine/components/branch.lua
shadmansaleh: >This is an issue in all components. If it's called options in lualine.lua

is it ? I've greped the project and couldn't find word options in any code ? This args is the options user gave while declaring component not the M.opt table

lua/lualine.lua
shadmansaleh: I was croping `lualine_a` to `a` to use it as a key to lookup theme table. It doesn't make much sense here . I'll move it to highlight.

shadmansaleh:
> We could create a general component function(this is what I was trying to achieve in oop). 

This seems like a good idea I'll look into it. If you have something on that oop idea. Can you push it to a separate branch I'd like to take a look.

shadmansaleh:
Can you tell me whats the purpose of theme_set veriable in lualine.lua . I donn't see it being used anywhere...

shadmansaleh:
I've now implemented a genaral logic for icons and removed duplicate code from component regarding icons .


>If we go down this route we should also export padding, themes etc to component_creator.

I donn't wana go in that route . Currently user created components can also use global options . Like if I'm using a veriable or a function I can still set the highligjts or padding with options from lualine . If we move those logic to component_create function then that wonn't be possible. I think it is better to support those too. Also these global options are genaral enough that it will work on anything :)

lua/lualine/components/location.lua
shadmansaleh: Removed icons from location .

lua/lualine.lua
shadmansaleh: Done

lua/lualine.lua
shadmansaleh: moved to highlight where it's relevent

lua/lualine/components/branch.lua
shadmansaleh: renamed args to options, M.opt to M.options and settings in utils.lua to options

lua/lualine/components/filename.lua
shadmansaleh: renamed modified to file_status, full_name to full_path

shadmansaleh:
@hoob3rt what do you think about moonscript ?
It's really easy to do oop in moon . The syntax is mostly lua but without some annoyences like no more local, then, end . It has += oparators, list comprihension, splicing, default args, switch, continue.  and way better oop support .
I definatly like writting moonscript then lua .  I donn't know why it isn't more popular then lua.

If you're interested I've converted lualine to moonscript in moonscript_conversion [branch](https://github.com/shadmansaleh/lualine.nvim/tree/moonscript_conversion) in my repo . NOTE: I'm didn't really convert it to get merged .I was checking out moonscript and thought converting this will be a nice exercise as I know the codebase.

If you're ever interested in moonscript these were quite usefull
- [learn x in y minutes moonscript](https://learnxinyminutes.com/docs/moonscript/)
- [reference](https://moonscript.org/reference/)
- [nvim-moonmaker](https://github.com/svermeulen/nvim-moonmaker)
- [moonscript-vim](https://github.com/leafo/moonscript-vim)

lua/lualine/components/branch.lua
hoob3rt: Nice, I didn't know that

lua/lualine.lua
hoob3rt: I think it's actually better with `theme`. Since user will have to remember only one variable name. I think component theme and component color are both perfectly readable. 

> ```
> if component.theme and component.theme ~= M.options.theme
> ```
> I'm realy confused by this . What's it suppose to do?

Well, if we rename it to theme then we will produce a slight issue. In line 78 we copy global options and if we try to update component's theme with global theme it will break

lua/lualine.lua
hoob3rt: >  I thought it was out of scope of this pr

No, we should do it here. No need to create a mess on master

> By the way what do you think of moving logic in itils.lua related to noterguicolors to separate file?
We should probably separate the whole utils/highlights/component functions better. Lmk what you have in mind. Imo there should be something like `global_utils`, `component_utils`, `highlight` and notermguicolors in separate file as well. These names are placeholders just so you get an idea

lua/lualine/components/filename.lua
hoob3rt: > > By this I mean [+] etc. Vim already has a char for that, you can check master for reference.
> 
> I couldn't really understand what you ment by this . if you're talking about '%m' the only reson I've implemented it in lua so genaral upper and lower works.

Fair enough. 

> > Also this variable shouldn't be called modified but fileinfo or something like that
> 
> How about file_status insted?

Yea sounds good to me

lua/lualine/components/filename.lua
hoob3rt: I looked at this 2 days ago and I already don't know what does what. Please come up with better naming scheme.

hoob3rt:
> Can you tell me whats the purpose of theme_set veriable in lualine.lua . I donn't see it being used anywhere...

Yep, sure. This variable is supposed to enable changing themes on the fly. But... you actually bricked it recently haha my bad for not noticing it. Check [this](https://github.com/hoob3rt/lualine.nvim/commit/e37b3ac92944d2562827b457c81994384c4ba274#diff-921f7a7fb67331a59c9f0c0b903088920375643e4a496e60283145818e7a23db) out. We should restore this if statement

lua/lualine.lua
shadmansaleh: >Well, if we rename it to theme then we will produce a slight issue. In line 78 we copy global options and if we try to update component's theme with global theme it will break

Well then that's another problem with theme .

I really donn't like the name theme . theme seems completely different than whats happening here and users will definatly expact theme names to work which wonn't

hoob3rt:
> I've now implemented a genaral logic for icons and removed duplicate code from component regarding icons .
> 
> > If we go down this route we should also export padding, themes etc to component_creator.
> 
> I donn't wana go in that route . Currently user created components can also use global options . Like if I'm using a veriable or a function I can still set the highligjts or padding with options from lualine , also the components like mode that we haven't touched can utilize the global options too. If we move those logic to component_create function then that wonn't be possible. I think it is better to support those too. Also these global options are genaral enough that it will work on anything :)

Good points. I agree

hoob3rt:
> @hoob3rt what do you think about moonscript ?
> It's really easy to do oop in moon . The syntax is mostly lua but without some annoyences like no more local, then, end . It has += oparators, list comprihension, splicing, default args, switch, continue. and way better oop support .
> I definatly like writting moonscript then lua . I donn't know why it isn't more popular then lua.
> 
> If you're interested I've converted lualine to moonscript in moonscript_conversion [branch](https://github.com/shadmansaleh/lualine.nvim/tree/moonscript_conversion) in my repo . NOTE: I'm didn't really convert it to get merged .I was checking out moonscript and thought converting this will be a nice exercise as I know the codebase.
> 
> If you're ever interested in moonscript these were quite usefull
> 
> * [learn x in y minutes moonscript](https://learnxinyminutes.com/docs/moonscript/)
> * [reference](https://moonscript.org/reference/)
> * [nvim-moonmaker](https://github.com/svermeulen/nvim-moonmaker)
> * [moonscript-vim](https://github.com/leafo/moonscript-vim)

This not a topic for this pr. But tldr I can tell you that I don't want to switch from lua since it's natively supported and after giving it some thought I don't want to go oop route(for now). If you want to continue this thread then reference this in a new issue

shadmansaleh:
>But... you actually bricked it recently haha my bad for not noticing it. Check this out. We should restore this if statement

Oh I didn't notice it . My bad . Will restore it . But does it actually work?

hoob3rt:
> > But... you actually bricked it recently haha my bad for not noticing it. Check this out. We should restore this if statement
> 
> Oh I didn't notice it . My bad . Will restore it . But does it actually work?

Yes it does. You can do
```
:lua require('lualine').theme = 'name'
```
To change theme at runtime

lua/lualine.lua
hoob3rt: Oh and all of that in a new utils directory of course

shadmansaleh:
>This not a topic for this pr. But tldr I can tell you that I don't want to switch from lua since it's natively supported and after giving it some thought I don't want to go oop route(for now). If you want to continue this thread then reference this in a new issue

Not really I wasn't going to merge it anyway . It was just a casual question  moonscript gets compiled to lua . And compilation can be done on dev's machine and compiled lua files will be exexited in users device. So It's not really unsupported as lua is what runs .

lua/lualine.lua
hoob3rt: And I'm not a fan of `color` but I guess I will let it slide for now. Let's keep `color`.

shadmansaleh:
>Yes it does. You can do

> `:lua require('lualine').theme = 'name'`
To change theme at runtime

Ok you can change it back or I'll open a pr soon

lua/lualine.lua
shadmansaleh: I do agree . utils is kind of a mess . Every thing that didn;t fit there ended up in these :)

lua/lualine/components/filename.lua
shadmansaleh: what's better than full_path & relative ?

hoob3rt:
> > Yes it does. You can do
> 
> > `:lua require('lualine').theme = 'name'`
> > To change theme at runtime
> 
> Ok you can change it back or I'll open a pr soon.
> I've broken my local repo . will have to fix it first.

I've fixed it at master already

shadmansaleh:
>I've fixed it at master already

ohh I didn't notice it. 

@hoob3rt can we review the prs in paralal . Since most of the time you and I aren't online at same time . If you put reviews I can fix them when I'm available . It will be faster than going though prs one by one . What do you say ? I'll refactor the utils.lua . Also let me know if you have other issues.

lua/lualine.lua
shadmansaleh: Done.

lua/lualine/components/filename.lua
shadmansaleh: Ok my bad . At first I thought you ment
full_name -> full_path & relative -> shorten.

But now I realize you ment fullname -> shorten & relative -> full_path.

This does make sence I've renamed them accordingly.

shadmansaleh:
The above issues have been resolved see if you can find any other else we can merge this

hoob3rt:
Okay, good progress. I like what you did with all the components. However, there is a bug in your code. Minimal example to reproduce:
```
local function test()
  return 'a'
end
lualine.sections.lualine_a = {test}
```

lua/lualine/components/filename.lua
hoob3rt: yep, that's exactly what I meant, thanks

hoob3rt:
> @hoob3rt can we review the prs in paralal 

If only I could... I will try to review all of them on monday

shadmansaleh:
>I like what you did with all the components. However, there is a bug in your code. Minimal example to reproduce:

Ok I'll check it out.

>Also, please clean up trailing whitespaces. I don't know why that's still an issue, you can set up vim to clear them out automatically.

Weird I did . I'll fix them. I think the problem is I cann't get sumnako_lua_lsp to work :( . It just does nothing .

shadmansaleh:
fixed.

shadmansaleh:
>If only I could... I will try to review all of them on monday

This is the only big pr other ones are small . Reviewing them wonn't take much time . That's why I had asked you to go through them but you chose this .

hoob3rt:
I would rename `utils/highlight.lua` to `utils/cterm_colors.lua` or something along those lines. Other than that lgtm

shadmansaleh:
>I would rename utils/highlight.lua to utils/cterm_colors.lua or something along those lines. Other than that lgtm

Renamed . Can you merge #91 it's a hotfix with 2LOC that has been delayed.

hoob3rt:
I've been using this for the past few days and it seems to be stable. Thanks for you work :)

PR #61, opened 2021-01-11
Add colors to signify

shadmansaleh:
### This branch is branched from #60 so it should be merged after #60 is merged.

Now Components can apply currect highlights for themselves . I've added colors to signify component. 

Demo

### it's WIP until #60 is merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,15 @@ Default options act as default for all components
 - fileformat
   - icons_enabled (true)\
    Whether to displays icon before component
+- signify
+  - colored        (true)\
+    Whether to show colors
+  - color_added    ('#90ee90')\
+    Foreground color of added section
+  - color_modified ('#f0e130')\
+    Foreground color of modified section
+  - color_removed  ('#ff0038')\
+    Foreground color of removed section
 
 **Example:**
 ```lua
--- a/README.md
+++ b/README.md
@@ -246,7 +246,9 @@ Default options act as default for all components
    Whether to displays icon before component
 - signify
   - colored        (true)\
-    Whether to show colors
+    Whether to show colors. Colors are automaticaly
+		extracted from colorscheme . If you want to change
+		any of those you can use options given below.
   - color_added    ('#90ee90')\
     Foreground color of added section
   - color_modified ('#f0e130')\
--- a/lua/lualine/components/signify.lua
+++ b/lua/lualine/components/signify.lua
@@ -8,40 +8,28 @@ local default_color_removed  = "#90ee90"
 local default_color_modified = "#ff0038"
 
 local function get_highlight(scope, color_group)
-  local color = vim.api.nvim_command_output([[ execute 'hi ]] .. color_group .. [[']])
-  if color:match([[links to]]) then
-    local last
-    for w in  string.gmatch (color, "%w+") do
-      last = w
-    end
-    if last then return get_highlight(scope, last) end
-  end
-  if color:match( [[gui=reverse]] ) then
-    if scope == 'guifg' then scope = 'guibg' else scope = 'guifg' end
-  end
-  color = color:match( scope .. [[=(#?%w+)]]) or nil
+  local color = string.format('#%06x', vim.api.nvim_get_hl_by_name(color_group, true)[scope])
   return color or nil
 end
 
 local function signify(options)
-  local colored = true
-  if options.colored ~= nil then colored = options.colored end
+  if options.colored == nil then options.colored = true end
   local color_added, color_modified, color_removed
   -- apply colors
   if options.color_added then
     color_added = options.color_added
   else
-    color_added = get_highlight('guifg', 'diffAdded') or default_color_added
+    color_added = get_highlight('foreground', 'diffAdded') or default_color_added
   end
   if options.color_modified then
     color_modified = options.color_modified
   else
-    color_modified = get_highlight('guifg', 'diffChanged') or default_color_modified
+    color_modified = get_highlight('foreground', 'diffChanged') or default_color_modified
   end
   if options.color_removed then
     color_removed  = options.color_removed
   else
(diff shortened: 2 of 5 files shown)


shadmansaleh:
#### Options for signify

- colored (default: true)
- color_added      (#90ee90)
- color_modified (#f0e130)
- color_removed (#ff0038)

lua/lualine/components/signify.lua
```diff
@@ -1,33 +1,72 @@
...
-   end
-end
+local utils = require'lualine.utils.utils'
+
+local function signify(args)
+  local colored = true
+  if args.colored ~= nil then colored = args.colored end
+
+  -- apply defaults
+  local color_modified = "#f0e130"
```
hoob3rt, line 11: Wouldn't it be better to have
```
if color_modified ==nil then color_modified = '%#DiffAdded`
```
etc. Most themes already define diff colors. I think having hardcoded one is bad.

hoob3rt (commented):
Okay this looks like a merging hell. I don't really know that to look at here but I have one input as of right now

lua/lualine/components/signify.lua
shadmansaleh: Good idea . Using colorscemes highlight didn't come to my mind

shadmansaleh:
>Okay this looks like a merging hell. I don't really know that to look at here but I have one input as of right now. Also the pr name is different from branch name. This is supposed to be only signify color option right?

Oviously it requires a rebase on master . after rebase merging hell no more. yes just signify . basic fuctionality for colors was added to master with #60 in color options . This was basicaly suppose to be a demo of how custom highlights can be used inside components + some nicer looks for signify :)

lua/lualine/components/signify.lua
shadmansaleh: Ok I've looked in it and I donn't like it.
First those highlight groups background are based on colorschemes background so they donn't blend with statuslines theme (looks oweful actualy). But when we create highlight groups it is created based on satuslines
theme so it blends with the background .
Secondly if the colorscheme doesn't support cterm colors our support for cterm colors goes out the window too.
Also I think it added unnesery complexcity as if we default to those hl_groups we might have to create some groups and not really on colorcheme for others . 

Also since colors are fully customizable haveing defaults hardcoded isn't that big of a deal in my opinion.

hoob3rt: thanks for pointing that out. I've pushed a solution to this problem. I think it works well now

lua/lualine/components/signify.lua
```diff
@@ -1,33 +1,73 @@
...
-       return table.concat(result, '')
-   else
-       return ''
-   end
-end
+local utils = require'lualine.utils.utils'
+
+local function signify(options)
+  local colored = true
+  if options.colored ~= nil then colored = options.colored end
```
hoob3rt, line 28: I would change this to look cleaner. 
```lua
if options.colored == nil then options.colored = true end
```
and work with options variables. Similar to all other option variables. 
This has two benefits
* while looking at the code you instantly know that `if options.colored` refers to an option. `if colored` can be a local variable
* no additional memory usage

lua/lualine/components/signify.lua
shadmansaleh: I think you're extracting fg colors . checkout #94 highlight.lua in get transitional_highlights function I thunk thats a better approch than trying to parse output.

lua/lualine/components/signify.lua
shadmansaleh: good idea . I'll change that

lua/lualine/components/signify.lua
hoob3rt: If you know how to improve it then go ahead. I knew there was a vim.api command to get highlight but forgot to do it here :neutral_face:

shadmansaleh: I'll do it . I was just passing the idea :)

lua/lualine/components/signify.lua
```diff
@@ -1,33 +1,89 @@
...
-       return ''
-   end
+local utils = require'lualine.utils.utils'
+
+local default_color_added    = "#f0e130"
+local default_color_removed  = "#90ee90"
+local default_color_modified = "#ff0038"
+
+local function get_highlight(scope, color_group)
+  local color = string.format('#%06x', vim.api.nvim_get_hl_by_name(color_group, true)[scope])
```
hoob3rt, line 11: ok lol, next time I will defo search for vim.api functions instead of implementing a mess myself :laughing:

lua/lualine/components/signify.lua
```diff
@@ -1,33 +1,89 @@
...
+  return color or nil
 end
 
+local function signify(options)
+  if options.colored == nil then options.colored = true end
+  local color_added, color_modified, color_removed
```
hoob3rt, line 17: This is also redundant. We can work on only `options.colors_added` etc

shadmansaleh: We definatly can

shadmansaleh:
We probably should make the color extracting function a  utils function.

hoob3rt:
> We probably should make the color extracting function a utils function.

Yep, we can. This should be the last thing in this pr.

README.md
```diff
@@ -244,6 +244,15 @@ Default options act as default for all components
 - fileformat
   - icons_enabled (true)\
    Whether to displays icon before component
+- signify
+  - colored        (true)\
+    Whether to show colors
```
hoob3rt, line 249: oh and readme should imply that when this option is set the colors are automatically set from current vim colorscheme. Same change to docs

shadmansaleh:
Ok I think now this can be merged....

README.md
shadmansaleh: Done

hoob3rt:
Yep, lgtm now. Thanks

PR #62, opened 2021-01-12
Add molokai theme

shadmansaleh:
ported from lightline


hoob3rt:
The screenshots could be better but I alrady created a script to update all screenshots to be nicer so I will do it later.
lgtm otherwise

PR #64, opened 2021-01-12
Adding wombat theme

shadmansaleh:
ported from lualine


shadmansaleh:
I'll add screenshots THEMES.md later

shadmansaleh:
Should I give screenshots based on current master or from a version where #55 and #60 are merged?

hoob3rt:
> Should I give screenshots based on current master or from a version where #55 and #60 are merged?

Obviously the current version. We'll take care of powerline symbols later

shadmansaleh:
If you happen to check them out . You can tell me if any color needs changing before I take the screenshots. I'm basically using afew macros to take color values from lightlines vimscript files.

hoob3rt:
> If you happen to check them out . You can tell me if any color needs changing before I take the screenshots. I'm basically using afew macros to take color values from lightlines vimscript files.

Sure I will check them out in detail the day after tomorrow.

> Plus tell me if any theme on your list has a higher priority

I think we should support all themes with over 800 stars including all their versions. (ayu for example has 3 versions). This would cover 95% of people imo, others could port their own theme or use `auto` when it's ready. You can check the top vim themes [here](https://vimcolorschemes.com/top)

PR #65, opened 2021-01-12
Adding material theme

shadmansaleh:
Ported from lightline


coxackie:
Could you also do gruvbox-material? The one [here](https://github.com/sainnhe/gruvbox-material/blob/master/autoload/lightline/colorscheme/gruvbox_material.vim)?

shadmansaleh:
>Could you also do gruvbox-material? The one here?

That one looks like a mess . I might try it later . Why not you give it a try till then..

coxackie:
> That one looks like a mess . I might try it later . Why not you give it a try till then..
Leave it if it is too complex. I thought you had a script of something that helped in this. At any rate, I can always use the "auto" option when it gets implemented. For the time being `gruvbox` is close enough to `gruvbox-material` for me to complain.

PR #66, opened 2021-01-12
Add 16color theme

shadmansaleh:
ported from lightline


shadmansaleh:
Does numbers come before or after laters in incremental sorting

hoob3rt:
> Does numbers come before or after laters in incremental sorting

numbers are generally before letters but why do you need it here anyways?

shadmansaleh:
>numbers are generally before letters but why do you need it here anyways?

the name of the theme is 16color . So I was wondering whether to put it on to or bottom in THEMES.md

Issue #67, opened 2021-01-12
Branch doesn't show anything on detached head state


shadmansaleh:
As The title says. If you're on detached head state branch component is empty . I think it should show the commit id.

hoob3rt:
this might be tricky since `git branch --show-current` has 0 exit code and no stdout in detached head state

shadmansaleh:
>this might be tricky since git branch --show-current has 0 exit code and no stdout in detached head state

Yep also I donn't think we can use empty output as a idicator . I think the async code returns empty output all the time . right?

shadmansaleh:
I think we can read .git/HEAD .

hoob3rt:
Yes this would be a better solution. I made the branch component use git commands for mvp. There is still room for improvements

shadmansaleh:
I'll create a pr when I implements it with .git/HEAD  . I donn't think it should be hard.
 go back to to find .git folder -> read .git/HEAD

Any idea on what exceptions we should handle?

shadmansaleh:
It's done

PR #68, opened 2021-01-13
Insted of calling git branch is read from HEAD

shadmansaleh:
addressing #67

let me know if we need to handle any other exceptions.


hoob3rt:
I think this works fine, I have not encounterd any errors while working on your branch.

Are you sure this will not produce errors similar to the branch function before `async` was implemented? I'm not sure how performance heavy are these io calls...

shadmansaleh:
>Are you sure this will not produce errors similar to the branch function before async was implemented? I'm not sure how performance heavy are these io calls...

I donn't really know what errors you are talking about . I wasn't using lualine before async was implemented . 

I donn't know how performance heavy those calls are .I didn't notice any performance drops . But the most used call is 
```local HEAD = head_file:read()``` in line 54 .Since the git directory is memoized .They are basicaly one off. That line can probably be made async easily .

Also since those calls are invoked by a timmer . Do they occur in main event loop or in a separate one without bloking ?

**Comment by hoob3rt** [comment]
> > Are you sure this will not produce errors similar to the branch function before async was implemented? I'm not sure how performance heavy are these io calls...
> 
> I donn't really know what errors you are talking about . I wasn't using lualine before async was implemented .
> 
> I donn't know how performance heavy those calls are .I didn't notice any performance drops . But the most used call is
> `local HEAD = head_file:read()` in line 54 .Since the git directory is memoized .They are basicaly one off. That line can probably be made async easily .
> 

> Also since those calls are invoked by a timmer . Do they occur in main event loop or in a separate one without bloking ?

This is actually a good question and one we need to research. If these calls don't block anything then we will be able to delete async module entirely.

For starters here is what the help says
```
schedule_wrap({cb})                                      *vim.schedule_wrap()*
                Defers callback `cb` until the Nvim API is safe to call.

```

**Comment by shadmansaleh** [comment]
>This is actually a good question and one we need to research. If these calls don't block anything then we will be able to delete async module entirely.

I think we can delete it now anyway . There's nothing else using it apart from branch component .
Any idea how to test whether they are blocking or not?

**Comment by shadmansaleh** [comment]
This branch genaration is way lighter than git cl calls . Also since actual file read is now async we can probably refreash it more often then once a second .

**Comment by shadmansaleh** [comment]
> ### int uv_timer_start(uv_timer_t* handle, uv_timer_cb cb, uint64_t timeout, uint64_t repeat)

>Start the timer. timeout and repeat are in milliseconds.
If timeout is zero, the callback fires on the next event loop iteration. If repeat is non-zero, the callback fires first after timeout milliseconds and then repeatedly after repeat milliseconds.

> ### void uv_timer_set_repeat(uv_timer_t* handle, uint64_t repeat)

>Set the repeat interval value in milliseconds. The timer will be scheduled to run on the given interval, regardless of the callback execution duration, and will follow normal timer semantics in the case of a time-slice overrun.

>For example, if a 50ms repeating timer first runs for 17ms, it will be scheduled to run again 33ms later. If other tasks consume more than the 33ms following the first timer callback, then the callback will run as soon as possible.

source : [libuv docs](http://docs.libuv.org/en/v1.x/timer.html#api)

I thinks timers aren't async . But since they aren't executed in first event loop thats why I get the effect of instantly loading neovim.

**Comment by shadmansaleh** [comment]
Now we donn't need to update every second.

**Comment by basilgood** [comment]
> Now we donn't need to update every second.
> 
> Any idea how can I wait for return value of an async function in lua like await ?
> 
> we pobably need somthing like [neovim async tutorial](https://github.com/ms-jpq/neovim-async-tutorial)

Hi! excuse me for getting into your discussion. 
Lualine from master branch with this setting:
lualine_b = { 'branch' } keep neovim on top five cpu consumers. 
On laptop will contribute to discharge the battery.
Without branch setting neovim it's always on the bottom of the consumers.
I think it's not necessary for lualine or just branch to update at every second. 
It's enough to update on events (like lightline does)
autocmd WinEnter,BufEnter,BufDelete,SessionLoadPost,FileChangedShellPost * call lightline#update()
I really like lualine plugin and I apologize once again for the inconvenience.

image

**Comment by shadmansaleh** [comment]
>Hi! excuse me for getting into your discussion.
Lualine from master branch with this setting:
lualine_b = { 'branch' } keep neovim on top five cpu consumers.
On laptop will contribute to discharge the battery.
Without branch setting neovim it's always on the bottom of the consumers.

That's bad . Thanks for letting us know . We'll look into that. I'll encourage you to create an issue with that . Though can you checkout this pr and see if this solves it issue.

>I really like lualine plugin and I apologize once again for the inconvenience.

No inconvenience at all.

**Comment by basilgood** [comment]
Yes I can confirm with this pr the problem is solved.
Only one minor thing that I observed.
Tmux with two panes in first neovim in the other terminal.
In terminal I write git rebase master and it stopts because it's a conflict.
The lualine branch changes correctly to git hash where it stopped.
This is done without focus gain in neovim.(super)
But when I write git rebase --abort in terminal the lualine branch changes to
am arbitrary git hash wich does not belong my git hases.
Correctly would be to show master but when neovim gain focus it changes to master(perfect).
I thing it's not a problem only just an observation.
Thank you.

**Comment by shadmansaleh** [comment]
>Yes I can confirm with this pr the problem is solved.

Thanks for confirming.

>In terminal I write git rebase master and it stopts because it's a conflict.
The lualine branch changes correctly to git hash where it stopped.
This is done without focus gain in neovim.(super)
But when I write git rebase --abort in terminal the lualine branch changes to
am arbitrary git hash wich does not belong my git hases.

I looked into it . I couldn't reproduce the 2nd one . 
#### Does the branch update instatly on your device or does it take a while to change after you run git rebase master ?? 
On my device it takes a while to change . But when I abort it doesn't change at all .
When tmux split is changed (comes to focus) BufEnter is not trigared. Since that's the event we track . And it updates without that event when tmux split switches that means file watch is working . I think it's the async read that gets delayed. If I turn it to a sync version it updates currectly and instently . What's funny is if I put a print() statment any where near the async function call to see the variables it works perfactly just like sync version . It reminds me of quantam mechanics double slit experiment lol. I'll look into it more later.

**Comment by shadmansaleh** [comment]
@basilgood can you check if this sync version works properly on your device.

**Comment by basilgood** [comment]
Yes with this sync version :tcd command changes the cwd for all the tabs and windows not only for the current.

**Comment by hoob3rt** [comment]
> Yes with this sync version :tcd command changes the cwd for all the tabs and windows not only for the current.

so the issue with #75 is not solved with this pr?

**Comment by basilgood** [comment]
Wait a sec.
I run nvim -u NONE and the behavior is the same it's change to all windows and tabs
I think it's a bug from neovim. 
Or my version of neovim 0.5.0-dev

**Comment by pianocomposer321** [comment]
@basilgood are you using Packer as you plugin manager? If so, you need to run `:PackerCompile` after nvim -u NONE to disable already installed plugins. When I do nvim with no init.vim and uninstall all packages it works for me.

**Comment by pianocomposer321** [comment]
@hoob3rt using @shadmansaleh's fork with the relevant branch does solve the issue for me 👍.

**Comment by basilgood** [comment]
Also I confirm this branch solved the issue for me :+1:

**Comment by shadmansaleh** [comment]
>@basilgood can you check if this sync version works properly on your device?

I was actualy talking about the rebase conflict and rebase abort issue you mentioned .

>Also I confirm this branch solved the issue for me 👍

Nice . Was it being caused by some other plugin?

**Comment by basilgood** [comment]
I don't know why it behaved like that. After a reboot everything went perfectly and yes I confirm that the problem with rebase abort has been resolved. :+1:

**Review by hoob3rt — COMMENTED** [review]
It's great that this pr solves a lot of underlying issues of the branch function. You can also remove the `async` module with this commit as it is no longer needed and I don't see any other feature it could be used for.

**Comment by pianocomposer321** [comment]
Cool, so @hoob3rt can this be merged now? I'm excited to be able to try adding tcd to my workflow 😉.

**Comment by hoob3rt** [comment]
> Cool, so @hoob3rt can this be merged now? I'm excited to be able to try adding tcd to my workflow wink.

not quite, I messed up my review and forgot to add comments, on it now :laughing:

File: lua/lualine/components/branch.lua [hunk 1]
```diff
@@ -1,33 +1,79 @@
...
+local function find_git_dir()
+  -- path seperator is not in the end add it
+  local file_path = vim.fn.expand('%:p:h')..sep
+  -- don't show branch for terminal buffers
+  if file_path:match('^term://.*$') then return nil end
+
+  local dir = file_path
+  -- do we need to iterate more than that?
+  local n = 30
+  while dir and n > 0 do
```
hoob3rt, line 23: Surely it would be better to have it search until $HOME universally on Linux, Mac and Windows. I'm not a fan of hardcoded variables and it will still be just as efficient while handling more edge cases (something like a huge java project may have a classic java directory hell)

lua/lualine/components/branch.lua
```diff
@@ -1,33 +1,79 @@
...
+    n = n - 1
+    -- See if we already know where it is
+    if known_git_dirs[dir] then
```
hoob3rt, line 26: This is another hardcoded variable I don't like but maybe there is no other way of improving performance. On the other hand galaxyline doesn't seem to be using any memorization at first glance, check it out [here](https://github.com/glepnir/galaxyline.nvim/blob/main/lua/galaxyline/provider_vcs.lua). Vimscript plugins like signify also don't use memorization as far as I know

hoob3rt:
Also won't all these issues be solved by using a better autocommand like @basilgood mentioned? This way we will have control on when the branch method runs. 
```vim
WinEnter,BufEnter,BufDelete,SessionLoadPost,FileChangedShellPos
```
I'm not sure if this is enough though and it could still produce the above mentioned tmux issue

shadmansaleh:
>You can also remove the async module with this commit as it is no longer needed and I don't see any other feature it could be used for.

I'll remove it . Even if we need it later we can recover it from git history

lua/lualine/components/branch.lua
shadmansaleh: I was actually hoping to expose it as options once #60 was merged . 
Though I was using it to survive infinite loops during devlopment as geting stuck in inf loops in vim isn't nice as keyboard intrupt signals are captured by vim . It will just hang forever until killed.
Also I donn't like exposingnlocal functions as global function to call it from a event . I'll change it to a require statment one #60 is merged . For now I can only return one function.

I donn't think $HOME is a good idea . User might be using a different partition then where his/her home directory is . Also I donn't think windows has  a $HOME (might be wrong here).

lua/lualine/components/branch.lua
shadmansaleh: This is related to upper one . Just ensures we iterate n times before giving up

shadmansaleh:
>Also won't all these issues be solved by using a better autocommand like @basilgood mentioned? This way we will have control on when the branch method runs instead of spamming it every second
```WinEnter,BufEnter,BufDelete,SessionLoadPost,FileChangedShellPos```

@basilgood  got it wrong . I've checked lightlines source code . Those ivents are used to change between active and inactive statusline just like we have as end of `lua/lualine.lua`.

As far as I know lightline doesn't provide git branches.

Also I think only watching for BufEnter for changing branches is enough because all the other events you mentioned will trigure a BufEnter later(I could be wrong). 

file watch on HEAD and BufEnter handles branch updating well . Only other instence I can think of where we can stop updating is when we reset file watch even when git dir has been evaluated to old value .

**Comment by shadmansaleh** [comment]
Does anyone have any idea why the issue with tmux may occure if async read is used? Changing just read to sync makes it work . Even using syncronus read from libuv works But async one doesn't . **Is it like neovim doesn't schedule async operations when not on focus ??** Also async io works without anything new when neovim comes to focus

**Comment by pianocomposer321** [comment]
@shadmansaleh If I'm understanding correctly, the issue is that if something about the git status changes outside of neovim, the statusline will not change until neovim gets focus again. To me this is not a dealbreaker, and maybe belongs in a separate issue? IDK, correct me if I'm wrong.

**Comment by shadmansaleh** [comment]
>If I'm understanding correctly, the issue is that if something about the git status changes outside of neovim

Nope it should change . As we are watching the HEAD file . If any other program changes head file kernal should notify us no matter whether neovim is in focus or not .

It works as expected when HEAD file is read syncronasly. It changes with external commands but doesn't work when the file is read asyncronusly

File: lua/lualine/components/branch.lua (hunk 1 above)
Comment (hoob3rt) [reply lua/lualine/components/branch.lua:23]: No way a user should have an option to change how this behaves. This will only produce issues down the line.

Please provide an example on how you plan on changing this to a require :)

 this still needs improvement, one reason why you could have been getting infinite loops was using dfs instead of bfs.
 you can check [rooter implementation for reference](https://github.com/airblade/vim-rooter/blob/master/plugin/rooter.vim) for inspiration

File: lua/lualine/components/branch.lua (hunk 2 above)
Comment (hoob3rt) [reply lua/lualine/components/branch.lua:26]: Won't this create another issue of possibly watching a lot of files.

Also a quick note, this is completely broken for submodules due to memorizing. I personally would prefer to ditch this part of pr if we can test this does not affect performance, I hope it doesn't. If it does we will have to look into an async implementation (other than the old async module)

File: lua/lualine/components/branch.lua (hunk 1 above)
Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:23]:
```
vim.cmd[[autocmd BufEnter * :lua require'lualine.components.branch'.update_branch()]]
return {
  init = function(args) return branch(args) end,
  update_branch = watch_head}
```

I actually haven't gotten any inf loops ever since implementation was stable (about a week) . I was getting them when I was writting the function . I just left them just incase. It most likely can be safely removed.

Whats dfs and bfs?

File: lua/lualine/components/branch.lua (hunk 2 above)
Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:26]: >Won't this create another issue of possibly watching a lot of files?

I'm only watching one file at a time . The watcher is reset before a new file is set to watch.

>Also a quick note, this is completely broken for submodules due to memorizing. 

I haven't tested with submodules I'll update after I've tested.

File: lua/lualine/components/branch.lua (hunk 1 above)
Comment (hoob3rt) [reply lua/lualine/components/branch.lua:23]: depth first search and breadth first search...

I hope we can remove this hardcoded value somehow without getting cpu heavy/infinite looping

File: lua/lualine/components/branch.lua (hunk 2 above)
Comment (hoob3rt) [reply lua/lualine/components/branch.lua:26]: There is no noticeable performance difference between between using or not using branch component on startup so there should be no issue while changing buffers either(without using memorization)

Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:26]: Are you sure it's due to memorization . Cause it still doesn't work after commenting out everything with memorization.

File: lua/lualine/components/branch.lua (hunk 1 above)
Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:23]: We are going up the directory tree not tranversing down. Does bfs have any value here at all?

Comment (hoob3rt) [reply lua/lualine/components/branch.lua:23]: Okay wait, this whole `find_git_dir` function is traversing up in line `38` to find possible match and traversing down in line `45` if no match is found? Am I interpreting this correctly now?

Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:23]: maybe your line numbers are off .
basically 
- check if dir/.git exists
- go up a dir
- repeat

File: lua/lualine/components/branch.lua (hunk 2 above)
Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:26]: Submodule bug is not for memorization . It's happening because submodule/.git has a relative path based on submodules dirextory . But that path is interpreted as users curent dir . I'll see how can I fix it.

File: lua/lualine/components/branch.lua (hunk 1 above)
Comment (hoob3rt) [reply lua/lualine/components/branch.lua:23]: yeah yeah ok. 

If that's the case then everything works as expected, but if that's the case then I think the io calls in `exists` are useless since in path
```
a/b/c/d/e/file
```
The git repo will be in one of a,b,c,d,e and I think we know for sure that these are directories, however we still need to figure out a better ending point for this loop, better than `n=30`

File: lua/lualine/components/branch.lua (hunk 2 above)
Comment (shadmansaleh) [reply lua/lualine/components/branch.lua:26]: Submodule bug fixed though I donn't like
```        if git_dir:sub(1,2) == '..' then
```
condition to detect relative path in submodules . I'd like something more genaral . Open to suggestions.

hoob3rt: nevermind, there was an issue with what I was trying to do. Either way, I don't think this small optimization will give us much value and we should remove the memorization part. 
vim-rooter does no memorization at all and runs on WinEnter, BufEnter and I had no issues for 2 years of using it.
Removing it will make the code cleaner while keeping performance marginally worse, I think it's worth it

lua/lualine/components/branch.lua
shadmansaleh: >I think the io calls in exists are useless since in path

I donn't know any other way to verify if a file exists or not If you know please suggest.

> however we still need to figure out a better ending point for this loop, better than n=30

I agree with that . but that was mostly to block inf loops . The thing is there might be some special char in filepath that may break the algorithm and cause inf loop . Since I donn't know all sort of filepath that are valid . I've left it there as a contingency

lua/lualine/components/branch.lua
shadmansaleh: Alright if you are so keen to remove memorization I'll remove it . 

Do you know how full paths start in windows and mac . Like in linux it always starts with '/'

lua/lualine/components/branch.lua
hoob3rt: I think it would be better to use `vim.fn.empty` and `vim.fn.isdirectory` since they are already implemented and are surely faster than io calls

lua/lualine/components/branch.lua
hoob3rt: on mac they start with / as well, don't remember how they are on windows, why do you need it?

lua/lualine/components/branch.lua
hoob3rt: this whole function is actually reduced to `vim.fn.isdirectory(path)`

We can keep io calls in `get_git_head` since we actually need to read this file. 

This change will reduce io calls to a total of ONE :)

lua/lualine/components/branch.lua
```diff
@@ -1,33 +1,83 @@
...
+        known_git_dirs[file_path] = known_git_dirs[dir] end
+      return known_git_dirs[dir]
     end
-  end,
-  on_stderr = function (_, data)
-    if data then
-      if data:find("fatal: not a git repository") then
-        git_branch = ''
+    local git_dir = dir..'.git'
+    local is_available, is_directory = exists(git_dir)
```
hoob3rt, line 33: this whole loop can be replaced with 
```lua
  local dir = file_path
  while dir do
    local git_dir = dir..'.git'
    local is_directory = vim.fn.isdirectory(git_dir)
    if is_directory==1 then
      return git_dir
    end
    dir = dir:match("(.*"..sep..").+$") -- "(.*/).+$"
  end
```

hoob3rt: please check if this does not produce any new issues or reproduce above mentioned ones. For the minimal testing I could do it works just fine

lua/lualine/components/branch.lua
hoob3rt: the change I suggested below works well for linux, it traverses to root. I will test it if it works the same way on mac an windows later today

lua/lualine/components/branch.lua
hoob3rt: actually there might be an even better way, check out `:h finddir` and `:h file-searching`

hoob3rt: try
```vim
:echo finddir('.git', '.;')
```
I think this is a drop in replacement for this whole function

shadmansaleh: Just looking for dir wonn't work as when 
`git init --separate-dir` is used or it's a submodule . In those cases .git is a file. Plus I'm sure this lua implenentation will way more efficient then vimL implementation.

hoob3rt: And I am sure that `finddir` is not written in vimscript but in C :)

shadmansaleh: >And I am sure that finddir is not written in vimscript but in C :)

Alright then it's a good point . But we still will have to handle those submodules . It still doesn't change the problem we are having with full path . as we will still have to handle it .

hoob3rt: full path? I think only submodules should be handled. I honestly have no idea how to do it and you seemed to have figured it out already so could you handle it? Maybe some smart if will do the trick.
Right now I am using 
```lua
  local git_dir = vim.fn.finddir('.git', '.;')
```
you can find the `.git` file easily with `findfile` but I have no idea on how to parse it

shadmansaleh: I'm not sure but I think finddir and findfile runs from curent dir insted of filedir . can you tell me what '.;' means ? 
It reintruduces the issue where branch doesn't change on non git dir.

hoob3rt: Oh snap it does...

well `.;` can be translated to 
`.` vim starting directory
`;` until you find a desired file or reach the top of tree

I pulled this up in 3 minutes, there should be an option to replace `.` with current dir. If not, we can just use the expanded filepath because it also accepts relative & full paths

hoob3rt: yeah, rather easy fix, I think this works
```lua
local function watch_head()
  file_changed:stop()
  local dir = vim.fn.expand('%:p:h')..sep
  local git_dir = vim.fn.finddir('.git', dir .. ';')
  if #git_dir > 0 then
    local head_file = git_dir..sep..'HEAD'
    get_git_head(head_file)
    file_changed:start(head_file, {}, vim.schedule_wrap(function()
      -- reset file-watch
      watch_head()
    end))
  else
    -- set to nil when git dir was not found
    git_branch = nil
  end
end
```

shadmansaleh:
I think this is equvelent of lua implementation .Can you test it and see if ypu can find bugs..

lua/lualine/components/branch.lua
```diff
@@ -1,34 +1,60 @@
...
-      end
+-- os specific path separator
+local sep = package.config:sub(1,1)
+
+-- returns full path to git directory for current directory
+local function find_git_dir()
+  local file_dir = vim.fn.expand('%:p:h') .. ';'
+  local git_dir = vim.fn.finddir('.git', file_dir)
+  local git_file = vim.fn.findfile('.git', file_dir)
+  if #git_file > 0 then git_file = vim.fn.fnamemodify(git_file, ':p') end
```
hoob3rt, line 11: care to explain what this does?

hoob3rt:
> I think this is equvelent of lua implementation .Can you test it and see if you can find bugs..

I think it's way better. io calls are expensive and we reduced them from n+2 to 2

lua/lualine/components/branch.lua
hoob3rt: I would also rename `file_dir` to `current_working_dir`

shadmansaleh:
>I think it's way better. io calls are expensive and we reduced them from n+2 to 2

Donn't you think findfiles and finddir themselves are using io calls.

hoob3rt:
> > I think it's way better. io calls are expensive and we reduced them from n+2 to 2
> 
> Donn't you think findfiles and finddir themselves are using io calls.

No I don't. Similarly to how ls doesn't.

lua/lualine/components/branch.lua
shadmansaleh: See the coments. Better to leave explanation for anyone who reads the code.

shadmansaleh:
>No I don't. Similarly to how ls doesn't.

How are you sure . If a program needs to know it has to communicate with hardware through kernal . Even if kernal is makeing the calls . The hardware can only tell what file exists and what not . If you are communicating with hardware it's an io call in my opinion . Though I can be very wrong as I'm not CS student.

shadmansaleh:
I've actually set sepcial highlights for trailing whitespaces yet I missed it . I should add an autocmd on write to remove whitespaces 🤔

hoob3rt:
> > No I don't. Similarly to how ls doesn't.
> 
> How are you sure . If a program needs to know it has to communicate with hardware through kernal . Even if kernal is makeing the calls . The hardware can only tell what file exists and what not . If you are communicating with hardware it's an io call in my opinion . Though I can be very wrong as I'm not CS student.

at least in linux, you can use `stat()` function in C to get file properties which are managed by the kernel. As far as I know there is no file opening involved

hoob3rt:
Linux kernel aside, I think this is ready to merge. Thank you so much for helping me with this pr. I love what you are doing :)

shadmansaleh:
>at least in linux, you can use stat() function in C to get file properties which are managed by the kernel. As far as I know there is no file opening involved

I know about stat . But kernal doesn't store file stats strage devices does. kernal probably quires filetable in disk . Will it not be considered an IO ?

hoob3rt:
> > at least in linux, you can use stat() function in C to get file properties which are managed by the kernel. As far as I know there is no file opening involved
> 
> I know about stat . But kernal doesn't store file stats strage devices does. kernal probably quires filetable in disk . Will it not be considered an IO ?

Welp, that's really specific, I don't know. Either way it surely is faster than `io:open`

shadmansaleh:
>Welp, that's really specific, I don't know. Either way it surely is faster than io:open

It surely is.

pianocomposer321:
> Linux kernel aside, I think this is ready to merge. Thank you so much for helping me with this pr. I love what you are doing :)

Indeed, @shadmansaleh thank you for your hard work. It is really appreciated!

PR #70, opened 2021-01-13
Update onedark+onelight theme

shadmansaleh:
ported from lightline


hoob3rt:
is't `vimone` just an iteration of `onedark`. I agree that this looks better with your implementation. So maybe we should replace the current onedark with this. Call them `onedark` and `onelight`? What do you think?

shadmansaleh:
> is't vimone just an iteration of onedark. I agree that this looks better with your implementation. So maybe we should replace the current onedark with this. Call them onedark and onelight? What do you think?

Yes it's a iteration of ondark I've just looked up. Since we donn't have a onelight we should replace the current ondark theme.

PR #71, opened 2021-01-13
Add nightfly theme

shadmansaleh:
~~I liked what #26 did to nightfly colorscheme I was using . So i dumped the color table tweeked some colors and got this 😁~~

I've changed it to official vertion of nightfly theme


hoob3rt:
looks perfect to me

PR #72, opened 2021-01-13
Adding gruvbox_metarial theme

shadmansaleh:
It's actualy quite simple to dump a theme
just add this to end of vimscript
```lua
lua<<END
f = io.open('theme.lua', 'w')
f:write(vim.inspect(vim.g["lightline#colorscheme#gruvbox_material#palette"]))
f:close()
END
```
replace lightline#colorscheme#gruvbox_material#palette
with the veriable name your theme is using .
Source the file with source command .
And you will have a lua table with the colorcodes theme.lua file.
some global search-replace and macros can convert that table to a lualine theme in notime.


coxackie:
Lol. So, you did it. And there *is* and easy way. Kudos, you are great!

PR #73, opened 2021-01-15
run vimscript inside vim.api.nvim_exec

shadmansaleh:
Insted of having vim.cmd for each vimL line we can wrap the whole thing in vim.api.nvim_exec


hoob3rt:
Yeah that's great!

PR #74, opened 2021-01-18
More themes

shadmansaleh:
Added themes:
- Tomorrow.lua
- ayu_dark.lua
- ayu_light.lua
- ayu_mirage.lua
- iceberg_dark.lua
- iceberg_light.lua
- jellybeans.lua
- oceanicnext.lua
- onehalf.lua
- seoul256.lua

I'm done with themes . I think I've got most of what we need.


THEMES.md
```diff
@@ -2,6 +2,24 @@
 
 All available themes are only best effort ports by myself/ other users. If you find a theme to be weird/ wrong please open an issue/ pr.
 
+### ayu dark
```
hoob3rt, line 5: all names should be changed to work directly in lualine. 
```vim
:lua require('lualine').theme = 'ayu dark'
```
Obviously breaks, should be `ayu_dark` in THEMES.md

shadmansaleh: fixed

hoob3rt:
`onehalf` looks like just another `onedark` implementation. We don't need it if #70 is merged

shadmansaleh:
> onehalf looks like just another onedark implementation. We don't need it if #70 is merged

Yes it seems both are quite similar. Though One Half doesn't state it's a derivation of onedark their statusline implementation seems same . It can be removed.

hoob3rt:
some improvements needed:
ayu_light and ayu_mirage blend with background, this is bad imo

shadmansaleh:
How about now?

hoob3rt:
I fixed it myself, looks better now imo

PR #79, opened 2021-01-23
Fix #33

### LINKED ISSUE #33: more universal way of setting the statusline
https://github.com/hoob3rt/lualine.nvim/blob/4cf4ac814d5c7acf6470f37047bf771d7cb67e55/lua/lualine.lua#L105-L115

There should be a better way of making lualine use different function calls for each window. This hack is ugly and will probably prove to be hard to maintain.

shadmansaleh:
A posible solution to #33

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -65,14 +65,7 @@ local function set_lualine_theme()
   theme_set = M.theme
 end
 
-local function statusline(is_focused)
-  local sections = M.sections
-  if not is_focused then
-    sections = M.inactive_sections
-  end
-  if M.theme ~= theme_set then
-    set_lualine_theme()
-  end
+local function statusline(sections, is_focused)
   local status = {}
   if sections.lualine_a then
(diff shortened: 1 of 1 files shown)


hoob3rt:
That looks a lot cleaner for sure. I will use it on my workbench tomorrow and merge if I don't stumble across any bugs. Thanks :)

hoob3rt:
lgtm

megalithic:
@hoob3rt i THINK this might have reintroduced the issue with inactive buffers replicating the info in those components, e.g. filename being the same between active and inactive, etc.

shadmansaleh:
>@hoob3rt i THINK this might have reintroduced the issue with inactive buffers replicating the info in those components, e.g. filename being the same between active and inactive, etc.

Yes you are right this intoduced it . I'm trying to fix the bug without reverting it . Though I cann't understand why it's happening . It seems accessing vim.bo gets the active buffer not the buffer the function was called for .

PR #81, opened 2021-01-30
Add codedark theme

Shatur95:
normal
insert
visual
replace

Ported from [vim-code-dark](https://github.com/tomasiser/vim-code-dark) using [lightline2lualine_theme_converter.lua](https://gist.github.com/shadmansaleh/000871c9a608a012721c6acc6d7a19b9).

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -8,6 +8,12 @@ All available themes are only best effort ports by myself/ other users. If you f
 ![visual](https://user-images.githubusercontent.com/13149513/104338631-e852ff80-5520-11eb-9341-0ecd87aed656.jpg)
 ![replace](https://user-images.githubusercontent.com/13149513/104338654-eee17700-5520-11eb-9ff3-d2308d1243e5.jpg)
 
+### codedark
+![normal](https://user-images.githubusercontent.com/22453358/106365012-66920d00-633b-11eb-97a2-a10a0c48b1dc.png)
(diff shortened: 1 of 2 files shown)


hoob3rt:
Onedark already exists. Is the current implementation not inline with how the theme is supposed to work?

Shatur95:
> Onedark already exists.

This is Code dark (VSCode default theme), not related to One dark (Atom default theme).

hoob3rt:
Oh yea yea, my bad. I misread it

THEMES.md
```diff
@@ -2,6 +2,12 @@
 
 All available themes are only best effort ports by myself/ other users. If you find a theme to be weird/ wrong please open an issue/ pr.
 
+### onedark
```
hoob3rt, line 5: should be changed to codedark then

Shatur95: Thanks, renamed.

hoob3rt:
Screenshot 2021-02-04 at 21 13 43

also, I've tested it with [this version of codedark for vim](https://github.com/tomasiser/vim-code-dark) and the blue seems a bit off. Am I missing something here?

Shatur95:
> also, I've tested it with this version of codedark for vim and the blue seems a bit off. Am I missing something here?

Hm... Looks ok to me. But do you want to suggest another color?

hoob3rt:
Well I guess fair enough. It looks all right for me and if you, a user of codedark, approve it then it's okay

Shatur95:
Thank you!

PR #83, opened 2021-01-31
Don't show signify status symbols if the count < 1

beauwilliams:
It is not very informative to have indicators in our statusline that tell us that there are 0 deletions or additions etc. This proposed change will only display relevant indicators if they have a count > 0. This leads to a cleaner statusline that is less cluttered and removes the dreaded '+0 -0 ~1' scenarios. Instead, only '~1' would show in that instance 😀


hoob3rt:
Yeah, that's better imo. Thanks! :)

PR #87, opened 2021-02-06
Drop signify dependency

shadmansaleh:
### Signify component is nolonger dependent on signify plugin
- Sign counts have been implemented inside lualine :)

We can probably change the name of the component too now.


shadmansaleh:
This one doesn't have any conflits.
So @hoob3rt ready for review

hoob3rt:
I would say we tackle this after #61, #94 and #95 are done as they are higher priority imo

shadmansaleh:
Ok .

shadmansaleh:
@hoob3rt I think we should rename signify component in this pr to something different So users donn't think it requires signify to be installed . What do you think ?

hoob3rt:
> @hoob3rt I think we should rename signify component in this pr to something different So users donn't think it requires signify to be installed . What do you think ?

Yes that should be done. I think `diff` makes sense but lmk if you have better ideas

shadmansaleh:
I think I'll rename it to 'git_diff' and add a stub for signify that gives user warning to change config and load 'git_diff' component after you submit reviews. We can remove the stub after a few days so users donn't think all of a sudden signify stoped wroking 😄

hoob3rt:
> I think I'll rename it to 'git_diff'

if we name this `git_diff` then `branch` should be called `git_branch` for consistency. It doesn't make sense to have one git related component have a git prefix and one with no prefix. I honestly don't think renaming `branch` is a good idea. It is pretty self explanatory so there should probably be a better name for this component. 

> add a stub for signify that gives user warning to change config and load 'git_diff' component after you submit reviews. We can remove the stub after a few days so users donn't think all of a sudden signify stoped wroking 😄

godd idea

lua/lualine/components/signify.lua
```diff
@@ -1,9 +1,90 @@
 -- Copyright (c) 2020-2021 shadmansaleh
 -- MIT license, see LICENSE for more details.
 
+local async = require'lualine.async'
```
hoob3rt, line 4: it would be nice to implement this without the async module. No need to have 75sloc just to call a single command here.

shadmansaleh: I think we should keep the async module . This will make it easy to write other components that rely on extarnal program on future. Also I donn't think this module will require much maintanace as it already does basicaly all it needs to do . 75loc doesn't seem too much to me.

shadmansaleh:
Ok then I'll use 'diff'

lua/lualine/components/signify.lua
hoob3rt: Yea my concern is that I don't see another use case for this module for now. But I agree that async module is really simple so we can keep it. I will review and hopefully merge the rest of this pr later today.

hoob3rt:
btw what do you mean by stub. I was thinking to add a warning, something like
```lua
if component == signify then print('some warning') 
```

shadmansaleh:
I was actualy thinking putting something like this in signify.lua
```lua
print('Some warning')
return require'lualine.components.diff'
```
than we can just delete the file later.

hoob3rt:
Yea that's a good solution as well, let's go with that

hoob3rt:
Unfortunately this does not work properly at all
Screenshot 2021-02-22 at 17 56 48
Screenshot 2021-02-22 at 17 56 57
Screenshot 2021-02-22 at 17 57 25

The `process_diff` function is not implemented well

shadmansaleh:
I've fixed the bug . Now you can see it .

README.md
```diff
@@ -141,8 +141,7 @@ lualine.inactive_sections = {
   * location (location in file in line:column format)
   * mode (vim mode)
   * progress (%progress in file)
-* plugin
-  * signify (signify status)
+  * diff (signify status)
```
hoob3rt, line 144: should be (git diff status)

doc/lualine.txt
```diff
@@ -166,8 +166,7 @@ Available components~
         *   location (location in file in line:column format)
         *   mode (vim mode)
         *   progress (%progress in file)
-    plugin~
-        *   signify (signify status)
+        *   diff (signify status)
```
hoob3rt, line 169: same here

lua/lualine/components/diff.lua
```diff
@@ -0,0 +1,183 @@
...
+  vim.schedule_wrap(function()
+    if get_git_diff then
+      updated = false
+      get_git_diff:start()
+    end
+  end)()
+end
+
+vim.api.nvim_exec([[
+  autocmd lualine BufEnter     * lua require'lualine.components.signify'.update_git_diff_getter()
```
hoob3rt, line 86: why require `signify` when we plan to remove it in a week?

lua/lualine/components/diff.lua
```diff
@@ -0,0 +1,183 @@
...
+local function update_git_diff()
+  vim.schedule_wrap(function()
+    if get_git_diff then
+      updated = false
+      get_git_diff:start()
+    end
+  end)()
+end
+
+vim.api.nvim_exec([[
```
hoob3rt, line 85: shouldn't this be placed somewhere else? probably in the `signify` function. I don't like global calls.

lua/lualine/components/diff.lua
```diff
@@ -0,0 +1,183 @@
...
+  if not options.color_removed then
+    options.color_removed = utils.extract_highlight_colors('DiffDelete', 'guifg') or default_color_removed
+  end
+
+  local highlights = {}
+
+  -- create highlights and save highlight_name in highlights table
+  local function create_highlights()
+    highlights = {
+      highlight.create_component_highlight_group({fg = options.color_added}, 'signify_added', options),
```
hoob3rt, line 113: `signify_added` is no longer relevant? Maybe `lualine_diff_added` or something along those lines

lua/lualine/components/diff.lua
```diff
@@ -0,0 +1,183 @@
...
+  autocmd lualine BufEnter     * lua require'lualine.components.signify'.update_git_diff_getter()
+  autocmd lualine BufEnter     * lua require'lualine.components.signify'.update_git_diff()
+  autocmd lualine BufWritePost * lua require'lualine.components.signify'.update_git_diff()
+  ]], false)
+
+local default_color_added    = "#f0e130"
+local default_color_removed  = "#90ee90"
+local default_color_modified = "#ff0038"
+
+local function signify(options)
```
hoob3rt, line 95: signify again...

hoob3rt (commented):
Looks good! Some small tweaks needed and we can merge this

shadmansaleh:
I can see I did a very sloppy rename 😅

shadmansaleh:
@hoob3rt  All the issues resolved

hoob3rt:
looks good now

toastal:
For the record from signify:

> Supports git, mercurial, darcs, bazaar, subversion, cvs, rcs, fossil, accurev, perforce, tfs, yadm.

Support for mercurial, darcs, bazaar, subversion, cvs, rcs, fossil, accurev, perforce, tfs, yadm is now lacking as _only_ git is supported.

PR #88, opened 2021-02-07
Load lualine asyncronasly to improve startup

shadmansaleh:
Well I've moved component and other loadings to a async. So Startup should be faster .
` nvim --startuptime` 
[tweekmonster/startuptime.vim](https://github.com/tweekmonster/startuptime.vim)
[norcalli/profiler.nvim](https://github.com/norcalli/profiler.nvim)
Does show startuptime improvements . 

Though I donn't have much experience in async. You can check it out if profilers are fooled or startup time actually improves . Though since lualine is already quite fast so I guess improvement isn't that huge . But in future newer components will add to that time . Also code complexcity isn't realy increased so I think it's a good idea to load lualine asyncronusly .

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -1,5 +1,5 @@
-local utils = require('lualine.utils')
-local highlight = require('lualine.highlight')
+local utils
+local highlight
 
 local M = { }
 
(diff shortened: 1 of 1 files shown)


hoob3rt:
I don't think this is necessary. First of all it does not load lualine for some reason when you open vim with no file provided.
All things you made async here are required for lualine to work properly so I don't really see any benefit it being async.

shadmansaleh:
>I don't think this is necessary.

It's to speed up startup time . Currently it still takes a moment to load also varies based on user config (What components they are loading).
What about just loading the components and extentions asyncronasly while loading visual stuff syncronasly ?

>First of all it does not load lualine for some reason when you open vim with no file provided.

I cann't reproduce it . On my device it loads lualine fine .Can you tell me how to reproduce it ?

>All things you made async here are required for lualine to work properly.

They are required but it doesn't really matter whether they are loaded syncronasly or asyncronasly.

> I don't really see any benefit it being async.

Ovious benefit is lualine wonn't block startup.

PR #94, opened 2021-02-15
Powerline like section seperators v2

shadmansaleh:
It's basicaly merge of #55 and #60 + some fixes and improvements.

The #55 couldn't handle special highlighting in #60 there were some conflits too , I've fixed it .

Changes:
- added component separators{left, right}
- added section separators{left, right}
- We were creating too many unnessary highlight groups (for providing defaults) . I've changed that . removed `apply_defaults` function from highlights.lua . Defaults are handled by `format_highlight` function directly
- Added `get_transition_highlight` function that provides highlight groups for section_separator.
- Other miscellanious changes to make everything work

**for review**:
Changes are confined in:
- lua/lualine.lua 
- lua/lualine/highlight.lua
- lua/lualine/utils/component.lua

**Notes:** If custom haghlights are managed by component itself and not using color option and changing components background color.  it should set custom_highlight option to true. So component separators aren't displayed .

@hoob3rt  you sould merge it after #60

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -38,6 +38,17 @@ M.inactive_sections = {
 M.extensions = {
 }
 
+local function check_single_separator()
+  local compoennt_separator = M.options.component_separators
+  local section_separator = M.options.section_separators
+  if type(M.options.component_separators) == 'string' then
+    M.options.component_separators = {compoennt_separator, compoennt_separator}
+  end
+  if type(M.options.section_separators) == 'string' then
+    M.options.section_separators = {section_separator, section_separator}
+  end
+end
+
 local function load_special_components(component)
   return function()
     -- precedence lualine_component > vim_var > lua_var > vim_function
@@ -176,7 +187,7 @@ local function statusline(sections, is_focused)
                    'lualine_'..section_name)
 			local section_data = utils_component.draw_section(sections['lualine_'..section_name], section_highlight)
 			if #section_data > 0 then
-				table.insert(status_builder, {section_name, section_data})
+				table.insert(status_builder, {name = section_name, data = section_data,})
 			end
     end
   end
@@ -186,7 +197,7 @@ local function statusline(sections, is_focused)
   local half_passed = false
   for i=1,#status_builder do
     -- midsection divider
-    if not half_passed and status_builder[i][1] > 'c' then
+    if not half_passed and status_builder[i].name > 'c' then
       table.insert(status, highlight.format_highlight(is_focused, 'lualine_c') .. '%=')
       half_passed = true
     end
@@ -194,29 +205,29 @@ local function statusline(sections, is_focused)
     if is_focused then
       -- component separator needs to have fg = current_section.bg
       -- and bg = adjacent_section.bg
-      local previous_section = status_builder[i-1] and status_builder[i-1] or {}
+      local previous_section = status_builder[i-1] or {}
       local current_section = status_builder[i]
-      local next_section = status_builder[i+1] and status_builder[i+1] or {}
+      local next_section = status_builder[i+1] or {}
       -- For 2nd half we need to show separator before section
-      if current_section[1] > 'x' then
-        local transitional_highlight = highlight.get_transitional_highlights(previous_section[2], current_section[2], true)
+      if current_section.name > 'x' then
+        local transitional_highlight = highlight.get_transitional_highlights(previous_section.data, current_section.data, true)
         if transitional_highlight and M.options.section_separators and M.options.section_separators[2] then
           table.insert(status, transitional_highlight .. M.options.section_separators[2])
         end
       end
 
       -- **( insert the actual section in the middle )** --
-      table.insert(status, status_builder[i][2])
+      table.insert(status, status_builder[i].data)
 
       -- For 1st half we need to show separator after section
-      if current_section[1] < 'c' then
-        local transitional_highlight = highlight.get_transitional_highlights(current_section[2], next_section[2])
+      if current_section.name < 'c' then
+        local transitional_highlight = highlight.get_transitional_highlights(current_section.data, next_section.data)
         if transitional_highlight and M.options.section_separators and M.options.section_separators[1] then
           table.insert(status, transitional_highlight .. M.options.section_separators[1])
         end
       end
     else -- when not in focus
-      table.insert(status, status_builder[i][2])
+      table.insert(status, status_builder[i].data)
     end
   end
   -- incase none of x,y,z was configured lets not fill whole statusline with a,b,c section
@@ -247,6 +258,7 @@ local function exec_autocommands()
 end
 
 function M.status()
+  check_single_separator()
   lualine_set_theme()
   exec_autocommands()
   load_components()
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -235,37 +235,38 @@ local function statusline(sections, is_focused)
   end
   -- incase none of x,y,z was configured lets not fill whole statusline with a,b,c section
   if not half_passed then
-    table.insert(status, highlight.format_highlight(is_focused,'lualine_c').."%=") end
-    return table.concat(status)
+    table.insert(status, highlight.format_highlight(is_focused,'lualine_c').."%=")
   end
+  return table.concat(status)
+end
 
-  local function status_dispatch()
-    if vim.g.statusline_winid == vim.fn.win_getid() then
-      return statusline(M.sections, true)
-    else
-      return statusline(M.inactive_sections, false)
-    end
+local function status_dispatch()
+  if vim.g.statusline_winid == vim.fn.win_getid() then
+    return statusline(M.sections, true)
+  else
+    return statusline(M.inactive_sections, false)
   end
+end
 
-  local function exec_autocommands()
-    _G.lualine_set_theme = lualine_set_theme
-    _G.lualine_statusline = status_dispatch
-    vim.api.nvim_exec([[
-    augroup lualine
-    autocmd!
-    autocmd ColorScheme * call v:lua.lualine_set_theme()
-    autocmd WinLeave,BufLeave * lua vim.wo.statusline=lualine_statusline()
-    autocmd WinEnter,BufEnter * setlocal statusline=%!v:lua.lualine_statusline()
-    augroup END
-    ]], false)
-  end
+local function exec_autocommands()
+  _G.lualine_set_theme = lualine_set_theme
+  _G.lualine_statusline = status_dispatch
+  vim.api.nvim_exec([[
+  augroup lualine
+  autocmd!
+  autocmd ColorScheme * call v:lua.lualine_set_theme()
+  autocmd WinLeave,BufLeave * lua vim.wo.statusline=lualine_statusline()
+  autocmd WinEnter,BufEnter * setlocal statusline=%!v:lua.lualine_statusline()
+  augroup END
+  ]], false)
+end
 
-  function M.status()
-    check_single_separator()
-    lualine_set_theme()
-    exec_autocommands()
-    load_components()
-    load_extensions()
-  end
+function M.status()
+  check_single_separator()
+  lualine_set_theme()
+  exec_autocommands()
+  load_components()
+  load_extensions()
+end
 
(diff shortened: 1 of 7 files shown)


hoob3rt:
This also requires a merge from master. There is a lot of code here which doesn't need to be changed at all to have this feature. tag me on all pr's which will be ready for review please

shadmansaleh:
rebased on master @hoob3rt ready for review.

lua/lualine.lua
```diff
@@ -150,32 +152,60 @@ local function statusline(sections, is_focused)
...
-    table.insert(status, utils_component.draw_section(sections.lualine_y,  hl))
+  -- status_builder stores statusline without section_separators
+  local status_builder = {}
+  -- The sequence sections should maintain
+  local section_sequence = {'a', 'b', 'c', 'x', 'y', 'z'}
+
+  for _, sec in ipairs(section_sequence) do
+    if sections['lualine_'..sec] then
+      -- insert highlight+components of this section to status_builder
+      local highlight = highlight.format_highlight(is_focused,
```
hoob3rt, line 163: redefined local highlight, yikes

lua/lualine.lua
```diff
@@ -150,32 +152,60 @@ local function statusline(sections, is_focused)
...
+    -- midsection divider
+    if not half_passed and status_builder[i][1] > 'c' then
+      table.insert(status, highlight.format_highlight(is_focused, 'lualine_c') .. '%=')
+      half_passed = true
+    end
+    -- provide section_separators when statusline is in focus
+    if is_focused then
+      -- component separator needs to have fg = current_section.bg
+      -- and bg = adjacent_section.bg
+      local prev_section = status_builder[i-1] and status_builder[i-1] or {[2] = nil}
```
hoob3rt, line 185: I think it would be cleaner to use a map rather than array here. `[2]=nil]` is hard to read.

hoob3rt (commented):
I would also change variable names to be more consistent with the whole codebase. We have nice variables names everywhere and here some are nice as well (`section_sequence`) and some are not inline with the rest (`cur_section`, `sec`)

lua/lualine.lua
shadmansaleh: I was kind of running out of names lol. I'll rename it any suggestion?

lua/lualine.lua
shadmansaleh: does {nil, nil} seem better?

shadmansaleh:
@hoob3rt I've updated the  veriable names. If you cann't see any other issue then we can probably merge it.

lua/lualine/utils/utils.lua
```diff
@@ -7,17 +7,35 @@ local M = {}
...
+  if gui_colors.background then
+    color.guibg = string.format('#%06x', gui_colors.background)
+    gui_colors.background = nil
+  end
+  if gui_colors.foreground then
+    color.guifg = string.format('#%06x', gui_colors.foreground)
+    gui_colors.foreground = nil
+  end
+  cterm_colors.background = nil
+  cterm_colors.foreground = nil
```
hoob3rt, line 35: why is `cterm_colors` reset to nil here?

lua/lualine.lua
```diff
@@ -12,9 +12,11 @@ local theme_set = {}
 M.options = {
   icons_enabled = true,
   theme = 'gruvbox',
-  separator = '|',
+	component_separators = {'', ''},
+	section_separators = {'', ''},
```
hoob3rt, line 16: I think it would be better to have this set up to work this way:
```lua
separator = { component = {'something', 'something'}, section = {'something', 'something'}
```
while also keeping the functionality of 
```lua
separator = '|'
```
I think the configuration will get easier if you want a static separator like it is right now and also make code in `component.lua` a lot easier.

lua/lualine.lua
```diff
@@ -148,34 +150,66 @@ end
...
+    -- midsection divider
+    if not half_passed and status_builder[i][1] > 'c' then
+      table.insert(status, highlight.format_highlight(is_focused, 'lualine_c') .. '%=')
+      half_passed = true
+    end
+    -- provide section_separators when statusline is in focus
+    if is_focused then
+      -- component separator needs to have fg = current_section.bg
+      -- and bg = adjacent_section.bg
+      local previous_section = status_builder[i-1] and status_builder[i-1] or {}
```
hoob3rt, line 197: I still don't get what's this doing. I'm guessing
```
previous_section = status_builder[i-1] if status_builder[i-1] else {}
```
right?

lua/lualine.lua
```diff
@@ -148,34 +150,66 @@ end
...
+      local current_section = status_builder[i]
+      local next_section = status_builder[i+1] and status_builder[i+1] or {}
+      -- For 2nd half we need to show separator before section
+      if current_section[1] > 'x' then
+        local transitional_highlight = highlight.get_transitional_highlights(previous_section[2], current_section[2], true)
```
hoob3rt, line 202: `get_trainsitional_highlight` is defined as `function(left_section, right_section)` yet you are passing section's data here, confusing code

lua/lualine.lua
```diff
@@ -148,34 +150,66 @@ end
...
+  local section_sequence = {'a', 'b', 'c', 'x', 'y', 'z'}
+
+  for _, section_name in ipairs(section_sequence) do
+    if sections['lualine_'..section_name] then
+      -- insert highlight+components of this section to status_builder
+      local section_highlight = highlight.format_highlight(is_focused,
+                   'lualine_'..section_name)
+			local section_data = utils_component.draw_section(sections['lualine_'..section_name], section_highlight)
+			if #section_data > 0 then
+				table.insert(status_builder, {section_name, section_data})
```
hoob3rt, line 179: I mentioned this before. It would be way better to have `status_builder` as a map
`{name = section_name, data=section_data}`
this will make it a lot easier to read down. instead of having `[1]` `[2]` everywhere

lua/lualine/utils/utils.lua
shadmansaleh: I'm freeing unessery values so vim.tbl_extend doesn't put them in . They have already been loaded in the list  with ctermfg and ctermbg name

lua/lualine.lua
shadmansaleh: Yes

lua/lualine.lua
shadmansaleh: I first tried map But maps doesn't store sequencial data . When cection exists and which setion comes after what is nessary for this to work

lua/lualine.lua
shadmansaleh: >I think the configuration will get easier if you want a static separator like it is right now and also make code in component.lua a lot easier.

I'm not sure how is it posible . Adding type checking to determine which type of configuration is being used will only increse complexity

lua/lualine.lua
hoob3rt: I'm talking about an array of maps.... Instead of `status_builder[i][2]` we can have `status_builder[i].data`. way cleaner

lua/lualine/utils/utils.lua
hoob3rt: I don't get it. What's the point of cterm_colors then? Seems to me like you get it just to clear it...

lua/lualine.lua
hoob3rt: Okay, can  you change it to a and if statement then? It will be more intuitive

lua/lualine.lua
shadmansaleh: >function(left_section, right_section) yet you are passing section's data here, confusing code

I ment section data . I'm not sure if ypu thought it was section name . But I'm basicaly parseing the data to figure out which highlight was used in the data so I can genarate the transitional highlight

hoob3rt: My point being that if you are passing section data then an argument to the function should be called `section_data` not `section`

lua/lualine.lua
shadmansaleh: Ok that's posible

lua/lualine.lua
shadmansaleh: >separator = { component = {'something', 'something'}, section = {'something', 'something'}

donn't you think that's abit too nested for an option? 
lualine.options.separator.section[1] ?

I can add single string to set both to same separator but that will be processed while loading components. Adding type checking wverywhere separators are used will increse complecity...

hoob3rt: Okay, let's leave it for now. I will think about it. Let's fix the other issues first

lua/lualine.lua
shadmansaleh: simplified

lua/lualine.lua
shadmansaleh: renamed arguments

lua/lualine.lua
shadmansaleh: applied

lua/lualine.lua
shadmansaleh: kept component_separators = '|'

lua/lualine.lua
hoob3rt: nice! looks way better now

lua/lualine/utils/utils.lua
shadmansaleh: I'm returning a table with
{ctermfg, ctermbg, guifg, guibg} + what ever parameters are set by gui and cterm sections (if available).

lua/lualine.lua
hoob3rt: lgtm

lua/lualine/utils/utils.lua
shadmansaleh: in line 23-25 I've alrwady loaded those values to The tavle I'll return(color)

shadmansaleh:
Thats weird did you change the indent count how did that many indebtations occur....

hoob3rt:
I mean, the indent count has always been 2 in this project. I just pressed gg=G

shadmansaleh:
>I mean, the indent count has always been 2 in this project. I just pressed gg=G

There are maybe bugs in ft-plugin indents are wrong from statusline function

hoob3rt:
yea, that's a bug.
```lua
  if not half_passed then
    table.insert(status, highlight.format_highlight(is_focused,'lualine_c').."%=") end
```
changing it now

lua/lualine.lua
hoob3rt: Okay, here's an idea. It is redundant to have both `separator` and `component_separators` as an option. Maybe change it to `component_separator` only? Then do `if #component_separator == 1` etc in `component.lua`

shadmansaleh:
ignore that waring . not having local causes a bug .

hoob3rt:
> ignore that warning . not having local causes a bug . I've a comment above that line stating that .

that's weird. okay

lua/lualine/utils/utils.lua
hoob3rt: then why do we have to extend `cterm_colors` if it contains nothing? Won't this yield the same results?
```lua
  color = vim.tbl_extend('keep', color, gui_colors)
```

shadmansaleh:
>that's weird. okay

yeah it is.I've no idea why either but that fixed it 😑

lua/lualine.lua
shadmansaleh: There's no separator option.

I'm just turning the component_separations into a table with two copies of same separator if a string is given insted of a table.

I think you missunderstood

lua/lualine/utils/utils.lua
shadmansaleh: >then why do we have to extend cterm_colors if it contains nothing? Won't this yield the same results?

cterm_colors table can cointain other flags like bold=true, reverse=true . That's why I'm useing tbl_extend. And removing the foreground and background entry as they have already been included in different name.

shadmansaleh:
So can this be  merged?

hoob3rt:
Yea, lgtm now

shadmansaleh:
So which ones next?

hoob3rt:
I will finish diagnostics in the coming days. You can take a look at #100 if you've got some spare time

PR #95, opened 2021-02-15
feat: diagnostics api

hoob3rt:
(WIP) adds diagnostics component
* display diagnostic count from multiple sources
* display line number for the closest diagnostic of each type


lua/lualine/components/diagnostics.lua
```diff
@@ -0,0 +1,41 @@
+-- Copyright (c) 2020-2021 hoob3rt
+-- MIT license, see LICENSE for more details.
+
+local function diagnostics(options)
+  return function()
+    local error_count, warning_count, info_count = 0,0,0
+    if options.sources~=nil then
+      for _, source in ipairs(options.sources) do
+        if source == 'lsp' then
```
shadmansaleh, line 9: I think a nicer way to implement it would be a look up table containing option
Like say:
```lua
local sources = {
  nvim_lsp = function()
    local error_count = vim.lsp.diagnostic.get_count(0, 'Error')
    local warning_count = vim.lsp.diagnostic.get_count(0, 'Warning')
    local info_count = vim.lsp.diagnostic.get_count(0, 'Info')
    return error_count, warning_count, info_count
  end
}
```
Then we can just call the sources with
```lua
local E, W, I = sources[source]()
```

I actualy thought of doing it this way . That way I think code will
be cleaner make it easier to expand . Let me know what you think

lua/lualine/components/diagnostics.lua
```diff
@@ -0,0 +1,41 @@
...
+          error_count = error_count +  vim.lsp.diagnostic.get_count(0, 'Error')
+          warning_count = warning_count +  vim.lsp.diagnostic.get_count(0, 'Warning')
+          -- info_count = info_count +  vim.lsp.diagnostic.get_count(0, 'Info')
+        end
+      end
+    end print(error_count)
+    local result = {}
+    local symbols = {
+      'E',
```
shadmansaleh, line 18: How about some nice icons too?

lua/lualine/components/diagnostics.lua
```diff
@@ -0,0 +1,41 @@
...
+    print(result[1])
+    if result[1] ~= nil then
+      return table.concat(result, '')
+    else
+      return ''
+    end
+  end
+end
+
+return { init = function(options) return diagnostics(options) end }
```
shadmansaleh, line 41: I think it will better if we also expose error,warn,info counts separately
as well . That will make it easier to make custom diagnostics components.
It was suppose to provide a api wasn't it ?

shadmansaleh (changes requested):

lua/lualine/components/diagnostics.lua
```diff
@@ -0,0 +1,41 @@
+-- Copyright (c) 2020-2021 hoob3rt
+-- MIT license, see LICENSE for more details.
+
+local function diagnostics(options)
+  return function()
+    local error_count, warning_count, info_count = 0,0,0
+    if options.sources~=nil then
+      for _, source in ipairs(options.sources) do
+        if source == 'lsp' then
+          error_count = error_count +  vim.lsp.diagnostic.get_count(0, 'Error')
```
shadmansaleh, line 10: I donn't think adding the numbers is a good idea. As multiple sources most
likely have common errors . How about displying all of them . Like if 2 sources
are configured then show like this `E: 5|4`

lua/lualine/components/diagnostics.lua
```diff
@@ -0,0 +1,41 @@
...
+      'I'
+    }
+    local data = {
+      error_count,
+      warning_count,
+      info_count
+    }
+    for range=1,3 do
+      if data[range] ~= nil and data[range] > 0 then
+      table.insert(result,symbols[range]..':'..data[range]..' ')
```
shadmansaleh, line 29: > data[range]..' '

Insted of appending space in the end. use space as an argument for table.concat() in line 34

lua/lualine/components/diagnostics.lua
shadmansaleh, line 9: lsp probably should be nvim_lsp as other cources like coc/ale uses lsp too

lua/lualine/components/diagnostics.lua
coxackie: from https://www.nerdfonts.com/cheat-sheet:
error -> nf-mdi-close_circle_outline (f659)
warning -> nf-mdi-alert_outline (f529)
info/hint -> nf-mdi-information_outline (f7fc)

Also, colouring is important - preferably: errors red; warnings yellow; info/hint white.

lua/lualine/components/diagnostics.lua
hoob3rt: Yep, we are on the same page here. This is the same way I intend to have it implemented but thanks for pointing that out

lua/lualine/components/diagnostics.lua
hoob3rt: Yep, coloring will be similar to #61, decent out of the box with further customization available

lua/lualine/components/diagnostics.lua
hoob3rt: Yep, that's the idea

lua/lualine/components/diagnostics.lua
hoob3rt: Well, I don't agree with this one. If you have multiple sources of errors showing the same ones then what's the point of even having multiple sources.

shadmansaleh: I donn't have multiple sources . But wonn't they pick up same errors?

shadmansaleh:
If you donn't mind I might push some changes to this branch to speed things up . Though I'm not sure if I can push to it....

hoob3rt:
> If you donn't mind I might push some changes to this branch to speed things up . 

I will do some things here right now, after I finish reviews so hold on to that.

> Though I'm not sure if I can push to it....

I don't think you can. You can always pr your branch into my diagnostic branch

shadmansaleh:
>I don't think you can. You can always pr your branch into my diagnostic branch

I can do that.

lua/lualine/components/diagnostics.lua
shadmansaleh: I'll implement and create a pr soon

lua/lualine/components/diagnostics.lua
hoob3rt: Well I use typescript a lot and while tsserver is nice for catching syntax errors I need eslint to catch styling errors.

shadmansaleh: donn't eslint catch syntax errors . Pardon my ignorance I haven't used eslint.

hoob3rt: Well eslint is just and example, I've worked with a few linters which don't catch syntax errors and still had to use both lsp and ale. Eslint can catch whatever you configure it to catch :)

shadmansaleh: > Eslint can catch whatever you configure it to catch

Then doesn't it make more sence to show output of separate sources separately?

shadmansaleh:
@hoob3rt if you merge #101 to here I think most functionalities are done .

I've added:
- coc.nvim
- ale
- api for diagnostics
- colors
- icons
- options.color_error kind of options
- options.section [to display specific sections like error & warning]

I think that's about it .

>display line number for the closest diagnostic of each type

I'm not sure what this is about .

shadmansaleh:
Are you sure abiut these 'diffNewFile', 'diffRemoved', 'Normal' Those groups aren't dedicated for this . Might look weird on some colorschemes....

shadmansaleh:
Ok 'diffNewFile' these isn't available in my colorscheme(nightfly) and causes lualine to crash when diagnostics component is loaded. 

Also opening neovim in default state with `nvim -u NONE` 'diffRemoved' hl group isn't present either .

personaly I'd like my error,warn highlights to stay constant across colorscheme.

lua/lualine/components/diagnostics.lua
hoob3rt: No, I prefer it this way

hoob3rt:
> Ok 'diffNewFile' these isn't available in my colorscheme(nightfly) and causes lualine to crash when diagnostics component is loaded.
> 
> Also opening neovim in default state with `nvim -u NONE` 'diffRemoved' hl group isn't present either .
> 

Good catch. I fixed it here and in signify component

> personaly I'd like my error,warn highlights to stay constant across colorscheme.

You can still set custom colors for each of these sections. no issue here

shadmansaleh:
>You can still set custom colors for each of these sections. no issue here

Ok . By the way what's left to do in this pr ?

hoob3rt:
I was thinking of providing a was to pass a custom function as a diagnostic source but couldn't get this to work. I will update README.md and merge this in current version. We can discuss this idea further in later issues

shadmansaleh:
You didn't update the docs/lualine.txt

hoob3rt:
Oh yeah, can you take care of that? I didn't look into how to do it automatically

PR #98, opened 2021-02-17
Show valid filenames on inactive windows

delphinus:
With #60, it shows the same filename for all windows. This PR fixes it.


shadmansaleh:
The main issue was untroduced by #79 . It affects  
all components using vim.b/ vim.w like scopes to access data . I'm trying a more genaral purpose solution to the big #97 you can try it out and see it solves your issue too :)

delphinus:
Great. I've found this is fixed, so I'll close.

PR #102, opened 2021-02-18
(HotFIX) Fix tags not wrking properly in docs

shadmansaleh:
'-' was acting as a word separator so replaced it with '_'


shadmansaleh:
@hoob3rt do merge it as soon as you can.

This one just changes '-' to '_' in tag_names in doc file.

PR #104, opened 2021-02-20
feature: Add theme auto genaration from colorscheme

shadmansaleh:
Improving @kxmndev pr #26 . Worked farely well in my tests :)

run this command
```vim
:au ColorScheme * lua package.loaded['lualine.themes.auto'] = nil ; require('lualine').options.theme = 'auto'
```

Now change colorschemes with coloscheme command and enjoy


hoob3rt:
This works surprisingly well. Awesome!

> ```viml
> :au ColorScheme * lua package.loaded['lualine.themes.auto'] = nil ; require('lualine').options.theme = 'auto'
> ```

Shouldn't this be the default?
```
if theme == 'auto' then exec(this autocommand)
```

shadmansaleh:
>Shouldn't this be the default?

Then no on is turning that autocmd off When theme is nolonger auto . also I doubt any one change colorscheme 2 times on average in a vim season . I just gave that command to make testing easier.

hoob3rt:
> Then no one is turning that autocmd off When theme is nolonger auto .

About that. I think we should ditch the theme change in runtime since no other option responds to runtime changes and it's not documented anywhere in the readme so I doubt anyone is using it. What do you think?

shadmansaleh:
>About that. I think we should ditch the theme change in runtime since no other option responds to runtime changes and it's not documented anywhere in the readme so I doubt anyone is using it. What do you think?

Do you know how much work I had too put to fix highlights for those runtime changes last week 😔
. Would have hapily diched it last week if you had told before . Now I donn't feel like udoing that much . You can do it if you want to . Else just document it 😄

shadmansaleh:
>no other option responds to runtime changes.

It's now exactly true . Component separators respond, components can be removed on runtime though can't load new components . But that's because we only load compoents at start . If we load the conponent when section changes The component options could be changed in runtime too . Do you wana go that route?

hoob3rt:
> Do you know how much work I had too put to fix highlights for those runtime changes last week 😔
. Would have hapily diched it last week if you had told before . Now I donn't feel like udoing that much . You can do it if you want to

Sorry man 😢   I will remove the option to change colorscheme at runtime tomorrow.
> > no other option responds to runtime changes.
> 
> It's now exactly true . Component separators respond, components can be removed on runtime though can't load new components . But that's because we only load compoents at start . The component options could be changed in runtime too . Do you wana go that route?

No. There is no need to have runtime changes support. It only adds unnecessary complications to the codebase.

This also requires change in `THEMES.md`. I would personally put `auto` as the first entry and instead of screenshots place a text. Something like 
```
if your favourite colorcheme is not yet supported you can use `auto` to make lualine match your colorscheme
```

PR #112, opened 2021-02-22
Customize symbols for the diagnostics component

delphinus:
Enable to set symbols for diagnostics.


shadmansaleh:
Nice idea . I think it should support setting single symble and leaving rest to default . Like if I want to change the error symbol I and I set only the property symbol.error warn and info should use default symbols ....

delphinus:
I see. I'll rewrite with `vim.tbl_extend()` and enable to overwrite a part of symbols.

hoob3rt:
lgtm, thanks

PR #121, opened 2021-02-27
Fix Dracula theme, make it similar to others

NFrid:
Dracula colorscheme was poorly made. I fixed command and inactive modes. Also renamed gray color variables to the right ones. Take a look.
image
image


hoob3rt:
lgtm, thanks

PR #126, opened 2021-03-10
feat: changed extension structure & added extensions

hoob3rt:
- [x] Change extensions to follow lualine structure
- [X] Update fzf extension accordingly
- [x] Fugitive extension
- [x] nerdtree extension
- [ ] add screenshots of all updated extensions

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -35,8 +35,32 @@ M.inactive_sections = {
   lualine_z = {  }
 }
 
-M.extensions = {
-}
+M.extensions = {}
+
+local function apply_configuration(config_table)
+  if not config_table then return end
+  local function parse_sections(section_group_name)
+    if not config_table[section_group_name] then return end
+    for section_name, section in pairs(config_table[section_group_name]) do
+      M[section_group_name][section_name] = config_table[section_group_name][section_name]
+      if type(section) == 'table' then
+        for _, component in pairs(section) do
+          if type(component) == 'table' and type(component[2]) == 'table' then
+            local options = component[2]
+            component[2] = nil
+            for key, val in pairs(options) do
+              component[key] = val
+            end
+          end
+        end
+      end
+    end
+  end
+  parse_sections('options')
+  parse_sections('sections')
+  parse_sections('inactive_sections')
+  if config_table.extensions then M.extensions = config_table.extensions end
+end
 
 local function check_single_separator()
   local compoennt_separator = M.options.component_separators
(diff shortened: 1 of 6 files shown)


lua/lualine.lua
```diff
@@ -210,7 +204,27 @@ local function lualine_set_theme()
   theme_set = M.options.theme
 end
 
-local function statusline(sections, is_focused)
+local function statusline(is_focused)
+  local sections = nil
+  for _, extension in ipairs(M.extensions) do
```
shadmansaleh, line 209: You should handle tabline too . It's not handled currently and will break the feature.
Actualy I think it's better to move these logic to `status_dispatch` function . I had left that function for this purpose . Also this function is already doing too much .

lua/lualine.lua
```diff
@@ -210,7 +204,27 @@ local function lualine_set_theme()
...
+    for _, filetype in ipairs(extension.filetypes) do
+      if vim.bo.filetype == filetype then
```
shadmansaleh, line 211: Is it not possible for a filetype to have more then one extention ? currently only the extention that matches at last will be displayed . Maybe we should extend the table instead of replacing it. Maybe we can provide a flag for extention . if the extention sets it to true we can display the extention alongside regular statusline instead of it completely replacing the statusline . What do you think ?

hoob3rt: > Is it not possible for a filetype to have more then one extention ? 

well my idea for extensions is that a lot of plugins use buffers in which a user does not need a regular statusline but a speciefied one. This is the aim of extensions. I don't see why you would need a lot of extensions for a single filetype in that case. 

> if the extention sets it to true we can display the extention alongside regular statusline instead of it completely replacing the statusline . What do you think ?

I don't like this. At this point this is a plugin component and not an extension

lua/lualine.lua
hoob3rt: Yeah, true. I will refactor it today. Thanks for the review :smile:

lua/lualine.lua
shadmansaleh: Makes sence . I think I naver really got the idea of extentions :)

hoob3rt: Yeah, maybe the naming is weird. I didn't think too much about it just copied the name from what airline is using for this. If you have any better suggestions then lmk. We can also change this in this pr

lua/lualine.lua
hoob3rt: This should be good now

lua/lualine.lua
```diff
@@ -293,16 +274,45 @@ local function statusline(is_focused)
...
+    if inactive_sections == nil then
+      inactive_sections = M.inactive_sections
+    end
+    return statusline(inactive_sections, false)
   end
 end
 
 local function tabline()
-  return statusline(M.tabline, true)
+  local extension_sections = get_extension_sections()
```
shadmansaleh, line 310: This doesn't make sense here . If an extention is used itwill apear both on statusline and tabline . It will conflict with 3rd party tablines and default tabline ( tabline was suppose to be purely o[t in feature) . Also it's weird if same thing apearing both at top and bottom . Also I was wondering if by your defination of extention will they ever need complexcity of separate config fro inactive an active section ? None of the examples do . We can probably just show same section for active and inactive state . Otherwize lgtm :).

hoob3rt: you are right, they can be simplified

PR #127, opened 2021-03-13
testing

hoob3rt:
WIP
research lualine testing through `busted`
Probably also a lot of refactoring

The things which should be tested here are:
- [x] most util functions 
- [ ] proper component loading
- [ ] component return values (defaults)
- [ ] all global & component options 
- [ ] probably the whole statusline string should be tested, separators tend to break

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,26 @@
+name: Lint Code Base
+
+on:
+  push:
+    branches-ignore: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    name: Lint Code Base
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          # Full git history is needed to get a proper list of changed files within `super-linter`
+         fetch-depth: 0
+
+      - name: Lint Code Base
+        uses: github/super-linter@v3
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: master
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,22 @@
+name: Linting
+
+on:
+  push:
+    branches-ignore: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build:
+    name: Luacheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install linter
+        run: |
+          sudo apt-get update
+          sudo apt-get install luarocks
+          sudo luarocks install luacheck
+      - name: Lint
+        run: make lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,10 +1,6 @@
 name: Linting
 
-on:
-  push:
-    branches-ignore: [master]
-  pull_request:
-    branches: [master]
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,10 @@
 name: Linting
 
-on: [push, pull_request]
+on: 
+  push:
+    branches-ignore: [master]
+  pull_request:
+    types: [opened, reopened]
 
 jobs:
   build:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,17 +10,13 @@ jobs:
   build:
     name: Lint Code Base
     runs-on: ubuntu-latest
-
     steps:
-      - name: Checkout Code
-        uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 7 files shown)


shadmansaleh:
Maybe we should use plenary test. It will probably be easier to setup. What do you think?

hoob3rt:
> Maybe we should use plenary test. It will probably be easier to setup. What do you think?

I truly don't know. I have not used any types of testing for nvim yet. I will take a look into it. I just saw that telescope uses `busted` and it looks like a nice framework

shadmansaleh:
I'm pretty sure telescope uses plenary test

hoob3rt:
> I'm pretty sure telescope uses plenary test

Yeah actually they use both. We should probably use both as well in proper places

shadmansaleh:
So what's decided are we using plenary or busted ?

My arguments for plenary would be
- It runs in nvim environment so has access to lua-stdlib, api etc . Unlike busted that runs in separate lua instence . If we are to test with busted we'll have to make mock for those (Feels like insane amount of works when plenary can already do the heavy lifting)
- There's a chanche plenary test might be merged to nvim core

I've checked telescpope repo they are just using plenary . Plenary imatates the syntax of busted that's why you might have thought busted was used .

Also this branch needs a rebase on master . Everything seems outdated😁

hoob3rt:
I agree with plenary. I did some research a while ago and came to the same conclusion as you. I want to start working on this now when I have free time so lmk if you have an idea on what tests we should have. I will update the description with mine now.

shadmansaleh:
> lmk if you have an idea on what tests we should have. 

I've basically 0 experience writting tests . I've written a few tests small python projects. And a few for prs in neovim . Really I'll have learn writting tests first before I can suggest what tests can we have 🤣

shadmansaleh:
Can I push to this branch ?

hoob3rt:
> Can I push to this branch ?

yeah go ahead, we are gonna squash this one anyways

shadmansaleh:
Last commit I pushed can be merged directly to master . should I leave it in this branch or create a separate pr @hoob3rt ?

hoob3rt:
> Last commit I pushed can be merged directly to master . should I leave it in this branch or create a separate pr @hoob3rt ?

probably better to cherry-pick into a new branch and create a new pr, no idea when this will be ready

Makefile
```diff
@@ -1,11 +1,14 @@
...
-	luacheck lua/lualine
-	luacheck tests
+	@luacheck lua/lualine
+	@luacheck tests
 
 format:
-	for file in `find -name '*.lua'`;do lua-format $$file -i; done;
+	@for file in `find -name '*.lua'`;do lua-format $$file -i; done;
 
 test:
```
shadmansaleh, line 10: @hoob3rt is there any difference in making this line?
```
    test: lint
```
instead of @make lint in following line ?

hoob3rt: > @hoob3rt is there any difference in making this line?
> 
> ```
>     test: lint
> ```
> 
> instead of @make lint in following line ?

yes there is, your solution is better :smile:

.github/workflows/lint.yml
```diff
@@ -10,17 +10,13 @@ jobs:
...
-          VALIDATE_ALL_CODEBASE: false
-          DEFAULT_BRANCH: master
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} 
+      - name: Install linter
+        run: |
+          sudo apt-get update
+          sudo apt-get install luarocks
+          sudo luarocks install luacheck
+      - name: Lint
+        run: luacheck lua/
```
shadmansaleh, line 22: Run `make lint`

hoob3rt:
@shadmansaleh  I would clean this pr up with what we currently have and merge, then I will create issues with appropriate tests. This way we will have a basic test setup for pr's and a good battle plan for the rest of tests

shadmansaleh:
ok . I was thinking of asking you to do a paetial merge . Add a workflow with `make test` :)

shadmansaleh:
You need to pull neovim nightly from releases + clone plenary and lualine in same directory

shadmansaleh:
we can copy paste most of it from [here](https://github.com/nvim-telescope/telescope.nvim/blob/master/.github/workflows/ci.yml) I think

hoob3rt:
> we can copy paste most of it from [here](https://github.com/nvim-telescope/telescope.nvim/blob/master/.github/workflows/ci.yml) I think

yeah I was looking at it as well, I will look into it

shadmansaleh:
I think I added the functionality to master and added the test to here without the functionality . That's why the test was failing :P

hoob3rt:
This looks good. I have an issue running tests on my local machine though and can't figure out why, have you encountered this error before?
```
/home/hubert/lualine.nvim/tests/spec/utils_spec.lua:83: attempt to call upvalue 'build_component_opts' (a nil value)
```

shadmansaleh:
Naver seen it before . Are you running with Makefile while at root of repo ?

hoob3rt:
> Naver seen it before . Are you running with Makefile while at root of repo ?

yeah, happens on both mac and linux, anyway if tests are running on your machine well and in CI then you can merge this one, I will debug this locally later

PR #129, opened 2021-03-13
fix bug component highlights not blending with bg

shadmansaleh:
changes:
- fix logical error in highlight.lua
- refactor mode component
- add tags to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,5 @@ luac.out
 *.x86_64
 *.hex
 
+# ctags
+tags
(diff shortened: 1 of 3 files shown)


hoob3rt:
I agree with all of these changes. But I would like to have them in separate commits as they refer to completely different things. Can you change it to be this way?

shadmansaleh:
Ok done . They would have gotten squashed during merge and were quite simple that's why I didn't bother to separate them my bad :)

hoob3rt:
They are different enough that I didn't want to squash them. Thanks :smile:

PR #130, opened 2021-03-14
refactor: removed ability to set theme in runtime

hoob3rt:
@shadmansaleh can you take a look at that and confirm that this is everything that needs to be removed?

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -2,12 +2,9 @@
 -- MIT license, see LICENSE for more details.
 
 local utils_component = require('lualine.utils.component')
-local utils = require('lualine.utils.utils')
 local highlight = require('lualine.highlight')
 
 local M = { }
-
-local theme_set = {}
 
 M.options = {
   icons_enabled = true,
@@ -149,7 +146,6 @@ local function component_loader(component)
         component.color, component.component_name, component)
       end
       update_color()
-      utils.expand_set_theme(update_color)
     end
   end
 end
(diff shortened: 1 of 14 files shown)


lua/lualine.lua
```diff
@@ -149,7 +146,6 @@ local function component_loader(component)
         component.color, component.component_name, component)
       end
       update_color()
-      utils.expand_set_theme(update_color)
```
shadmansaleh, line 152: You shouldn't remove expand_set_theme from anywhere . They are there to load the colors on ColorScheme event . So if users change colorscheme the highlights remain . As all colorscheme at start remove all set highlights. It's unrelated to lualine themes . Previously we were doing a separate autocmd for each time we set highlights . That's why I had added that function to simplify cutom highlights so we can do it with just one autocmd (That is added in set_autocomand func(or something like that).

lua/lualine.lua
```diff
@@ -191,29 +187,11 @@ end
...
-          if type(component) == 'table' then
-            component.theme = M.options.theme
-          end
-        end
-      end
-    end
-    reset_component_theme(M.sections)
-    reset_component_theme(M.inactive_sections)
   end
-  utils.clear_highlights()
```
shadmansaleh, line 208: We should still clear the highlights before creating the new ones I think

shadmansaleh:
I actually not clearly remember what I had added just for this . I was doing several things at once so   memory is rather tangled :D . I think I added some edge case handlings for this to work . Some of them were probably also required for component separator highlights . I don't think anything outside of init.lua requires a change for this .

lua/lualine.lua
hoob3rt: Yeah, I thought this will only run once but sourcing a config file might execute this an additional time. I will revert it

PR #131, opened 2021-03-15
Update CONTRIBUTING.md

joshuachp:
Fix setting the custom theme documentation from `lualine.theme` to `lualine.options.theme`.


hoob3rt:
good catch, thanks

PR #140, opened 2021-03-23
fix typo

JayceFayne:


hoob3rt:
Thanks for catching this. It will be gone in 2 days but we can have proper english on errors for that time 😄

Issue #44, opened 2021-01-07
Create plugin extension for chadtree


hoob3rt:
Statusline in chadtree window should display chadtree's directory.

PR #145, opened 2021-03-26
Added chadtree extension

samirettali:
Hi!
I saw issue #44 about chadtree extension, it turns out I had implemented it a couple days ago so here is it.

It shows the current directory with a `~` replacing the user's home, I don't know if there are better ways to do it.

Hope it's useful :)

### DIFF
--- a/lua/lualine/extensions/chadtree.lua
+++ b/lua/lualine/extensions/chadtree.lua
@@ -1,0 +1,17 @@
+os = require('os')
+
+local function short_cwd ()
+  local cwd = os.getenv('PWD')
+  local home = os.getenv('HOME')
+  return cwd:gsub(home, '~')
+end
+
+local M = {}
+
+M.sections = {lualine_a = { short_cwd }}
+
+M.inactive_sections = M.sections
+
+M.filetypes = {'CHADTree'}
+
+return M


lua/lualine/extensions/chadtree.lua
```diff
@@ -0,0 +1,17 @@
+os = require('os')
+
+local function short_cwd ()
+  local cwd = os.getenv('PWD')
+  local home = os.getenv('HOME')
+  return cwd:gsub(home, '~')
```
hoob3rt, line 6: I think we can get away without calling `os` since vim already caches the working directory.

Try this in command mode
```
:substitute(getcwd(),$HOME,'~','')
```

hoob3rt (commented):
> It shows the current directory with a ~ replacing the user's home, I don't know if there are better ways to do it.

No idea how I have not come up with this while doing the `nerdtree` extension. Nice :smile:

shadmansaleh:
>I don't know if there are better ways to do it.

There is a filename modifier for this

```lua
vim.fn.fnamemodify(cwd, ':~')
```
to checkout all of them `:h filename-modifiers` 

Also os library is loaded by default you don't have to require it . What I'm really wondering is does chadtree update the $PWD when going through directories in chadtree . I haven't really ever used chadtree . But vim plugin changing shell variable feels unexpected to me :) (NerdTree doesn't do it it seems)

PR #152, opened 2021-03-31
forest_night -> everforest

evantravers:
sainnhe changed the name of his repo and theme: https://github.com/sainnhe/everforest/issues/43


hoob3rt:
Thanks, I probably wouldn't have noticed

shadmansaleh:
Probably should have been. merged with notification and buffer period . Like other breaking changes we made . currently anyone having forest_night in their configs will crash lualine and create errors on startup not s fun experinece .

shadmansaleh:
It's really simple .
You can just put these line in forest_night.lua
```lua
print('Whatever warning')
return require'lualine.themes.everforest'
```

It's useualy how we do breaking changes .

hoob3rt:
I didn't do it looked to me like the creator of the theme broke `colo forest_night` himself so everyone who uses this theme should be aware of this but I agree that doing it how we do it normally would be cleaner

PR #141, opened 2021-03-23
Refactor: Classify components

shadmansaleh:
The idea is to create components in somewhat oop style . The component.lua file added willa ct as a base class and components will extend it by mainly implementing update_status() function.

@hoob3rt should I go forward whith this ?

### Todo
- [x] Add component base class
- [x] Migrate components to this structure
- [x] Update init.lua + section.lua to use this system
- [x] Change special and functiom components to this format .
- [x] Remove all the unesseary stuff from utils and init.lua
- [x] Testing (I think I'm satisfied with the testing. You can report bugs if you find any)

PR #153, opened 2021-03-31
Enhance theme

shadmansaleh:
Changes:
- loaded theme table is nolonger stored in 
  configs and thus not passed
  to components. Insted it's cached in highlights.lua
- Reintroducing clear_highlights turns out themes can still change if
  setup is called more than once
- Now lualine doesn't crash from invalid theme name

These changes are actually better out of #141 luckily cherry-pick went easy on me this time .
There's one more line of highlight change that I would like to bring here but int's too deep . I might look into it later if #141 isn't merged by then :)

### DIFF
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -272,8 +272,13 @@ local function setup_theme()
     local ok
     ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
     if not ok then
-      vim.api.nvim_echo({{'theme ' .. theme_name .. ' not found defaulting to gruvbox', 'ErrorMsg'}}, true, {})
-      theme = require'lualine.themes.gruvbox'
+      vim.api.nvim_echo({
+        {
+          'theme ' .. theme_name .. ' not found defaulting to gruvbox',
+          'ErrorMsg'
+        }
+      }, true, {})
+      theme = require 'lualine.themes.gruvbox'
     end
   end
   highlight.create_highlight_groups(theme)
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -267,9 +267,8 @@ local function tabline() return statusline(config.tabline, true) end
 
 local function setup_theme()
   local theme_name = config.options.theme
-  local theme
+  local ok, theme
   if type(theme_name) == 'string' then
-    local ok
     ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
     if not ok then
       vim.api.nvim_echo({
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -266,9 +266,15 @@ end
 local function tabline() return statusline(config.tabline, true) end
 
 local function setup_theme()
-  local theme = config.options.theme
-  if type(theme) == 'string' then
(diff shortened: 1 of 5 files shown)


shadmansaleh:
@hoob3rt If you have some time to review today . can you take a look at this one. It's fairly small. And you probably won't find issue with it . Also I'll be fairly free today . So if you've time we can do a bit of work on #141 😃

lua/lualine/init.lua
```diff
@@ -266,10 +266,22 @@ end
 local function tabline() return statusline(config.tabline, true) end
 
 local function setup_theme()
-  if type(config.options.theme) == 'string' then
-    config.options.theme = require('lualine.themes.' .. config.options.theme)
+  local theme_name = config.options.theme
+  local theme
+  if type(theme_name) == 'string' then
+    local ok
+    ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
```
hoob3rt, line 273: this is weird... Wont' having `local ok, theme = ...` the same effect?

hoob3rt (commented):
Make sure that lua-format is applied to all files. After that lgtm

lua/lualine/init.lua
shadmansaleh: Theme is required out of the if scope . If i do `local ok, theme = ..` it will create a new var named theme and shadow the one from outer scope .

shadmansaleh:
> Make sure that lua-format is applied to all files. After that lgtm

I've applied lua-format on last commit

lua/lualine/init.lua
hoob3rt:
```lua
  if type(theme_name) == 'string' then
    local ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
    if ok then
      highlight.create_highlight_groups(theme)
    else
      vim.api.nvim_echo({
        {
          'theme ' .. theme_name .. ' not found defaulting to gruvbox',
          'ErrorMsg'
        }
      }, true, {})
      local gruvbox = require 'lualine.themes.gruvbox'
      highlight.create_highlight_groups(gruvbox)
    end
  end
```

shadmansaleh: I want to allow custon themes to be directly passed though theme option  . Currently it's easily posible to require a theme file change some colors and use that option theme to that table and have it work . Your implementation will break it . as table themes aren't used

shadmansaleh: Will this work for you ?

```lua
local ok, theme
  if type(theme_name) == 'string' then
    ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
```

hoob3rt: oh right, I see what you mean. Your latest suggestion is the best one so we can stick with it

shadmansaleh: Ok I've made the changes.

hoob3rt:
Yeah lgtm now, sorry for lagging with reviews. I have a lot of stuff going on atm 😕

shadmansaleh:
👍 Are we getting #141 too?

hoob3rt:
> 👍 Are we getting #141 too?

Not right now unfortunately as stated in another pr

PR #155, opened 2021-03-31
Fix displaying relative paths.

### LINKED ISSUE #154: Bug: Displayed relative paths could be wrong in some cases
It looks like there are some cases that relative file paths displayed by lualine are wrong.
I set `full_path = true, shorten = true` for the `filename` component, and I still see the absolute path of a file opened from [vim-startify](https://github.com/mhinz/vim-startify).
I believe [this post](https://stackoverflow.com/a/24463362/2182622) explains why it happens and how to fix it.

jaekyeom:
Fixes #154

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -11,7 +11,7 @@ local function filename(options)
     if not options.full_path then
       data = vim.fn.expand('%:t')
     elseif options.shorten then
-      data = vim.fn.expand('%')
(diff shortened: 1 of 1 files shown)


jaekyeom:
@hoob3rt 
Thanks for the merging.
Btw, I've realized that making file paths that are not under the current working directory relative to the home directory (by modifier `:~`) might be a side effect that you didn't want.
Is it okay with you?

PR #160, opened 2021-04-06
Miscellaneous changes

shadmansaleh:
### Changes

1. Add condition option
I quite like the idea of conditions from galaxyline . This pr adds condtion option . It works similer to galaxylines condition . It takes a function . And if the function returns true the component is displayed otherwise it isn't .
2. Also I've updated the options section of docs . What I wrote last time was too confusing even for me 😅 . These changes are on separate commits so don't sqash them .
3. color option now can take a highlight group name insted of bg/fg table
4. evil_lualine screenshot
5. mention lightline2lualine_converter in CONTRIBUTING.md
6. load cterm colors only when termguicolors is disabled to reduce startup overhead
7. Now functions are no longer exposed as global variable
8. Cleanup eval_lua component and document it
9. Document theme tweeking
10. Genarate help tags . #159 missed that .

This is based on #141 . So I'll rebase this on master when #141 is merged :) .

@hoob3rt If you're reviewing it . Check diff of 2nd and 3rd commit as 1st commit is just #141

I'm wondering should we provide some conditions ourselves like [galaxyline](https://github.com/glepnir/galaxyline.nvim/blob/main/lua/galaxyline/condition.lua) .

### DIFF
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -251,19 +251,17 @@ end
 
 local function set_tabline()
   if next(config.tabline) ~= nil then
-    _G.lualine_tabline = tabline
-    vim.o.tabline = '%!v:lua.lualine_tabline()'
+    vim.o.tabline = "%!v:lua.require'lualine'.tabline()"
     vim.o.showtabline = 2
   end
 end
 
 local function set_statusline()
   if next(config.sections) ~= nil or next(config.inactive_sections) ~= nil then
-    _G.lualine_statusline = status_dispatch
-    vim.o.statusline = '%!v:lua.lualine_statusline()'
+    vim.o.statusline = "%!v:lua.require'lualine'.statusline()"
     vim.api.nvim_exec([[
-    autocmd lualine WinLeave,BufLeave * lua vim.wo.statusline=lualine_statusline()
-    autocmd lualine WinEnter,BufEnter * setlocal statusline=%!v:lua.lualine_statusline()
+    autocmd lualine WinLeave,BufLeave * lua vim.wo.statusline=require'lualine'.statusline()
+    autocmd lualine WinEnter,BufEnter * set statusline<
     ]], false)
   end
 end
@@ -279,4 +277,8 @@ local function setup(user_config)
   set_tabline()
 end
 
-return {setup = setup}
+return {
+  setup = setup,
+  statusline = status_dispatch,
+  tabline = tabline,
+}
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,6 +12,11 @@ please use lua-format before creating a pr :smile:
 * refer to example below to see how themes are defined
 * take 4 screenshots showing a different vim modes (normal, insert, visual, replace)
 * add your theme with screenshots attached to [THEMES.md](./THEMES.md) while maintaining alphabetical order
+* If the colorscheme you're trying to add already support lightline. You can use
+[lightline2lualine theme converter](https://gist.github.com/shadmansaleh/000871c9a608a012721c6acc6d7a19b9) to easily port the theme to lualine.
+
+**Note to colorscheme authors** : If you want to support lualine. You can put your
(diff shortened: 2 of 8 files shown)


hoob3rt:
> I'm wondering shoukd we provide some conditions ourselves like [galaxyline]

I'm aware on how galaxyline does this but I think we should only expose this to developers not users. Otherwise we just become galaxyline with extra steps, I don't want to scare users away with too many config options, the core idea behind this plugin is that it's easy to configure. more options -> more complications imo. But I like this idea overall.

I will take a look at it after we merge #141. There should be no more issues there hopefully but I want to give it a final look at it before merge (this will happen whenever I have free time that's not 3am :smile:)

shadmansaleh:
> I'm aware on how galaxyline does this but I think we should only expose this to developers not users. 

Ok then just the condition option and no template.

>I don't want to scare users away with too many config options, the core idea behind this plugin is that it's easy to configure. more options 

I don't think users can get scared away if they get some sane defaults :)

hoob3rt:
> > I don't want to scare users away with too many config options, the core idea behind this plugin is that it's easy to configure. more options
> 
> I don't think users can get scared away if they get some sane defaults :)

Maybe that's just me but I do 😆 anyway, we shall discuss this at a later date

shadmansaleh:
@hoob3rt I've written a config for [evilline](https://gist.github.com/shadmansaleh/cd526bc166237a5cbd51429cc1f6291b)

Screenshot_20210407-070245_Termux


Show them this if anyone says lualine is less configureble :P

hoob3rt:
> @hoob3rt I've written a config for [evilline](https://gist.github.com/shadmansaleh/cd526bc166237a5cbd51429cc1f6291b)
> 
> Screenshot_20210407-070245_Termux
> 
> 
> Show them this if anyone says lualine is less configureble :P

Hahaha thats sick! I was actually thinking of providing better screenshots for readme and example configs alongside. Drop a config for this one here please :D

shadmansaleh:
I'll clean it up and make a pr :P . Though it depends on this pr so after its merged .

shadmansaleh:
@hoob3rt I've added evilline config as 4th commit to this pr . Just remember not to squash them . Also can you give me a date on when you will be free ?

shadmansaleh:
Ready for review . Though it'll probably be better if  diff is checked commit by commit as changes are isolated in single commit.

lua/lualine/init.lua
```diff
@@ -251,19 +251,17 @@ end
...
 local function set_statusline()
   if next(config.sections) ~= nil or next(config.inactive_sections) ~= nil then
-    _G.lualine_statusline = status_dispatch
-    vim.o.statusline = '%!v:lua.lualine_statusline()'
+    vim.o.statusline = "%!v:lua.require'lualine'.statusline()"
     vim.api.nvim_exec([[
-    autocmd lualine WinLeave,BufLeave * lua vim.wo.statusline=lualine_statusline()
-    autocmd lualine WinEnter,BufEnter * setlocal statusline=%!v:lua.lualine_statusline()
+    autocmd lualine WinLeave,BufLeave * lua vim.wo.statusline=require'lualine'.statusline()
+    autocmd lualine WinEnter,BufEnter * set statusline<
```
hoob3rt, line 264: is the `<` sign something I don't know in lua? I have no idea why that' here. Otherwise lgtm. I have to say, these changes are pretty epic.

I have an idea on how to extend these changes and make it even better but that's out of scope of this pr. I will create an issue with it

hoob3rt: looks like it silences an echo? Probably better to have
```
 autocmd lualine WinEnter,BufEnter * silent set statusline
```

shadmansaleh: It's a vim feature . set statusline< unsets the winlocal option set through setlocal statusline . Checkout `:h setlocal` it's documented in 2nd and 3rd row .

hoob3rt: oh right, I learn something about vim everyday 😆 . It's all good then

PR #162, opened 2021-04-07
Fix: Restore support for dictionary-typed theme setup parameter

freehaha:
This was the original behaviour before 37a3b8c, using the provided `config.options.theme` as a dictionary to build highlight groups unless it's a string.

### DIFF
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -279,6 +279,8 @@ local function setup_theme()
       }, true, {})
       theme = require 'lualine.themes.gruvbox'
     end
+  else
+    theme = config.options.theme
   end
   highlight.create_highlight_groups(theme)
   vim.api.nvim_exec([[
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Oops . This is on me . @freehaha thanks for fixing it . @hoob3rt can you merge this as soon as you can ...

lua/lualine/init.lua
```diff
@@ -279,6 +279,9 @@ local function setup_theme()
       }, true, {})
       theme = require 'lualine.themes.gruvbox'
     end
+  else
+    -- use the provided theme as-is, assuming it's a dictionary
```
hoob3rt, line 283: no need to assume anything, better to have `elseif type == table` here

and below `else` with error log + defaulting to gruvbox

shadmansaleh: Merge this one for now. It will fix the bug on master . I'll make a pr regarding the changes you're looking for.

freehaha: Sorry, was working during the day so wasn't able to update it. I saw there's another PR opened for this so I'll resolve this convo. Thanks for the merge.

hoob3rt: np, thanks for noticing, I've applied additional changes already

PR #178, opened 2021-04-14
Better fix to nil on vimfunc

shadmansaleh:
This can differentiate between string 'nil' and value nil.

Plus cleaner IMO


hoob3rt:
Yep looks better for sure

hoob3rt (approved):

PR #182, opened 2021-04-16
refactor: moved loading to new module

hoob3rt:
- [x] export loading to new module
    - [x] clean up loading functios
- [x] export all config parsing etc to new (or defaults.lua) module
    - [x] also needs refactoring of `check_single_separator` function
- [ ] config tests


lua/lualine/init.lua
```diff
@@ -267,12 +203,14 @@ local function set_statusline()
 end
 
 local function setup(user_config)
-  apply_configuration(vim.g.lualine)
-  apply_configuration(user_config)
-  check_single_separator()
+  if vim.g.lualine then
+    apply_configuration(vim.g.lualine)
+  elseif user_config then
```
shadmansaleh, line 208: I think we should prioritize user_config over global lualine variable.

Makefile
```diff
@@ -2,12 +2,12 @@
...
 lint:
 	@luacheck lua/lualine
-	@luacheck tests
+	@luacheck lua/tests
 
 format:
 	@for file in `find -name '*.lua'`;do lua-format $$file -i; done;
 
 test:
-	@nvim --headless -u tests/minimal_init.lua -c "PlenaryBustedDirectory tests/ { minimal_init = './tests/minimal_init.lua' }"
```
shadmansaleh, line 11: I'm actuallu against putting tests in lua. As it unnesserily adds them to user's `path` . You should already be able to do `require'lualine'` As it's in path . If you want to add things from tests to path you can apply the patch for minimal_init.lua 7fd434f76f2c2d06c03d75ea1ebe8b55eb8eec83 .

lua/lualine/init.lua
```diff
@@ -239,33 +132,34 @@ local function tabline() return statusline(config.tabline, true) end
...
+        local theme = get_theme_from_config()
+        highlight.create_highlight_groups(theme)
+        vim.api.nvim_exec([[
     augroup lualine
     autocmd ColorScheme * lua require'lualine.utils.utils'.reload_highlights()
     augroup END
     ]], false)
-    async_loader:close()
-  end))
+        async_loader:close()
```
shadmansaleh, line 161: The indents for commands look wrong

lua/lualine/utils/loader.lua
```diff
@@ -0,0 +1,62 @@
...
+local function load_extensions(config)
+  for index, extension in pairs(config.extensions) do
+    local local_extension = require('lualine.extensions.' .. extension)
+    load_sections(local_extension.sections, config.options)
+    load_sections(local_extension.inactive_sections, config.options)
+    config.extensions[index] = local_extension
+  end
+end
+
+return {load_components = load_components, load_extensions = load_extensions}
```
shadmansaleh, line 62: Maybe we can just expose one function `loader()` that calls bot component and extention loader . Since we always need to do the both this modul3 might just handle that . What do you think?

lua/tests/spec/config_spec.lua
```diff
@@ -0,0 +1,53 @@
+local luassert = require 'luassert'
+local config_module = require 'lualine.config'
+
+local eq = luassert.are.same
```
shadmansaleh, line 4: Just pull them from the helper

Makefile
```diff
@@ -2,12 +2,12 @@
...
+	@nvim --headless -u lua/tests/minimal_init.lua -c "PlenaryBustedDirectory lua/tests/ { minimal_init = './lua/tests/minimal_init.lua' }"
```
shadmansaleh, line 11: I'm actually against putting `tests` in `lua/` . As it unnessarily puts them on user's path .Ypi shluld already be able to do `require'lualine'` . As it's in path. If you want to require test files too apply the patch for `minimal_init` from 7fd434f76f2c2d06c03d75ea1ebe8b55eb8eec83 . I couldn't find much use of it thet's why I removed it later /

lua/lualine/init.lua
```diff
@@ -239,33 +132,34 @@ local function tabline() return statusline(config.tabline, true) end
...
+        local function get_theme_from_config()
+          local theme_name = config.options.theme
+          if type(theme_name) == 'string' then
+            local ok, theme = pcall(require, 'lualine.themes.' .. theme_name)
+            if ok then return theme end
+          elseif type(theme_name) == 'table' then
+            -- use the provided theme as-is
+            return config.options.theme
+          end
+          vim.api.nvim_echo({
```
shadmansaleh, line 146: I we should use `nvim_err_writeln()` instead of `nvim_echo`.
Or maybe `nvim_notify` . What do you think ?

lua/lualine/init.lua
hoob3rt: my lua-format broke or something? I will apply formatting today

lua/lualine/utils/loader.lua
hoob3rt: yeah I like it

lua/lualine/init.lua
hoob3rt: I will check both of these functions out and let you know what I think

Makefile
shadmansaleh: > This is exactly the problem I'm having however this patch does not help.

`require'tests.helper'` will need to be changed to `require'helper'` .

Also maybe try puting lualine's directory (from pwd) instead of '.'

lua/lualine/init.lua
shadmansaleh: `nvim_err_writeln()` has simpler semantics and `nvim_notify()` can be redirected to where ever user likes . For example mine goes to a float instead of messages .

Makefile
hoob3rt: > `require'tests.helper'` will need to be changed to `require'helper'` .

this does not help unfortunately.

> Also maybe try puting lualine's directory (from pwd) instead of '.'

lualine modules load properly, the issue is loading test files. Does this work properly on your machine with clean branch?

shadmansaleh: >Does this work properly on your machine with clean branch?

Just cloned master on a separate directory and tests ran fine

shadmansaleh: Do you just get error for `build_component_opts` and not for `eq` or `meths` ?

Also It should be `require'helpers'` I missed the s in the end last time.

lua/lualine/init.lua
hoob3rt: I've went with `nvim_err_writeln` since it does exactly what we need it to do with no maintanence needed on our side :smile:

hoob3rt:
I will merge this one since I cannot resolve this test issue mentioned earlier...

PR #184, opened 2021-04-19
spelling: instence -> instance

timbedard:
Quick spelling fix.

PR #185, opened 2021-04-19
feat(filename): add support for custom filename symbols (modified, readonly)

timbedard:
Let's let users customize these symbols as well. Color would be nice too, eventually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,7 @@ Option   | Default | Behaviour
 file_status | true | Displays file status (readonly status, modified status)
 full_path | false | Displays relative path if set to `true`, absolute path if set to `false`
 shorten | true | if `full_path` is true and `shorten` is `false` it shortens absolute path `aaa/bbb/ccc/file` to `a/b/c/file`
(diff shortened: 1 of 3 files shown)


shadmansaleh:
The implementation looks right . But a few things need to change before it can be merged.

1. You need to document the option in README.md and doc/lualine.txt
2. Changes from #184 needs to go . Those have been merged already :P 
3. I don't know why you've changed the single quotes to double quotes . In the repo singles are used . Run lua-format from base of repo . That should fix it . I'm assuming you ran it from some other dir thus config didn't get applied

timbedard:
Ha ha. Too fast for you.

And the quotes were because of lua-fmt, which doesn't have a single quote option. 🙄 Guess I need to actually get lua-format working.

Should be good to go.

shadmansaleh (approved):

shadmansaleh:
> Should be good to go

Indeed in my opinion . But I can't merge it . When @hoob3rt comes along he'll do it :]

hoob3rt (approved):
lgtm, we can merge this one, thank you both :smile:

PR #189, opened 2021-04-22
feat/Add toggle

shadmansaleh:
Currently there's no proper way to disable lualine and restore it later. I think this is something we should allow .

with this pr it can be done easily with 
```lua
require'lualine'.toggle()
```

### DIFF
--- a/doc/lualine.txt
+++ b/doc/lualine.txt
@@ -39,9 +39,10 @@ CONTENTS                                                     *lualine_contents*
         1.3.6. Custom options...........................|lualine_custom_options|
         1.3.7. Using tabline as statusline...............|lualine_using_tabline|
         1.3.8. Loading plugin extensions.....|lualine_loading_plugin_extensions|
-        1.3.9  Config examples.........................|lualine_config_examples|
-            1.3.9.1. Packer.nvim......|lualine_config_example_using_packer.nvim|
-            1.3.9.2 VIML example.......|lualine_full_config_example_inside_viml|
+        1.3.9. Toggling lualine.................................|lualine_toggle|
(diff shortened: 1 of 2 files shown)


shadmansaleh:
@hoob3rt do you agree with this change ?

hoob3rt:
I don't see why we need this. What's the use case for toggling your statusline?

shadmansaleh:
It's so other things can take control of statusline like [TrueZen](https://github.com/kdav5758/TrueZen.nvim) for example . Also because of BufEnter autocommands it's hard to get rid of lualine and restoring is imposible .

hoob3rt:
does it not work with truezen? Then it's an issue on their side. [goyo](https://github.com/junegunn/goyo.vim) handles it correctly and lualine works just fine

hoob3rt:
even if you do `set laststatus=1` and then `set laststatus=2` it restores lualine correctly

shadmansaleh:
>even if you do set laststatus=1 and then set laststatues=2 it restores lualine correctly

Ok I totaly forgot about that option :P

hoob3rt:
> > even if you do set laststatus=1 and then set laststatues=2 it restores lualine correctly
> 
> Ok I totaly forgot about that option in vim :P

np, happens to me all the time as well :smile:

shadmansaleh:
It only works on 1 window

shadmansaleh:
> 	The value of this option influences when the last window will have a
	status line:
		0: never
		1: only if there are at least two windows
		2: always
	The screen looks nicer with a status line if you have several
	windows, but it takes another screen line. |status-line|

@hoob3rt It seems to only deal with last window . Others are unaffected by this.

PR #194, opened 2021-04-25
Fix themes incorrectly setting the bold attribute

agarof:
Some themes set the `bold` attribute incorrectly.
This is probably due to an error in [lightline2lualine theme converter](https://gist.github.com/shadmansaleh/000871c9a608a012721c6acc6d7a19b9).


hoob3rt:
good catch, thanks :smile:

PR #197, opened 2021-04-28
use project relative path

NullVoxPopuli:
Path now starts from your CWD where you opened neovim :D 

Example: 

image

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -23,7 +23,8 @@ FileName.new = function(self, options, child)
 end
 
 FileName.update_status = function(self)
-  local data = vim.fn.expand('%:p')
+  local data = vim.fn.expand('%')
+
   if not self.options.full_path then
     data = vim.fn.expand('%:t')
   elseif self.options.shorten then


shadmansaleh:
Ok . You've broken displaying full path . What you're doing here can already be done by setting full_path=true shorten=true like I've said many times before in your issue. I see no point of this pr :P

NullVoxPopuli:
> . What you're doing here can already be done by setting full_path=true shorten=true

nope -- I've provided screenshots about this -- I think we maybe misunderstand each other?. full_path includes a my entire path, which I don't want. I only want from my CWD to the file. That impacts the length significantly which then makes _all paths_ shorten, cause the _base path_ is longer than 40 characters... 

switching to a relative path means that we have more characters to work with -- and the display path is more meaningful.

shorten=true will _only_ show the file name if the path is collapsed, and then one character per directory -- which isn't useful to me

shadmansaleh:
> switching to a relative path means that we have more characters to work with -- and the display path is more meaningful.

This is what shorten does .

> shorten=true will only show the file name if the path is collapsed, and then one character per directory -- which isn't useful to me

The docs are wrong . Did you try the config with full_path=true shorten=true ??

NullVoxPopuli:
> . Did you try the config with full_path=true shorten=true ??

Here is what these options do for me:
image
^ this is the CWD

image
^ happily surprised to see this

image
^ I went up two directories, and this is back to being not helpful

so, I apologize for being hasty, but _I could have sworn_ I was getting super full path in here like `/home/user/etc/etc/etc/finallyCWD/`

NullVoxPopuli:
so maybe I'll close this, and think about how to implement gradual collapsing of the left-most directory until under max_chars, or whatever we end up calling it

shadmansaleh:
>  think about how to implement gradual collapsing of the left-most directory until under max_chars,

Now that's a good idea 👍

PR #198, opened 2021-04-28
Add max_chars to filename

NullVoxPopuli:
requires: https://github.com/hoob3rt/lualine.nvim/pull/197
(only because I'm using both features locally)

Preview (same config as the other PR): 
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,7 @@ Option   | Default | Behaviour
 :------: | :------: | :----:
 file_status | true | Displays file status (readonly status, modified status)
 full_path | false | Displays relative path if set to `true`, absolute path if set to `false`
+max_chars | 40 | limits how many characters the path may use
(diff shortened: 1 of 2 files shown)


hoob3rt:
There is no need for this to have #197 changes inside. Remove 05fda1f from this pr and we can merge this.

NullVoxPopuli:
@hoob3rt commit dropped

shadmansaleh:
I agree with thia change but max_char is probablya  bad name . It's not quite understandable what it does seeing that name . Also the deascription in docs is miss leading since pathshorten doesn't garantee length will be under max_char. Plus you didn't update the docs in `doc/lualine.txt`

NullVoxPopuli:
maybe `shorten_at_length`?

shadmansaleh:
> maybe shorten_at_length

Maybe . Better then max_chars imo .I'm not sure about the name . Maybe you can take input from @hoob3rt :)

NullVoxPopuli:
less configuration! https://github.com/hoob3rt/lualine.nvim/pull/201

PR #201, opened 2021-04-29
Dynamic shortening

NullVoxPopuli:
This may eliminate the need for both `shorten` and `max_chars` (#198)

Some examples with the same config, but different buffer widths:
image
image
image
image
image
image
image
image


lua/lualine/components/filename.lua
```diff
@@ -24,6 +28,9 @@ end
 
 FileName.update_status = function(self)
   local data = vim.fn.expand('%:p')
+  local winW = vim.fn.winwidth(0)
+  local estimatedSpaceAvailable = winW - 40
```
shadmansaleh, line 32: We use snake case in the repo . So please follow the style.

Also shouldn't it be just 40 ? doesn't winW - 40 mean it'll leave 40 chars for all the other component and itself shorten when it can't prov9de that ?

lua/lualine/components/filename.lua
```diff
@@ -32,8 +39,12 @@ FileName.update_status = function(self)
 
   if data == '' then
     data = '[No Name]'
-  elseif vim.fn.winwidth(0) <= 84 or #data > 40 then
-    data = vim.fn.pathshorten(data)
+  end
+
+  for i = 0, count(data, "/"), 1 do
```
shadmansaleh, line 44: You're hard coding path separator but remember windos uses forward slash instead of back slash .You can use `package.config:sub(1,1)` as path separator if I remember right . See the branch component for example .

lua/lualine/components/filename.lua
```diff
@@ -32,8 +39,12 @@ FileName.update_status = function(self)
...
+    if winW <= 84 or #data > estimatedSpaceAvailable then
+      data = data:gsub("([^/])[^/]+%/", "%1/", 1)
```
shadmansaleh, line 46: Same as above path separator hard coded .

shadmansaleh:
> This may eliminate the need for both shorten and max_chars (#198)

shorten allows you to have relative path instead of full paths . So it won't be obsolete by this pr . But docs needs to be fixed to say what it actually does . Though it's out of scope if this pr .

lua/lualine/components/filename.lua
NullVoxPopuli: I thought this was the window width, so the estimated space would be dynamically 40 less than the width of the window

shadmansaleh: yes it's window width . What I was saying was previously it would've shorted when the filename > 40 in all windows sizes . Now it's when filename > 80 for winW 120 filename > 20 for winW 60 . It's not nesserily a bad thing . I was just asking  if you were trying to replicate previous behavior :)

hoob3rt:
> > This may eliminate the need for both shorten and max_chars (#198)
> 
> shorten allows you to have relative path instead of full paths .

`shorten` actually just shortens the path, `full_path` is for relative/full.

As for this pr, I really like this, I had something like this in mind for a long time but didn't have time to implement it. I think this should be default behaviour for `full_path`.

Regarding `full_path` and `shorten`... They are both obsolete imo. We should remove them both and have one flag called `relative` (defaulting to `false`). Do this after @shadmansaleh suggestions and we will be able to integrate this. This will also remove the confusing naming of both of these variables. Even I and @shadmansaleh seem to confuse them constantly.

shadmansaleh:
> shorten actually just shortens the path, full_path is for relative/full.

Seriously man I know what it does I've written it . You can test it out . full_path enables paths in file name while shorten turns that path in relative path when enabled . The path shorting we have that is default behavior when screen width < 82 or filename > 40 . Also if you remember shorten was first [proposed](https://github.com/hoob3rt/lualine.nvim/pull/60#discussion_r573962895) to be named relative .

> Regarding full_path and shorten... They are both obsolete imo. We should remove them both and have one flag called relative (defaulting to false). 

If we're only keeping one option then it should probably be an integer to account for all 3 states ( maybe named path)
where :
- path = 0 means no path and just filename
- path = 1 means relative path
- path = 2 means full path

Default being 0 as that's current default behavior
And shorting being applied by default .

> This will also remove the confusing naming of both of these variables. Even I and @shadmansaleh seem to confuse them constantly.

That is indeed true :D

hoob3rt:
> If we're only keeping one option then it should probably be an integer to account for all 3 states ( maybe named path)
> where :
> 
> * path = 0 means no path and just filename
> * path = 1 means relative path
> * path = 2 means full path
> 
> Default being 0 as that's current default behavior
> And shorting being applied by default .

That's better, let's go this route

hoob3rt:
@NullVoxPopuli are you up for wrapping this up? I really want these changes implemented so lmk if I should wait on you or continue from here myself

PR #205, opened 2021-05-02
Performence improvments

shadmansaleh:
This reduces startup time.

### Changes :
- Only load default config when needed
- Only load cterm colors when needed
- Async load components and theme

Without these changes according to packers [profiler](https://github.com/wbthomason/packer.nvim#profiling) lualine takes around 73ms in my device with these changes around 12ms . Thats **6x** improvement.

norcallies [profiler](https://github.com/norcalli/profiler.nvim) shows similar improvements.

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -1,7 +1,7 @@
 -- Copyright (c) 2020-2021 hoob3rt
 -- MIT license, see LICENSE for more details.
 local M = {}
-local utils_colors
+local cterm_colors
 local utils = require 'lualine.utils.utils'
 local section_highlight_map = {x = 'c', y = 'b', z = 'a'}
 local active_theme = nil
@@ -10,16 +10,16 @@ function M.highlight(name, foreground, background, gui, reload)
   local command = {'highlight', name}
   if foreground and foreground ~= 'none' then
     table.insert(command, 'guifg=' .. foreground)
-    if utils_colors then
+    if cterm_colors then
       table.insert(command,
-                   'ctermfg=' .. utils_colors.get_cterm_color(foreground))
+                   'ctermfg=' .. cterm_colors.get_cterm_color(foreground))
     end
   end
   if background and background ~= 'none' then
     table.insert(command, 'guibg=' .. background)
-    if utils_colors then
+    if cterm_colors then
       table.insert(command,
-                   'ctermbg=' .. utils_colors.get_cterm_color(background))
+                   'ctermbg=' .. cterm_colors.get_cterm_color(background))
     end
   end
   if gui then
@@ -36,7 +36,7 @@ function M.create_highlight_groups(theme)
   utils.clear_highlights()
   active_theme = theme
   if not vim.o.termguicolors then
-    utils_colors = require 'lualine.utils.cterm_colors'
+    cterm_colors = require 'lualine.utils.cterm_colors'
   end
   for mode, sections in pairs(theme) do
     for section, colorscheme in pairs(sections) do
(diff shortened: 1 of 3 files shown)


hoob3rt:
I like these changes. There is one thing I'm not sure of as far as code quality goes. Why do you define `async loader` this way?
```lua
  local async_loader
  async_loader = vim.loop.new_async(vim.schedule_wrap(function()
```

I get that you are sending `async_loader:close()` inside this function but is there no better way to do this? Won't a stop signal be enough? Also where did you find docs for the neovim api, I cannot find it myself... 😞

shadmansaleh:
> I get that you are sending async_loader:close() inside this function but is there no better way to do this?

I don't really know.

> Won't a stop signal be enough? 

What's that ?

> Also where did you find docs for the neovim api, I cannot find it myself... 😞

By api you mean `vim.api` stuff ?

lua/lualine/highlight.lua
```diff
@@ -1,24 +1,23 @@
 -- Copyright (c) 2020-2021 hoob3rt
 -- MIT license, see LICENSE for more details.
 local M = {}
-local utils_colors = require 'lualine.utils.cterm_colors'
+local utils_colors
```
hoob3rt, line 4: I would name this `cterm_colors` since that's all it is really

hoob3rt (commented):
> I basically followed example in docs

Okay :smile:

> The luv docs are [here](https://github.com/luvit/luv/blob/master/docs.md)
> uv docs [here](http://docs.libuv.org/en/v1.x/)

Okay, that's what I found as well, for some reason I coudn't find `new_async` in this file an hour ago...

lua/lualine/highlight.lua
shadmansaleh: I agree . Renamed

hoob3rt:
lgtm, this is really nice. 2.5ms -> 0.12ms with my config, thank you sir

nifoc:
I think this breaks setting custom colors for (at least) the diagnostics component.

```
...
color_error = lualine_utils.extract_highlight_colors('LspDiagnosticsSignError', 'fg'),
color_warn = lualine_utils.extract_highlight_colors('LspDiagnosticsSignWarning', 'fg'),
color_info = lualine_utils.extract_highlight_colors('LspDiagnosticsSignInformation', 'fg'),
...
```

Work as expected with 64ab49f55ada99db1e613fc1b9b484921c37c21e, but doesn't use the extracted colors with the current master.

hoob3rt:
> I think this breaks setting custom colors for (at least) the diagnostics component.
> 
> ```
> ...
> color_error = lualine_utils.extract_highlight_colors('LspDiagnosticsSignError', 'fg'),
> color_warn = lualine_utils.extract_highlight_colors('LspDiagnosticsSignWarning', 'fg'),
> color_info = lualine_utils.extract_highlight_colors('LspDiagnosticsSignInformation', 'fg'),
> ...
> ```
> 
> (I use the `extract_highlight_colors` function in my configuration)
> 
> Work as expected with [64ab49f](https://github.com/hoob3rt/lualine.nvim/commit/64ab49f55ada99db1e613fc1b9b484921c37c21e), but doesn't use the extracted colors with the current master.

setting custom colors works fine in my config, even with your snippet. If this issue persists for you then please create a new issue

shadmansaleh:
@nifoc what colorscheme are you useing . After aton of digfing I found out the colorcheme was [setting](https://github.com/folke/tokyonight.nvim/blob/4d7251fe0ade96516a6401f58075df006cbc213e/lua/tokyonight/theme.lua#L337) those highlight groups :P . Lualine works normally :)

hoob3rt:
I had to revert this one as it causes a prety major issue. It basically replicates every component's state from an active window to all other windows, 

<details><summary>take a look at this screenshot</summary>

Screenshot 2021-05-06 at 04 50 48

</details>

This may be resolved with #210, although we would have to set this newly introduces parameter to all components

shadmansaleh:
Take a look at #213

Issue #208, opened 2021-05-03
Bug: Lualine does not load after recent updates


rmagatti:
### Self Checks

- [x] I'm using the latest lualine and neovim nightly.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem
Loading the plugin with these configs on the latest commit does not work, as in lualine just doesn't load.
```lua
require('lualine').setup{
  theme = 'gruvbox',
  extensions = { 'fzf', 'fugitive' },
}
```

Everything works fine however if I rollback to this commit https://github.com/hoob3rt/lualine.nvim/commit/6ba2b80b594c3ead11ab9bd1dbc94c0b4ea46c33

### Actual behaviour
Lualine doesn't load.

### Expected behaviour
Lualine loads :)

### Minimal config to reproduce the issue
I don't have one but let me know if you can't reproduce and I can try and provide one.

<details>
<summary>Config</summary>

</details>

shadmansaleh:
Lualine loads but you don't have any components in any of the sections so it's not displaying anything . It's a intended behavior. Previously the components defined in user config would extend the default . That causes conflicts and extra boilarplate when you want to omit some  sections. So the behavior was changed in f08e8514ddbe73cb02644cbd7b4b98e3681a7f9d .Now default sections don't get applied if you provide configuration (by passing a table to setup function or by g:lualine) . In short `require'lualine'.setup()` would load the default sections but `require'lualine'.setup({})` won't . If you want the default components just copy paste the [default config](https://github.com/hoob3rt/lualine.nvim/blob/master/lua/lualine/defaults.lua) . If you think your issue is resolved then close the issue :)

goolord:
i suppose if you want the defaults to be managed for you you could write
```lua
sections = require'lualine.defaults'.sections,
inactive_sections = require'lualine.defaults'.inactive_sections
```
for as long as that works :d

sdemura:
Might I suggest that `README.md` be updated to reflect this change?

michzappa:
I have also been trying to figure this out for a bit. This is what I have following @shadmansaleh's advice of copying in the default config.
```lua
  use { 'hoob3rt/lualine.nvim',
        config = function()
          require('lualine').setup {
            options = {
              theme = 'auto',
              icons_enabled = false,
              section_separators = '',
              component_separators = '',
              disabled_filetypes = {}
            },
            sections = {
              lualine_a = {'mode'},
              lualine_b = {'branch'},
[...]

PR #210, opened 2021-05-04
Add component evaluation for inactive sections

### LINKED ISSUE #157: Bug: File modification indicator ([+]) not updated for inactive windows
Thanks for the great project 👍

Btw, it looks that the indicator for file modification, [+], doesn't get updated for inactive windows.
The following gifs show some demonstrations.

filemodified

diffedit

shadmansaleh:
Now filename component updates on inactive_status .

closes #157

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,5 @@
 # Contributing to lualine.nvim
 
-
-### General
-
-please use lua-format before creating a pr :smile:
-
-### All contributions are very welcome but themes require a lot of work on my part if not done properly so here's a guide on how to do them.
-
 ### Adding a theme
 
 * refer to example below to see how themes are defined
@@ -30,10 +23,8 @@ But if unspecified then they default to c, b, a sections theme respectively .
 
 Adding theme is really easy in lua. Here is and example of a gruvbox theme.
 ```lua
-local gruvbox = {  }
-
 local colors = {
-  black        = "#282828",
+  black        = '#282828',
   white        = '#ebdbb2',
   red          = '#fb4934',
   green        = '#b8bb26',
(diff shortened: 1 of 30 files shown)


hoob3rt:
this reproduces #208 at least for my machine.

Also I think we should discuss which components should update in inactive state. I would go for
* filetype (already done here)
* diagnostics 
* progress
* location (not sure about this one)

My reasoning behind this is all of above are cheap to run performance wise and could change while doing a sed-like edit as mentioned in #157.

shadmansaleh:
> this reproduces #208 at least for my machine.

Of cource I branched off before the patch . A rebase on master will fix it on this branch.

There's an issue with this `%{}` evaluation escapes `%` in results and I've no idea how to tackle that . It won't be a problem for components that don't emmit % codes but the ones that do causes issues . I'm not sure how progress or location may change in an inactive buffer . They seem like a prime candidate for static component .

hoob3rt:
> I'm not sure how `progress` or `location` may change in an inactive buffer . They seem like prime candidates for static component .

Yeah, my bad. I double checked and my idea was wrong, nevermind that. 



> There's an issue with `%{}` evaluation escapes `%` in results and I've no idea how to tackle that . It won't be a problem for components that don't emmit % codes but the ones that do causes issues .

Could you elaborate? I don't understand the issue atm.

shadmansaleh:
> Could you elaborate? I don't understand the issue atm.

Means %3P from progress displays as %3P as (neo)vim either doesn't evaluate the result or escapes the % in %3P

hoob3rt:
> > Could you elaborate? I don't understand the issue atm.
> 
> Means %3P from progress displays as %3P as (neo)vim either doesn't evaluate the result or escapes the % in %3P

We can easily move to our custom statusline functions instead of using the builtins, that should resolve our problem right?

`location` can be handled easily by `vim.api.nvim_win_get_cursor`, although location would require us to reimplement `%3P` logic ourselves from scrach unless there is another `vim.api` function to handle it

shadmansaleh:
> We can easily move to our custom statusline functions instead of using the builtins, that should resolve our problem right?

Those components aren't the real issue the highlights are . They are applied like `%#hl_name#` . The themes and color option work as they get applied outside of `%{}` evaluation . But components that have their own highlight group and handle them internaly doesn't (diagnostics for example).  Maybe we can provide diagnostics without highlights just with symbols for inactive_status.

shadmansaleh:
@hoob3rt any idea why checks aren't running here ?

hoob3rt:
> Maybe we can provide diagnostics without highlights just with symbols for inactive_status.

I don't like this tbh.

Honestly I don't know what to do with this issue... We should find a working solution for this otherwise it could cause problems down the line

shadmansaleh:
> I don't like this tbh.

I don't like it much either. But I've no idea how to solve it :D

PR #212, opened 2021-05-06
Revert "Performence improvments"

hoob3rt:
Reverts hoob3rt/lualine.nvim#205

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -1,10 +1,11 @@
 -- Copyright (c) 2020-2021 hoob3rt
 -- MIT license, see LICENSE for more details.
 local M = {}
-local cterm_colors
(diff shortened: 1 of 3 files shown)


shadmansaleh:
How come this still has double checks :D . I think we need to ignore every other branch on but the master for OnPush event . Is there a way to do this ?

hoob3rt:
> How come this still has double checks :D . I think we need to ignore every other branch on but the master for OnPush event . Is there a way to do this ?

I think we need onpush for other branches, otherwise updates for branches won't work(I think). I will take a look into it

hoob3rt:
I think it should be enough to change
```
on: 
  push:
    branches-ignore: [master]
  pull_request:
    types: [opened, reopened]
```
to
```
on: 
  push:
    branches-ignore: [master]
  pull_request:
    types: [reopened]
```

How do we test this though I'm not sure...

PR #216, opened 2021-05-07
Use LspDiagnosticsDefault colors when available before falling back to DiffDelete, DiffText and Normal

folke:


hoob3rt:
lgtm, thanks :smile:

PR #222, opened 2021-05-10
Proof of concept

JonathanGin52:


lua/lualine/components/filetype.lua
```diff
@@ -8,7 +10,12 @@ FileType.update_status = function(self)
     local ok, devicons = pcall(require, 'nvim-web-devicons')
     if ok then
       local f_name, f_extension = vim.fn.expand('%:t'), vim.fn.expand('%:e')
-      self.options.icon = devicons.get_icon(f_name, f_extension)
+      icon, icon_highlight_group = devicons.get_icon(f_name, f_extension)
+      highlight_color = vim.fn.synIDattr(vim.fn.hlID(icon_highlight_group), 'fg')
+      self.options.icon_highlight = highlight.create_component_highlight_group(
```
shadmansaleh, line 15: You can add a check with utils.utils.highlight_exists() that way you won't need to create new highlight on every load. maybe create hl_name like 'lualine_'..icon_highlight_group check if it exists if it doesn't create it . That way  all filetype colors won't have to go through same name .

lua/lualine/component.lua
```diff
@@ -90,8 +90,11 @@ local Component = {
   end,
 
   -- Apply icon in front of component
-  apply_icon = function(self)
+  apply_icon = function(self, default_highlight)
```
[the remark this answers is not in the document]
shadmansaleh:
```lua
is_focused = vim.g.statusline_winid == vim.fn.win_getid() 
```

lua/lualine/components/filetype.lua
```diff
@@ -8,7 +11,24 @@ FileType.update_status = function(self)
...
+      self.options.icon = icon
+
+      if self.options.icon and (self.options.colored or self.options.colored == nil) then
+        self.options.colored = true
+
+        local highlight_color = vim.fn.synIDattr(vim.fn.hlID(icon_highlight_group), 'fg')
+        local is_focused = vim.g.statusline_winid == vim.fn.win_getid() 
+        local default_highlight = highlight.format_highlight(is_focused, self.options.self.section)
+        local icon_highlight = self.options.self.section .. '_' .. icon_highlight_group
+        if not utils.highlight_exists(icon_highlight .. '_normal') then
```
JonathanGin52, line 24: I made the assumption that checking if the highlight exists in normal mode is enough. Is this a safe assumption? @shadmansaleh

lua/lualine/components/filetype.lua
```diff
@@ -8,7 +11,25 @@ FileType.update_status = function(self)
...
     if ok then
       local f_name, f_extension = vim.fn.expand('%:t'), vim.fn.expand('%:e')
-      self.options.icon = devicons.get_icon(f_name, f_extension)
+      local icon, icon_highlight_group = devicons.get_icon(f_name, f_extension)
+      self.options.icon = icon
+
+      if self.options.icon and (self.options.colored or self.options.colored == nil) then
+        self.options.colored = true
+
+        local highlight_color = vim.fn.synIDattr(vim.fn.hlID(icon_highlight_group), 'fg')
```
hoob3rt, line 20: our `utils.extract_highlight_colors` function does the same thing. @shadmansaleh I think we can do a reformat of that function (in another pr obviously), this looks cleaner imo

shadmansaleh, line 20: You can call `utils.extract_highlight_colors()` here instead of `synIDattr(hlID())`

hoob3rt: same idea hehe

hoob3rt:
That was really fast, I appreciate it :smile: There are a few things missing before we can merge it:
* apply lua-format
* update lualine.txt docfile
* get tests for this functionality ( lmk if you can wrap your head around it, you will need to add icon plugin to path to get it working )

lua/lualine/components/filetype.lua
shadmansaleh: Well I knew about this solution but I thought of using api instead of vim fn calls .

lua/lualine/components/filetype.lua
hoob3rt: > Assuming normal will work . But I think you're giving wrong arg to highlight exist as we prepend all highlight groups with 'lualine_'. Your query should have that too 

it already does, `icon_highlight` contains `self.options.self.section`

> Now I'm thinking since you can be sure once you call create_component_highlight the highlight will exist . A simpler solution might be to just keep track of what highlight groups you've created in self .

I don't think it would be simpler. This is one function call and we already track the created highlights there. Am I missing something here?

shadmansaleh: > it already does, icon_highlight contains self.options.self.section

My bad I thought we were placeing only section names in self.section . Thanks for correcting me .

> I don't think it would be simpler. This is one function call and we already track the created highlights there. Am I missing something here?

ok this seems fine :)

lua/tests/spec/component_spec.lua
```diff
@@ -241,6 +242,57 @@ describe('Fileformat component', function()
...
+    stub(hl, 'create_component_highlight_group')
+    hl.create_component_highlight_group.returns('MyCompHl')
+
+    local opts = build_component_opts({
+      component_separators = {'', ''},
+      padding = 0
+    })
+    -- assert.stub(hl.create_component_highlight_group).was_called_with(
+    --   {fg = '#000'}, 'test_highlight_group', {}
+    -- )
```
JonathanGin52, line 291: Can't seem to get this assertion to work. Left what I tried as comments in case anyone has any ideas. Will remove if unsuccessful.

JonathanGin52:
This is my first time contributing / testing a neovim plugin. I had a few naïve questions that I would appreciate your guys' expertise in 😃 

- Is there a way to have print statements print to stdout? All my print statements seem to get swallowed into the void when testing in headless mode.
- If not, is it possible to log to a file instead?
- Do you have any resources to get started with setting up nvim for lua development?

lua/tests/spec/component_spec.lua
```diff
@@ -241,6 +242,57 @@ describe('Fileformat component', function()
...
+    })
+    assert_component('filetype', opts, 'lua')
+  end)
+
+  it('colors nvim-web-devicons icons', function()
+    package.loaded['nvim-web-devicons'] = {
+      get_icon = function(_, _)
+        -- return icon, icon_highlight_group
+        return '*', 'test_highlight_group'
+      end
```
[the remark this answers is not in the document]
shadmansaleh: This will work . You could always use stub . I'm sure it does somethung similar behind the scens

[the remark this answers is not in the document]
JonathanGin52: I'm not sure how stub would work with `require`, so I went with this approach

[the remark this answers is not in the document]
shadmansaleh: This is ok . You can keep it this way.

JonathanGin52:
> @JonathanGin52 patch for the called_with assert you were trying to do
> 
> ```diff
> diff --git a/lua/lualine/components/filetype.lua b/lua/lualine/components/filetype.lua
> index 73aa1c2..df4eb91 100644
> --- a/lua/lualine/components/filetype.lua
> +++ b/lua/lualine/components/filetype.lua
> @@ -18,8 +18,8 @@ FileType.update_status = function(self)
>            (self.options.colored or self.options.colored == nil) then
>          self.options.colored = true
>  
> -        local highlight_color = vim.fn.synIDattr(
> -                                    vim.fn.hlID(icon_highlight_group), 'fg')
> +        local highlight_color = utils.extract_highlight_colors(
> +                                    icon_highlight_group, 'fg')
>          local is_focused = vim.g.statusline_winid == vim.fn.win_getid()
>          local default_highlight = highlight.format_highlight(is_focused,
>                                                               self.options.self
> diff --git a/lua/tests/spec/component_spec.lua b/lua/tests/spec/component_spec.lua
> index 228cd21..f71435d 100644
> --- a/lua/tests/spec/component_spec.lua
> +++ b/lua/tests/spec/component_spec.lua
> @@ -279,17 +279,26 @@ describe('Filetype component', function()
>      -- mock_hl.create_component_highlight_group.returns('MyCompHl')
>  
>      local hl = require 'lualine.highlight'
> +    local utils = require 'lualine.utils.utils'
>      stub(hl, 'create_component_highlight_group')
> +    stub(utils, 'extract_highlight_colors')
>      hl.create_component_highlight_group.returns('MyCompHl')
> +    utils.extract_highlight_colors.returns('#000')
>  
>      local opts = build_component_opts({
>        component_separators = {'', ''},
>        padding = 0
>      })
> -    -- assert.stub(hl.create_component_highlight_group).was_called_with(
> -    --   {fg = '#000'}, 'test_highlight_group', {}
> -    -- )
>      assert_component('filetype', opts, '%#MyCompHl_normal#*%#lualine_c_normal# lua')
> +    assert.stub(utils.extract_highlight_colors).was_called_with(
> +      'test_highlight_group', 'fg'
> +    )
> +    opts.icon = '*'
> +    assert.stub(hl.create_component_highlight_group).was_called_with(
> +      {fg = '#000'}, 'test_highlight_group', opts
> +    )
> +    hl.create_component_highlight_group:revert()
> +    utils.extract_highlight_colors:revert()
>    end)
>  end)
> ```

This is great, thanks!

shadmansaleh (approved):

Shatur95:
Is this option works now?
I even tried to specify it explicitly, but have icons without colors:

изображение


[Here](https://github.com/Shatur95/neovim-config/blob/fc60ab07b0dbaf175ca8261af014bc089f562fea/lua/config/plugins/lualine.lua#L85) is my configuration.

It have colors in other plugins that use nvim-web-devicons.

PR #224, opened 2021-05-10
Dynamic shortening 2

hoob3rt:
closes #201 

@shadmansaleh I think I fixed what was wrong with #201, take a look. (needs docs obviously but that's for later)

I also added a shortening method for `mode` component ( I liked it when I was using airline ). I think that having another optional shorten method for the `component` class can make this easier to read and extend from, what do you think? If we go that route then I think we should also discuss possible shortening methods for other components other than `filename` and `mode` :smile:


shadmansaleh:
TBH I don't like mode shorting that being forced is a big no for me . You could always do it in config with format option . Why is does it need to be in here ?

```lua
lualine_a = {
  {'mode', format = function(data)
     local windwidth = vim.fn.winwidth(0)
      if windwidth <= 84 then return data:sub(1,1) end
      return data
  }
}
```

hoob3rt:
> TBH I don't like mode shorting that being forced is a big no for me . You could always do it in config with format option . Why is does it need to be in here ?

Alright, I will create an issue for discussion about this. Let's focus on `filename` here

lua/lualine/components/filename.lua
```diff
@@ -2,38 +2,50 @@
...
-  if data == '' then
-    data = '[No Name]'
-  elseif vim.fn.winwidth(0) <= 84 or #data > 40 then
-    data = vim.fn.pathshorten(data)
+  local windwidth = vim.fn.winwidth(0)
+  local estimated_space_available = 40
+  local path_separator = package.config:sub(1, 1)
+  for _ = 0, count(data, path_separator), 1 do
+    if windwidth <= 84 or #data > estimated_space_available then
+      data = data:gsub('([^/])[^/]+%/', '%1/', 1)
```
shadmansaleh, line 47: You're still using hardcoded '/' as separator

lua/lualine/init.lua
```diff
@@ -260,8 +261,8 @@ local function set_statusline()
   if next(config.sections) ~= nil or next(config.inactive_sections) ~= nil then
     vim.o.statusline = '%!v:lua.require\'lualine\'.statusline()'
     vim.api.nvim_exec([[
-    autocmd lualine WinLeave,BufLeave * lua vim.wo.statusline=require'lualine'.statusline()
-    autocmd lualine WinEnter,BufEnter * set statusline<
+    autocmd lualine VimResized, WinLeave,BufLeave * lua vim.wo.statusline=require'lualine'.statusline()
```
shadmansaleh, line 264: I'm actually not sure what this change does . Will having a separate auto command for VimResized executing `:redrawsratus` have same effect?

hoob3rt: > I'm actually not sure what this change does . 
Well obviously it redraws the statusline on `VimResized` 😆 I think it's useful, at least for me. If you maximize neovim window you probably want the shortened path to be displayed as big as possible again

>Will having a separate auto command for VimResized executing `:redrawstatus` have same effect?
yes, that is better imo.

lua/lualine/components/filename.lua
```diff
@@ -6,6 +6,13 @@ local function count(base, pattern)
   return select(2, string.gsub(base, pattern, ''))
 end
 
+local function shorten_path(path, path_separator)
+  -- ('([^/])[^/]+%/', '%1/', 1)
+  return path:gsub(
+             '([^' .. path_separator .. '])[^' .. path_separator .. ']+%' ..
```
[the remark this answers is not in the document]
hoob3rt: I will need a help with this one, I really suck at regex 😥

NullVoxPopuli:
thanks for the assist! I got busy. 

🎉

NullVoxPopuli:
🤔 something is goofy

with path = 0, I get just the file name, as documented
but with both path 1 & 2, there is far-too-aggressive path collapsing: 

here is `path = 1`
image
the path is: 
`home/.config/nvim/lua/plugin-config/statusline.lua` -- only 50 characters, and takes about 1/3rd of my available space.

here is `path = 2`
image
same as above, but includes the absolute path from `/` on disk, which is `/home/my-username/Development/NullVoxPopuli/dotfiles/`, which totals to 103 characters, and because this path=2 test was for my inactive buffer, it takes up _less_ total space than the active.

NullVoxPopuli:
figured out the missing piece. PR incoming.

ShiChenCong:
image
i do not know，why show short filename，width is enough

shadmansaleh:
@ShiChenCong  Use [shadmansaleh/lualine.nvim](https://github.com/shadmansaleh/lualine.nvim)

[hoob3rt/lualine.nvim](https://github.com/hoob3rt/lualine.nvim) is currently not maintained . See https://github.com/hoob3rt/lualine.nvim/pull/311

I think the issue you are facing was fixed there.

PR #226, opened 2021-05-11
Remove bold statements from jellybeans

WhyNotHugo:
The lightline config does not include these as bold, so this change makes the theme match what lightline uses.

### DIFF
--- a/lua/lualine/themes/jellybeans.lua
+++ b/lua/lualine/themes/jellybeans.lua
@@ -18,25 +18,25 @@ local colors = {
 
 local jellybeans = {
   visual = {
-    a = {fg = colors.color2, bg = colors.color3, gui = 'bold'},
+    a = {fg = colors.color2, bg = colors.color3},
     b = {fg = colors.color4, bg = colors.color5}
   },
   replace = {
-    a = {fg = colors.color2, bg = colors.color8, gui = 'bold'},
+    a = {fg = colors.color2, bg = colors.color8},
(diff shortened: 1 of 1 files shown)


hoob3rt:
I don't think we state anywhere that lualine should behave exactly like lightline, I will not accept this change, every theme has bold in this section for consitency. If you want to change a theme for local use here is how to do it:
```lua
local my_jellybeans = require'lualine.themes.jellybeans'
my_jellybeans.normal.a.gui = nil
require'lualine'.setup { options = { theme = my_jellybeans } }
```

WhyNotHugo:
Does it really make sense for themes to be consistent with one another?

I think the logical thing is for the `jellybeans` theme to be consistent with other `jellybeans` themes.

I think if there's no intent in maintaining these themes consistent with existing counterparts, it might be best to give them a different name for clarity's sake.

hoob3rt:
> I think if there's no intent in maintaining these themes consistent with existing counterparts, it might be best to give them a different name for clarity's sake.

There is no way to please everyone unfortunately. lightline doesn't use bold text for this and [airline does](https://github.com/vim-airline/vim-airline/wiki/Screenshots). If I approve your change then there might be someone who used airline and want it bold and create a pr for this claiming their theme is more important than yours and that I should maintain themes according to airline and not lightline. I don't want this cycle so I have to stand my ground with how lualine's themes look by default. Hopefully this clears things up and you will manage to set up font weight to your liking with above example :smile:

PR #229, opened 2021-05-11
feat: quickfix extension

hoob3rt:
changes statusline behaviour on quickfix list
Screenshot 2021-05-11 at 19 30 53

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,11 +387,12 @@ extensions = { 'fzf' }
 <details>
 <summary><b>Available extensions</b></summary>
 
+* chadtree
 * fugitive
 * fzf
 * nerdtree
-* chadtree
 * nvim-tree
+* quickfix
 
 </details>
 
--- a/doc/lualine.txt
+++ b/doc/lualine.txt
@@ -506,11 +506,12 @@ using a plugin which is supported you can load it this way:
 <
 
 Available extensions
+    * chadtree
     * fugitive
     * fzf
     * nerdtree
-    * chadtree
     * nvim-tree
+    * quickfix
 
 --------------------------------------------------------------------------------
 CONGIG EXAMPLES                                        *lualine_config_examples*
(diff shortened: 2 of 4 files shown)


shadmansaleh:
This looks alright to me . I was wondering since all the extentions just deepcopy section to inactive_sections . Why not just we load same section for both inactive and active in status_dispatch()

hoob3rt:
> @hoob3rt is this a bug with extentions in genaral ? When I open the quickfix I get the quickfix extention . But if I move out of the window and move in again It becomes normal status . 

this does not happen on my machine. No idea why that happens

However there is an issue with extensions in general. If you do `cclose` and `copen` then there will be a default statusline on the quickfix list. I think we need a new autocmd for that in addition to `WinEnter` and `BufEnter`

hoob3rt:
> This looks alright to me . I was wondering since all the extentions just deepcopy section to inactive_sections . Why not just we load same section for both inactive and active in status_dispatch()

yeah, I will open a new pr with that

shadmansaleh:
BufWinEnter change completely broke the extention for me. Now I don't get the extention at all

hoob3rt:
> > However there is an issue with extensions in general. If you do cclose and copen then there will be a default statusline on the quickfix list.
> 
> Strange this works fine for me . I get the quickfix extention if I cclose and copen .

are you sure you are getting the extension and not the default statusline? It is VERY similar for quickfix list 
Screenshot 2021-05-11 at 20 19 08

hoob3rt:
> When did you last compile neovim ?

just now

shadmansaleh:
I'm sure
Screenshot

hoob3rt:
> Also even in your screenshot the "Quickfix List" isn't highlighted right :/

it is... this is the gruvbox inactive highlight

hoob3rt:
I removed the faulty `BufWinEnter` it did cause some problems. This works perfectly fine on my side, just like any other extension. Create an issue with what's going on on your side when you have time

shadmansaleh:
I've found my issue . I had types extentions instead of extensions :P . My quickfix statusline was being set by qf filetype file . It was quite similar "QuickfixList quickfix title" so I thought it was the extension

PR #234, opened 2021-05-12
enhancement: show diff when component is loaded

jlbas:
This makes the diff component present when initialized. Currently, when opening a file on startup, the component only appears once the `BufEnter` autocmd's are executed.


shadmansaleh (approved):

hoob3rt:
looks good, thanks :)

Issue #227, opened 2021-05-11
Feat: Theme for right side items


WhyNotHugo:
### Requested feature

Allow configuring theme for right-side items.

### Motivation

I'm trying to port my lightline setup to lualine, but it doesn't seem to be possible to theme the right hand side -- so elements at the far right will always look off.

For example, the left-most element usually displays the `mode`, so it makes sense for that one to stand out a lot, however, the right-most element seldom needs to stand out so much.

hoob3rt:
you can already color each component however you like, see `options for components section` in README

shadmansaleh:
@WhyNotHugo with #238 merged you can now theme the right side separately . Just define x,y,z along with a,b,c in your themes.

PR #238, opened 2021-05-12
Allow x,y,z section to be themed independenly

shadmansaleh:
If theme for x,y,z is unspecified the default to c,b,a respectively

Follow up on #31 
resonable solution to #227

The refractors I did to highlighter made it simple to allow this . So just flipping the switch here :)

@hoob3rt I think we should allow this . This is 100% backword compatable and absolutly optional . But makes asymatric themeing easily possible .


hoob3rt:
lgtm, I'm not a big fan of this but after so many requests and little to no code added I can agree with this change

PR #241, opened 2021-05-13
feat: add separate hint diagnostic

timbedard:
I split the hint diagnostics out to be displayed separately. I'm not really sure why they were combined.

The default color and symbol can be changed to whatever folks prefer. I just put in what seemed useable.

It also looks like style warnings and errors were being ignored from ALE, so I added those into the error and warning counts, respectively. Let me know if this is something we don't actually want though.

I haven't had the opportunity to test yet with ALE or coc.nvim, so I'd love help with that. Otherwise I'll try to get to it this weekend if time allows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,12 +292,13 @@ sections = {
       -- nvim_lsp, coc, ale, vim_lsp
       sources = nil,
       -- displays diagnostics from defined severity
-      sections = {'error', 'warn', 'info'},
+      sections = {'error', 'warn', 'info', 'hint'},
       -- all colors are in format #rrggbb
       color_error = nil, -- changes diagnostic's error foreground color
       color_warn = nil, -- changes diagnostic's warn foreground color
       color_info = nil, -- Changes diagnostic's info foreground color
-      symbols = {error = 'E', warn = 'W', info = 'I'}
+      color_hint = nil, -- Changes diagnostic's hint foreground color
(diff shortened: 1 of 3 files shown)


hoob3rt:
> The default color and symbol can be changed to whatever folks prefer. I just put in what seemed useable.

I would change the default **symbol** to [this](https://fontawesome.com/icons/lightbulb)

> It also looks like style warnings and errors were being ignored from ALE, so I added those into the error and warning counts, respectively. Let me know if this is something we don't actually want though.

Yeah that's good, we want that.

> I haven't had the opportunity to test yet with ALE or coc.nvim, so I'd love help with that. Otherwise I'll try to get to it this weekend if time allows.

Have you had time to test this already? If you did then change the symbol and we are good to go :smile:

timbedard:
I'm down for the lightbulb. Only reason I'd avoided it was too try and keep it unicode for compatibility.

Nah, haven't had a chance to test yet. 😅 Hopefully soon though.

timbedard:
@hoob3rt Alright, I did some testing with coc and ALE and everything seems to be working fine.

I also changed the icon to a lightbulb. Most patched fonts (i.e. Nerd Fonts) don't have the icon variants from Font Awesome (just the basic one), so I used another that seemed like it fit the best next to the other symbols.

hoob3rt:
looks awesome, thank you sir :smile:

PR #248, opened 2021-05-16
Quickfix extension: label for location list

HiPhish:
Adds a separate label for locations lists than for quickfix lists. The logic is a bit roundabout, but I don't know of any better way to tell the difference between the two list types.

### DIFF
--- a/lua/lualine/extensions/quickfix.lua
+++ b/lua/lualine/extensions/quickfix.lua
@@ -1,14 +1,26 @@
 -- Copyright (c) 2020-2021 hoob3rt
 -- MIT license, see LICENSE for more details.
-local function quickfix() return 'Quickfix List' end
+--
+local function is_loclist()
+  return vim.fn.getloclist(0, {filewinid = 1}).filewinid ~= 0
+end
 
-local function quickfix_title() return vim.fn.getqflist({title = 0}).title end
+local function label()
+  return is_loclist() and 'Location List' or 'Quickfix List'
+end
+
+local function title()
+  if is_loclist() then
+    return vim.fn.getloclist(0, {title = 0}).title
+  end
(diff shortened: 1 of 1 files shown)


lua/lualine/extensions/quickfix.lua
```diff
@@ -1,8 +1,21 @@
 -- Copyright (c) 2020-2021 hoob3rt
 -- MIT license, see LICENSE for more details.
-local function quickfix() return 'Quickfix List' end
+--
+local function is_loclist()
+	local filewinid = vim.fn.getloclist(0, {filewinid = 1}).filewinid
+	return filewinid ~= 0
```
hoob3rt, line 6: this can be a oneliner

lua/lualine/extensions/quickfix.lua
```diff
@@ -1,8 +1,21 @@
...
+end
 
-local function quickfix_title() return vim.fn.getqflist({title = 0}).title end
+local function quickfix()
```
hoob3rt, line 9: probably should be named differently now

lua/lualine/extensions/quickfix.lua
```diff
@@ -1,8 +1,21 @@
...
+	return is_loclist() and 'Location List' or 'Quickfix List'
+end
+
+local function quickfix_title()
```
hoob3rt, line 13: same here

hoob3rt (changes requested):
> The logic is a bit roundabout, but I don't know of any better way to tell the difference between the two list types

It's okay, I couldn't find a better way myself. Some small changes and we can merge this

HiPhish:
I have updated the PR to your wishes.

hoob3rt:
looks good now, thank you :smile:

PR #255, opened 2021-05-21
Feat: Allow transitional separators to be set on component basis

shadmansaleh:
This separates the transitional separators from section . And allows them to be uses on component scope.
Also the `statusline` function is refractored to simplify it .

Example :
Screenshot


shadmansaleh:
Merged in https://github.com/shadmansaleh/lualine.nvim

shadmansaleh:
Part of #311

PR #278, opened 2021-07-01
Update README.md

ofsaleem:
i was looking at the instructions in the customizing themes section, and the colorscheme (solarized_dark in this case) didn't load until i put quotes around it there. im not sure if its always required, but it was in my case, so perhaps it should be fixed in the readme


shadmansaleh:
The README is correct . It was showing using table as theme . custom_grouvebox is a table that was created previously . When you pass a string as option it loads the theme from /lua/lualine/themes folder when you pass a table it uses that table as theme

PR #265, opened 2021-05-28
fix: estimate_space_available must include window width

NullVoxPopuli:
Resolves: https://github.com/hoob3rt/lualine.nvim/pull/224#issuecomment-850633672

the estimated_space_available _must_ subtract from the window width. The 40 value is the estimated amount of space decorated on either side (combined).

Before: Top: absolute, bottom: relative

image

image

After: Top: absolute, bottom: relative
image

PR #279, opened 2021-07-01
feat: filename trucation as option

fladson:
👋 hello there!

I just discovered this and it has everything I wanted for a status line, great work, and thank you!

One thing I noticed is that the filename truncation is not taking into consideration if I have a lot of free space. It is currently hardcoded to `40`, so I thought about having the possibility to control that as an option. What do you think?

I changed some things just as a POC, so I am very open to suggestions on how to achieve this, if you have interest, of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,7 @@ sections = {
       'filename',
       file_status = true, -- displays file status (readonly status, modified status)
       path = 0 -- 0 = just filename, 1 = relative path, 2 = absolute path
+      truncate = 0 -- 0 = high truncation, 1 = mid truncation, 2 = low truncation
     }
   }
 }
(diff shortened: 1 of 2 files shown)


shadmansaleh:
#265 might be a better solution this . You can take a look at that and see if that satisfies you .

fladson:
> #265 is a better solution this imo . You can take a look at that and see if that satisfies you .

Thanks for the suggestion @shadmansaleh, I like that approach, we could maybe combine both? Because the hardcoded value is an assumption of the number of components on the line so I still think it would be nice to be able to configure this as an option for the cases where there are not many components.

shadmansaleh:
I think it'll be better if we expose estimated_space_available as a option . So users can specify the number they need . Instead of having to chose between 3 modes . We need a better name for that option though .

I think we can combine both #265 and this.

shadmansaleh:
I believe this is resolved in https://github.com/shadmansaleh/lualine.nvim/commit/d3600399692e734854cf8d51a6480f9b8bd98461

 There I've provided access to estimated space available wuth shorting_target option .

PR #295, opened 2021-07-14
feat: allow xterm values in theme

sbaildon:
Would you be willing to accept patches that allow setting themes using xterm 256 color codes?

I made these changes because I want `lualine` to inherit my terminal colours without having to change my lualine config.

Example theme

```lua
return {
	normal = {
		a = {bg = 4, fg = 15},
		b = {bg = 19, fg = 15},
		c = {bg = 18, fg = 7}
	},
	insert = {
		a = {bg = 3, fg = 0},
		b = {bg = 19, fg = 15},
		c = {bg = 18, fg = 7}
	}
}
```

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -6,22 +6,28 @@ local utils = require 'lualine.utils.utils'
 local section_highlight_map = {x = 'c', y = 'b', z = 'a'}
 local active_theme = nil
 
+local function context_hl(command, color_value, context)
+  if not color_value or color_value == 'none' then return end
+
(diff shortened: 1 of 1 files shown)


shadmansaleh:
I've resolved this in #311 .
With that you can set fg/bg using #rrggbb/color_name(like red/gray)/cterm_color(0-255) .

sbaildon:
> I've resolved this in #311 .
> With that you can set fg/bg using #rrggbb/color_name(like red/gray)/cterm_color(0-255) .

Ace :-)

PR #304, opened 2021-07-22
Add ability to set fg and bg colors to diagnostics section

gillyb:
I did this pretty quickly, because I didn't like the contrast of the colors in the diagnostics section.
I realized there is no way to change the background colors.

If you are willing to merge this, then let me know first, and I will fix the documentation to match and make sure the tests are updated and pass.


shadmansaleh:
Implemented in [shadmansaleh/lualine.nvim](https://github.com/shadmansaleh/lualine.nvim) #311

Issue #263, opened 2021-05-27
Bug: 


jbromley:
### Self Checks

- [ X ] I'm using the latest lualine and neovim nightly.
- [ X ] I didn't find the issue in existing issues or prs.

### How to reproduce the problem
Run neovim with the configuration reproduced below, which attempts to use the `auto` theme.

### Actual behaviour
Neovim launches and produces the error message `theme auto not found, defaulting to gruvbox`. The gruvbox theme is applied.

### Expected behaviour

lualine loads lualine/themes/auto.lua and attempts to use the current theme colors.

### Aditional information

The file `auto.lua` exists in the themes directory and loading several other themes has worked without any problems. It appears the line that requires the theme file is failing.

### Minimal config to reproduce the issue

<details>
<summary>Config</summary>

```
-- Set up the package manager and install lualine.
vim.cmd 'packadd paq-nvim'
local paq = require('paq-nvim').paq
paq {'savq/paq-nvim', opt = true}
paq {'kyazdani42/nvim-web-devicons', opt = true}
paq {'hoob3rt/lualine.nvim'}

require('lualine').setup {options = { theme = 'auto' }}
```
</details>

jbromley:
I looked into this a little further and this is what I've found. 

The problem is that the call to `utils.extract_highlight_colors('Normal', 'bg')` at `lualina/themes/auto.lua:120` was returning `nil` and calling `get_color_brightness` (which calls `rgb_str2num`) on this nil was causing the load of `auto.lua` to fail. One odd thing is that I inspected the value returned for the Normal highlight group by `vim.api.nvim_get_hl_by_name` and got the following:
```
{ [true] = 6 }
```
so I don't know if there is a deeper problem with the vim.api call. I could get `auto.lua` to load by checking if the returned highlight group color was nil before calling `get_color_brightness`, like so
```
-- Change brightness of colors
-- darken in case of light theme, lighten in case of dark theme
local normal_color = utils.extract_highlight_colors('Normal', 'bg')
if normal_color ~= nil and get_color_brightness(normal_color) > 0.5 then
  brightness_modifier_parameter = -brightness_modifier_parameter
end
```

shadmansaleh:
Do you have a trasparant background in your colorscheme ?

jbromley:
[...]

PR #308, opened 2021-07-24
Fixed using 'auto' theme with default colorscheme.

jbromley:
When setting theme to 'auto' and using the default colorscheme, lualine would report "theme not found, defaulting to gruvbox". Fixed by checking that we get an actual color back when calling utils.extract_highlight_colors before attempting to apply a brightness modifier.

### DIFF
--- a/lua/lualine/themes/auto.lua
+++ b/lua/lualine/themes/auto.lua
@@ -117,7 +117,8 @@ local colors = {
 -- Change brightness of colors
 -- darken incase of light theme lighten incase of dark theme
 
-if get_color_brightness(utils.extract_highlight_colors('Normal', 'bg')) > 0.5 then
+local normal_color = utils.extract_highlight_colors('Normal', 'bg')
+if normal_color ~= nil and get_color_brightness(normal_color) > 0.5 then
   brightness_modifier_parameter = -brightness_modifier_parameter
 end
 


jbromley:
This pull request is meant to correct issue #263 (Using auto theme with default nvim colorscheme does not work).

shadmansaleh:
Only run `brightness_modifer()` when normal color is found . line 126

jbromley:
Yes, I will make that change tonight.

jbromley:
I moved the loop that applies the brightness modifier inside the if guard statement that checks if normal_color was found.

lua/lualine/themes/auto.lua
```diff
@@ -117,13 +117,14 @@ local colors = {
 -- Change brightness of colors
 -- darken incase of light theme lighten incase of dark theme
 
-if get_color_brightness(utils.extract_highlight_colors('Normal', 'bg')) > 0.5 then
+local normal_color = utils.extract_highlight_colors('Normal', 'bg')
+if normal_color ~= nil and get_color_brightness(normal_color) > 0.5 then
   brightness_modifier_parameter = -brightness_modifier_parameter
+  for name, color in pairs(colors) do
```
shadmansaleh, line 123: No this is wrong . We need this to run when get_color_brightness return < 0.5 too .

jbromley: Sorry about that. I have now separated the if that checks if the normal color is not nil from the if that checks if the brightness modifier should be negated. The brightness modifier is now applied whenever normal_color ~= nil.

shadmansaleh:
LGTM . If you'd like I could merge it in my fork with #311

jbromley:
That would be great if you merge it in with your changes. Thanks.

shadmansaleh:
Ok merged in [shadmansaleh/lualine.nvim](https://github.com/shadmansaleh/lualine.nvim) #311

PR #311, opened 2021-07-25
All changes made to my fork in the time @hoob3rt is absent :)

shadmansaleh:

PR #318, opened 2021-08-13
Add moonfly theme

bluz71:
Hello,

Author of [moonfly](https://github.com/bluz71/vim-moonfly-colors) and [nightfly](https://github.com/bluz71/vim-nightfly-guicolors) speaking.

I notice that the nightly theme is already supported (thanks).
I thought it would be nice to also add lualine support for my other theme (moonfly).

Cheers.

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -141,6 +141,14 @@ lualine try to match your colorscheme.
 <img width='700' src='https://user-images.githubusercontent.com/41551030/108648958-b12b2300-74bc-11eb-8232-550949cbbfb3.png'/>
 </p>
 
+### moonfly
+<p>
+<img width='700' src='https://user-images.githubusercontent.com/11382509/129304429-cd2af22d-3b88-4c17-bf99-99f7608576d4.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/11382509/129304441-81a6c88e-364c-476a-9050-0ada4cae1e3b.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/11382509/129304453-2b18900a-13a9-4476-9977-3875e55aabaf.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/11382509/129304469-37fb0168-0484-406b-8b2e-d309232612ab.png'/>
+</p>
+
 ### nightfly
 <p>
 <img width='700' src='https://user-images.githubusercontent.com/41551030/108648974-b9835e00-74bc-11eb-9c52-c5187aa3ff2c.png'/>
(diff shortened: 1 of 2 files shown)


shadmansaleh:
Great . But since you're the author of the colorscheme I think it'll be best if you put the theme in the colorscheme repo .That way you can tweek it later easily if you change some colors in the colorscheme without coming here for change.

Since you're already providing lightline & airline themes it'll be pretty much the same .

You'll just have to put the same `moonfly.lua` file in the same folder `/lua/lualine/themes/` in your repo and lualine will pick it up :)

bluz71:
Sure, I can do that quite easily.

But, I believe that should also happen with the nightfly theme then as well, adding `nightfly.lua` to my nightfly repo `/lua/lualine/themes` directory. Will that be a problem with the `nightly.lua` that exists here?

shadmansaleh:
> But, I believe that should also happen with the nightfly theme then as well, adding nightfly.lua to my nightfly repo /lua/lualine/themes directory. Will that be a problem with the nightly.lua that exists here?

You can put that in your repo too. Though same name in two places may cause conflict . If you want I can remove the nightfly theme from from [shadmansaleh/lualine.nvim](https://github.com/shadmansaleh/lualine.nvim) That's the one I'm currently maintaining .  I'll probably add something lualine loads theme from external plugin instead in from itself when there's a conflict .

bluz71:
Cool, I will add both themes to my two repos.

The `nightfly.lua` I add will be exactly the same as yours.

Cheers.

bluz71:
Is your `lualine` the one I should reference in my repositories or should it be this one?

Is this project abandoned whilst yours is mantained?

shadmansaleh:
For now I can't contact @hoob3rt. I haven't hard forked it yet . The changes and fixes that I've made is currently a pr to this repo #311 . I'll give @hoob3rt some more time . If he doesn't contact back I'll just hard fork this. So for now you can reference the one you like .

bluz71:
moonfly & nightfly no support lualine.

I will keep an eye on #311 to see if you hard fork, if you do I will change my README files to reference your version (if that becomes the maintained version).

Thanks.

shadmansaleh:
@bluz71 after https://github.com/shadmansaleh/lualine.nvim/commit/32727039a97ece7ee4eb6d889be0b3e50b570daa

nightfly theme from your repo will be loaded when the colorschme is installed instead of the one bundled with lualine :)

bluz71:
Thanks, that is good to know.

It is a nice feeling having both my themes now supporting lualine along with airline/lightline.

PR #346, opened 2021-09-27
Prevent E5108 when leaving LSP buffer that is not connected to a file

callebstrom:
### Background
When leaving a buffer without a proper file handle (e.g. a buffer containing decompiled java sources accessed via nvim LSP) I get the following error:
```
Error detected while processing BufLeave Autocommands for "*":
E5108: Error executing lua [string ":lua"]:1: E539: Illegal character <C>
```

### Changes
- Add `silent!` to affected `autocmd` declarations

### DIFF
--- a/lua/lualine/init.lua
+++ b/lua/lualine/init.lua
@@ -169,8 +169,8 @@ local function set_statusline()
     vim.api.nvim_exec([[
       augroup lualine
         autocmd!
-        autocmd WinLeave,BufLeave * lua vim.wo.statusline=require'lualine'.statusline()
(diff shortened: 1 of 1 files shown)


callebstrom:
@hoob3rt Does this look like a sane default or would you prefer to have it configurable via the `config` object. I.e. something like:
```
local config = {
  silence_autocommand_errors = true
}
```

shadmansaleh:
Use shadmansaleh/lualine.nvim

This repo is currently not maintained . See #311

I don't think this issue occures there as I've removed those autocommands.

PR #415, opened 2021-11-03
enhance(filetype): filetype option as fallback of file extension for icon detection

fitrh:
Enhances icon detection for files without extension e.g. shebang or
modeline by using filetype option before the extension.

### DIFF
--- a/lua/lualine/components/filetype.lua
+++ b/lua/lualine/components/filetype.lua
@@ -29,7 +29,9 @@ function M:apply_icon()
   local icon, icon_highlight_group
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
-    local f_name, f_extension = vim.fn.expand '%:t', vim.fn.expand '%:e'
+    local f_name = vim.fn.expand '%:t'
+    local f_extension = vim.api.nvim_buf_get_option(0, 'filetype')
+    f_extension = f_extension ~= '' and f_extension or vim.fn.expand '%:e'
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Can you check why this is causing test failure ?

fitrh:
I also wondering about that, the changes only add `vim.api.nvim_get_buf_option`

How can I run the test locally ?

shadmansaleh:
> How can I run the test locally ?

```
make test
```

fitrh:
I've tested it, and all test passes

shadmansaleh:
Ok in that case I'll keep this in hold and check why it's failing here when I get time.

shadmansaleh:
Nice this now passes.

lua/lualine/components/filetype.lua
```diff
@@ -30,6 +30,7 @@ function M:apply_icon()
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
     local f_name, f_extension = vim.fn.expand '%:t', vim.fn.expand '%:e'
+    f_extension = f_extension ~= '' and f_extension or vim.api.nvim_buf_get_option(0, 'filetype')
```
shadmansaleh, line 33: I think we should use `vim.bo` instead of `api.nvim_buf_get_option` since that's what used in rest of the code base.

fitrh: Sure

shadmansaleh:
Thanks :)

PR #421, opened 2021-11-05
fixup: missing hint value inside vim_lsp source

askobara:


shadmansaleh:
lgtm . Thanks for the fix.

PR #426, opened 2021-11-07
fix: Replace battle.net icon with sharing screen

Anakael:
I replaced battle.net icon with xf838 as it described in comment.
But I think it is still not appropriate icon and I'd appreciate suggestion about it.

### DIFF
--- a/lua/lualine/components/diagnostics/config.lua
+++ b/lua/lualine/components/diagnostics/config.lua
@@ -8,7 +8,7 @@ M.symbols = {
     error = ' ', -- xf659
     warn = ' ', -- xf529
     info = ' ', -- xf7fc
-    hint = '', -- xf835
+    hint = ' ', -- xf835
   },
   no_icons = { error = 'E:', warn = 'W:', info = 'I:', hint = 'H:' },
 }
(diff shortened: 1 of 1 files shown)


shadmansaleh:
IIRC symbol for hint is a `lightbulb` . It might have been miss labeled if xf838 is not that 🤦‍♂️ . Not sure where battle.net icon came from what font are you using ?

Anakael:
I am using Noto Sans with with `Nerd` patch.
Battle.net icon is xf835 and probably was copied by accident.

shadmansaleh:
> Battle.net icon is xf835

xf835 is lightbulb in nerdfonts according to https://www.nerdfonts.com/cheat-sheet

I think there's something wrong with the patch on font you're using . I'm using fira-code and it gets displayed as lightbulb.

You can fix the documentation error in this pr but I don't think we should change the hint icon to sharing screen

Anakael:
I've fixed commentary :)

shadmansaleh (approved):

shadmansaleh:
lgtm . Thanks :)

PR #433, opened 2021-11-12
feature(buffers): option to show buffer index

Ktan32:
Hey!

Here's a little feature to the `buffers` component which allows to display buffer index (also known as "number")

It's useful when you have mappings like `4gb` (open buffer 4).

image

Thanks for the plugin, btw!

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,7 +421,10 @@ sections = {
     {
       'buffers',
       show_filename_only = true, -- shows shortened relative path when false
-      show_modified_status = true -- shows indicator then bufder is modified
+      show_modified_status = true -- shows indicator then buffer is modified
+      mode = 0, -- 0 shows buffer name
+                -- 1 buffer index (bufnr)
+                -- 2 shows buffer name + buffer index (bufnr)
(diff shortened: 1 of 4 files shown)


shadmansaleh:
I think we should use a `mode` option like [tabs](https://github.com/nvim-lualine/lualine.nvim#tabs-component-options) component for this.

> Thanks for the plugin, btw!

Welcome. Thanks for opening a pr instead of issue :)

Ktan32:
@shadmansaleh 
Done, I've changed this option to a `mode = 0 | 1 | 2` like the `tabs` component.

Ktan32:
There is one test failing, though from a completely different component. Not sure if this test is flacky in general or I should double check this PR.

shadmansaleh:
It might be flacky, though hasn't acted like this before . I don't see any reason for it to fail either. I'll checkout the test while reviewing .

shadmansaleh:
The tests seem to pass locally. Maybe something wrong with CI. The changes looks good to me. Thanks for your contribution.

shadmansaleh (approved):

PR #436, opened 2021-11-13
Introduce Base 16 Ocean theme

tarebyte:
This PR adds the base16-ocean theme from http://chriskempson.com/projects/base16 which I borrowed the [vim-airline-themes implementation](https://github.com/vim-airline/vim-airline-themes/blob/97cf3e6e638f936187d5f6e9b5eb1bdf0a4df256/autoload/airline/themes/base16_ocean.vim) of.

Screen Shot 2021-11-13 at 10 34 20 AM
Screen Shot 2021-11-13 at 10 34 32 AM
Screen Shot 2021-11-13 at 10 34 50 AM
Screen Shot 2021-11-13 at 10 34 42 AM

Thanks for a great plugin!

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -43,6 +43,14 @@ it'll try it's best to generate one.
 It's a combination of ayu_light, ayu_dark & ayu_mirage. It loads one of these
 them based you your `g:ayucolor` option.
 
+### base16_ocean
+<p>
+<img width='700' src='https://user-images.githubusercontent.com/564113/141649910-1f14e0bc-98b6-4439-9e85-4584106022ad.png'/>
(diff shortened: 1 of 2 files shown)


shadmansaleh:
We are currently not accepting regular themes. I think themes make more sense along with colorschemes. Because the themes in this repo gets rarely updated, most of this repo is already themes and it's not possible for me to keep track if the colorscheme of the theme changes some of it's color. Many color schemes already support lualine and I'd like to encourage that.

You can use the theme yourself without adding it to this repo by placing it in your `~/.config/nvim/lua/lualine/themes/` folder.
Or can even have the theme as separate plugin just by placing it in `/lua/lualine/themes/` folder. You can [checkout](https://github.com/nvim-lualine/lualine.nvim/wiki/Plugins).

PR #439, opened 2021-11-18
tests: add more tests

shadmansaleh:


shadmansaleh:
This puts us at `61.4%` code coverage. Though there are still some edge cases that we should test in this . At least I found a few bugs while writing tests :D

For some reason some tabline tests pass locally but not in CI, I'm commenting out all tabline tests for now . Will check out later why they were failing.

PR #448, opened 2021-11-22
Add branch component option to the README

jupblb:
I found out that there is already code handling icon change in https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/components/branch/init.lua


shadmansaleh:
`icon` is actually a [general option](https://github.com/nvim-lualine/lualine.nvim#general-component-options) means it's available for all components . It's not specific to `branch` component.

jupblb:
Right! Sorry for missing that. 🤦

PR #449, opened 2021-11-22
fix(diagnostics): adapt nvim source to breaking change

clason:
Explicitly pass `bufnr` for both `nvim` and `nvim-lsp` diagnostic sources, as required by upstream breaking change: https://github.com/neovim/neovim/pull/16405

(Separate commits in case you don't want to change the "legacy" source.)


mjlbach:
This will be unnecessary in a few minutes.

PR #455, opened 2021-11-27
Create simple.lua

bryant-the-coder:
I created this as a theme. I was asking a tons of simple yet annoying question to do this theme. I hope my PR is accepted.

### DIFF
--- a/examples/simple.lua
+++ b/examples/simple.lua
@@ -19,15 +19,6 @@ local colors = {
   purple        = '#cd8fe0',
 }
 
-local conditions = {
-  buffer_not_empty = function()
-    return vim.fn.empty(vim.fn.expand '%:t') ~= 1
-  end,
-  hide_in_width = function()
-    return vim.fn.winwidth(0) > 80
-  end,
-}
-
 local function center()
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Example configs are there to show new ideas about lualine config. So users can easily incorporate those ideas in there config or take inspiration from them.

For example:
- evilline: Showed ways to break traditional lualine configs and get completely new look compared to default.
- bubbleline: Showcased using section separators on component when they were added.
- slanted-gap: Showed using custom components. And how lualine can be used easily to create unique looks.

What does this add ?

bryant-the-coder:
Some simple aesthetic looks maybe? Tried my best. It is ok if it doesn't fits. It suits me :)

shadmansaleh:
> Some simple aesthetic looks maybe? 

Sorry but asthetics isn't the point of examples they are there to show new ideas about configuring.

> Tried my best

Yes. Thanks for your effort and considering to contribute to lualine.

> It suits me :)

Glad to here that :)

bryant-the-coder:
Noted sir

Issue #462, opened 2021-11-30
Bug: expected table, got function, after neovim was inactive for some time


Gelio:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem

I don't have any good reproduction steps.

After leaving neovim idle for some time (> 5 minutes, but I didn't find a pattern as to when it happens), the following error is shown:

```
E5108: Error executing lua ...m/nvim-linux64/share/nvim/runtime/lua/vim/diagnostic.lua:704: opt[6.9283908576005e-310]: expected table, got function
stack traceback:
        [C]: in function 'error'
        vim/shared.lua:607: in function 'validate'
        ...m/nvim-linux64/share/nvim/runtime/lua/vim/diagnostic.lua:704: in function 'get'
        ...im-linux64/share/nvim/runtime/lua/vim/lsp/diagnostic.lua:344: in function 'get_count'
        ...line.nvim/lua/lualine/components/diagnostics/sources.lua:8: in function <...line.nvim/lua/lualine/components/diagnostics/sources.lua:7>
        ...line.nvim/lua/lualine/components/diagnostics/sources.lua:58: in function 'get_diagnostics'
        ...lualine.nvim/lua/lualine/components/diagnostics/init.lua:72: in function 'update_status'
        ...pack/packer/start/lualine.nvim/lua/lualine/component.lua:175: in function 'draw'
        .../packer/start/lualine.nvim/lua/lualine/utils/section.lua:26: in function 'draw_section'
        ...nvim/site/pack/packer/start/lualine.nvim/lua/lualine.lua:153: in function 'statusline'
        ...nvim/site/pack/packer/start/lualine.nvim/lua/lualine.lua:316: in function <...nvim/site/pack/packer/start/lualine.nvim/lua/lualine.lua:297>
Press ENTER or type command to continue
```

![image](https://user-images.githubusercontent.com/889383/144049865-8c640c43-f6f8-49da-87ba-3b08b0880e0a.png)

After that, lualine stops working and I am back to the default neovim statusline.

As I investigated the function calls, looks like there is nothing in lualine's code that is wrong here. The stacktrace mentions:

https://github.com/neovim/neovim/blob/04c7b550384b2246d37c1a893c0f858bc332cb8b/runtime/lua/vim/diagnostic.lua#L704-L707

and 

[...]

PR #463, opened 2021-12-02
upgrade defaults to target neovim 0.6.0

williamboman:
See #462. Not sure if this can be considered solving it as @Gelio seems to be still be running into the issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,20 +18,20 @@ jobs:
           git clone --depth 1 https://github.com/nvim-lua/plenary.nvim ~/.local/share/nvim/site/pack/vendor/start/plenary.nvim
           git clone --depth 1 https://github.com/kyazdani42/nvim-web-devicons ~/.local/share/nvim/site/pack/vendor/start/nvim-web-devicons
           ln -s $(pwd) ~/.local/share/nvim/site/pack/vendor/start
-      - name: Get neovim v0.5.0
+      - name: Get neovim v0.6.0
         uses: actions/cache@v2
         with:
           path: |
-            build/neovim/v0.5.0
+            build/neovim/v0.6.0
           key: ${{ runner.os }}-appimage-0.5
-      - name: Run Test on neovim v0.5.0
+      - name: Run Test on neovim v0.6.0
         run: |
-          test -d build/neovim/v0.5.0 || {
-            mkdir -p build/neovim/v0.5.0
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -18,20 +18,20 @@ jobs:
           git clone --depth 1 https://github.com/nvim-lua/plenary.nvim ~/.local/share/nvim/site/pack/vendor/start/plenary.nvim
           git clone --depth 1 https://github.com/kyazdani42/nvim-web-devicons ~/.local/share/nvim/site/pack/vendor/start/nvim-web-devicons
           ln -s $(pwd) ~/.local/share/nvim/site/pack/vendor/start
-      - name: Get neovim v0.5.0
```
williamboman, line 21: I could otherwise recommend using the [`rhysd/action-setup-vim@v1`](https://github.com/rhysd/action-setup-vim) action. I've been using it and it's been stable so far, it'll help cut down build times as well.

williamboman:
I pushed a bit more than I perhaps intended in the beginning. Let me know if you want me to undo some things. I think at least checking if current instance has `nvim-0.6.0` _and_ `nvim_lsp` is a source there should be a deprecation warning, the rest is mostly just updating defaults.

shadmansaleh:
I think using `vim.version()` will be more efficient then `vim.fn.has()` .  The notification should go in diagnostics component [initialization](https://github.com/nvim-lualine/lualine.nvim/blob/1ae4f0aa74f0b34222c5ef3281b34602a76b2b00/lua/lualine/components/diagnostics/init.lua#L18) not in `config.lua` to keep it isolated to diagnostics component . Also should use [lualine-notices](https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/utils/notices.lua) .

The CI changes should be in a separate pr . I was actually planning on spliting the CI job in matrix and run test on multiple versions.

williamboman:
> I think using `vim.version()` will be more efficient then `vim.fn.has()` .

Hmm seeing as it's called twice I'd say any perf penalties are entirely negligible, but maybe I misunderstand? With `vim.version()` I guess it'd look something like:

```lua
local version = vim.version()
local is_06 = version.major >= 0 and version.minor >= 6
```

> The CI changes should be in a separate pr . I was actually planning on spliting the CI job in matrix and run test on multiple versions.

Yeah totally agree, will undo

williamboman:
Or actually seeing as you were about to look into this anyway, I'll close this and let you do it the proper way :D

PR #470, opened 2021-12-06
Add some extensions

kuntau:
Add some extensions that I use, nothing fancy, just to hide the full statusline on small window ✌🏽 

1. [Neogit](https://github.com/TimUntersberger/neogit)
2. [Trouble](https://github.com/folke/trouble.nvim)
3. [Undotree](https://github.com/mbbill/undotree)
4. [Vim-Plug](https://github.com/junegunn/vim-plug)

### DIFF
--- a/README.md
+++ b/README.md
@@ -627,6 +627,10 @@ extensions = {'quickfix'}
 - quickfix
 - toggleterm
 - symbols-outline
+- [Neogit](https://github.com/TimUntersberger/neogit)
+- [Trouble](https://github.com/folke/trouble.nvim)
+- [Undotree](https://github.com/mbbill/undotree)
+- [Vim-Plug](https://github.com/junegunn/vim-plug)
 
 #### Custom extensions
 
--- a/README.md
+++ b/README.md
@@ -631,6 +631,7 @@ extensions = {'quickfix'}
 - [Trouble](https://github.com/folke/trouble.nvim)
 - [Undotree](https://github.com/mbbill/undotree)
 - [Vim-Plug](https://github.com/junegunn/vim-plug)
+- [Startify](https://github.com/mhinz/vim-startify)
 
 #### Custom extensions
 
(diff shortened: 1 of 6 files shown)


gegoune:
Maybe also an empty one for Telescope?

shadmansaleh:
I don't see any reason to add extensions like this to repo. These add nothing specific to the plugin at all .

If you need to hide full statusline just use [custom extensions](https://github.com/nvim-lualine/lualine.nvim#custom-extensions) as part of your config.

For example instead of this pr you can just add following to your config.

```lua
local ft_extension = {
    sections = {lualine_a = {'filetype'}},
    filetypes = {'startify', 'Trouble', 'undotree', 'vim-plug'}
}
local neogit_extension = {
    sections = {lualine_a = {'branch'}},
    filetypes = {'NeogitStatus'}
}

require'lualine'.setup {extensions = {ft_extension, neogit_extenstion}}
```

gegoune:
@shadmansaleh I have tried to use that to set up custom status line for Telescope:

```lua
local telescope = {
  sections = { lualine_a = { 'mode' } },
  filetypes = { 'TelescopePrompt', 'TelescopeResults' },
}
```
Unfortunately it is still drawing status line from `inactive_sections`. I figured out why, but am not sure sure how to fix it. If I add `'filetype'` section to `inactive_sections` I can see that filetype is still the same as the buffer from which I have opened  telescope. For example, if I open telescope from lua, I get lua in `inactive_sections`'s `'filetype'` rather than any of Telescope ones.

Any idea how to tackle that?

shadmansaleh:
> Unfortunately it is still drawing status line from inactive_sections

floating windows don't have statusline only splits do . The statusline you're seeing belongs the window behind telescopes fold . That's why it's drawn in that windows context . Which is an inactive window in this case as focus is on telescopes floating window.

gegoune:
@shadmansaleh Thank you kindly for taking your time to explain it. It makes very much sense now.

Issue #473, opened 2021-12-08
Bug: `vim.lsp.diagnostic.get_count()` is deprecated


numine777:
### Self Checks

- [X] I'm using the latest lualine.
- [X] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem
1. Update to neovim nightly
2. Open neovim with lualine plugin

### Expected behaviour
Neovim functions normally with lualine integration

### Actual behaviour
A persistent notification for `vim.lsp.diagnostic.get_count() is deprecated` will appear.

### Minimal config to reproduce the issue
`require 'lualine'.setup()`

<details>
<summary>Config</summary>

</details>

### Aditional information
This is root caused to `lua/lualine/components/diagnostics/sources.lua`.  I will work on setting up a PR to solve this issue, but wanted to bring awareness.

numine777:
Pull request https://github.com/nvim-lualine/lualine.nvim/pull/474 solves the issue

shadmansaleh:
use the `nvim` source instead of `nvim-lsp` for diagnostics.

numine777:
> Use the `nvim` source instead of `nvim-lsp` for diagnostics.

That's what my PR does 😉

shadmansaleh:
> That's what my PR does 😉

No your pr is just implementing the functionality that already exists in `nvim` diagnostics source to `nvim-lsp` diagnostics source .

numine777:
> > That's what my PR does 😉
> 
> No your pr is just implementing the functionality that already exists in [nvim](https://github.com/nvim-lualine/lualine.nvim/blob/d68631d2c02bd31d937349d739c625cc81dd9ac1/lua/lualine/components/diagnostics/sources.lua#L14-L24) diagnostics source to `nvim-lsp` diagnostics source 😄

Got it, thanks!

flaviusbuffon:
When will you solve this problem?
Thanks！

shadmansaleh:
> When will you solve this problem?

TBH solution is just to change default config .

For now any one facing this issue just need to slightly change there config like following.

```diff
- { 'diagnostics', sources = {'nvim-lsp'}}
+ { 'diagnostics', sources = {'nvim'}}
```

GaetanLepage:
No, I still have the deprecation warning...

gegoune:
There is bunch of plugins that did not get updated to avoid deprecation warning yet. lualine is not one of them (given `nvim` source is used).

[...]

PR #474, opened 2021-12-08
Update diagnostics sources to be compatible with nightly

numine777:
-Replace call to vim.lsp.diagnostics.get_count() with #vim.diagnostics.get() to be compatible with neovim nightly


shadmansaleh:
https://github.com/nvim-lualine/lualine.nvim/issues/473#issuecomment-989320853

PR #483, opened 2021-12-10
Feat: customise diagnostic rendering

tapayne88:
Hi 👋 thanks for the great project!

I'm hoping to be able to customise the rendering of a single diagnostic based on the count. I can't see a way of doing it currently so sometime like this would be super useful to me.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,7 @@ sections = {
       colored = true, -- displays diagnostics status in color if set to true
       update_in_insert = false, -- Update diagnostics in insert mode
       always_visible = false, -- Show diagnostics even if count is 0, boolean or function returning boolean
+      render = function(icon, count) return icon .. count end, -- Customise diagnotic render, function returning string
     }
   }
 }
--- a/lua/lualine/components/diagnostics/config.lua
+++ b/lua/lualine/components/diagnostics/config.lua
@@ -20,6 +20,7 @@ M.options = {
   always_visible = false,
   sources = { vim.fn.has 'nvim-0.6' == 1 and 'nvim_diagnostic' or 'nvim_lsp', 'coc' },
   sections = { 'error', 'warn', 'info', 'hint' },
+  render = function(icon, count) return icon .. count end,
   diagnostics_color = {
     error = {
       fg = utils.extract_color_from_hllist(
--- a/lua/lualine/components/diagnostics/init.lua
+++ b/lua/lualine/components/diagnostics/init.lua
@@ -141,13 +141,13 @@ function M:update_status()
     end
     for _, section in ipairs(self.options.sections) do
       if diagnostics_count[section] ~= nil and (always_visible or diagnostics_count[section] > 0) then
-        table.insert(result, colors[section] .. self.symbols[section] .. diagnostics_count[section])
+        table.insert(result, colors[section] .. self.options.render(self.symbols[section], diagnostics_count[section]))
       end
     end
   else
     for _, section in ipairs(self.options.sections) do
       if diagnostics_count[section] ~= nil and (always_visible or diagnostics_count[section] > 0) then
-        table.insert(result, self.symbols[section] .. diagnostics_count[section])
+        table.insert(result, self.options.render(self.symbols[section], diagnostics_count[section]))
       end
     end
   end


tapayne88:
This is an example of what I'm hoping to do

```lua
local function diagnostic_renderer(icon, count)
    if count > 0 then
        return string.format(" %s%s ", icon, count)
    else
        return ''
    end
end
```

Screenshot 2021-12-10 at 19 08 11

This way I can keep the diagnostic rendered (with separators). Seems like I could do this with `fmt` but I'd have to parse the count out of the final string which doesn't feel great 😄

shadmansaleh:
You can already achive this easily by spliting diagnostics sections as separate components like slanted-gap [example](https://github.com/nvim-lualine/lualine.nvim/blob/3a17e2dbca8accbd3733b4fb076f01e00a456ba4/examples/slanted-gaps.lua#L82-L93).

tapayne88:
> edit: Actually I'm not sure if I understand what you want to do :D

Haha yeah! It's not easy to explain! 😄 

I basically want the sections to always show but only render the icon and count if `count > 0`, otherwise be an empty block.

**Desired**
Screenshot 2021-12-10 at 19 08 11

**Master**
Screenshot 2021-12-11 at 00 46 33

Config
```lua
{
    'diagnostics',
    source = {'nvim_diagnostic'},
    sections = {'error'},
    diagnostics_color = {
        error = {bg = colors.nord14_gui, fg = colors.nord3_gui}
    },
    symbols = {error = lsp_symbols.error},
    separator = {left = "", right = ""},
    render = function(icon, count)
        if count > 0 then
            return string.format(" %s%s ", icon, count)
        else
            return ''
        end
    end,
    always_visible = true,
    cond = conditions.has_lsp
}
```

tapayne88:
This is a bit of a work around I have

```lua
{
    ...,
    fmt = function(status)
        -- parse status which looks like
        -- %#lualine_diagnostics_info_0_1_2_no_mode# 0
        local hightlight, icon, num =
            string.match(status, '([%%#%w_]+)(.+) (%d+)')

        if num and tonumber(num, 10) > 0 then
            -- stitch everything back together with some padding
            -- only want the padding if we're not empty
            return string.format('%s %s %s ', hightlight, icon, num)
        end

        -- Count is 0 so return highlight so we still get the colored slants
        return hightlight
    end,
}
```

shadmansaleh:
Oh so you want the slant colors to be based on hidden diagnostics section . That's cool . Also looks like you've figured out fmt :D . I think we can drop this pr as `render` seems really specific to this usecase . I don't think many will find use for this option . Also fmt is can achive the same effect . What do you say ?

tapayne88:
Yeah, I think we can close it down - thanks for you help!

For anyone interested this is my workaround

```lua
local diagnostic_section = function(cfg)
    local default_cfg = {
        -- custom diagnostic component to allow empty content
        diagnostic_empty,
        source = {'nvim_diagnostic'},
        separator = {left = "", right = ""},
        -- no padding so the slanty isn't too wide when no diagnostics
        padding = 0,
        fmt = function(status)
            if tonumber(status, 10) > 0 then
                -- stitch everything back together with some padding
                -- only want the padding if we're not empty
                return string.format(' %s%s ', cfg.symbol, status)
            end

            -- Count is 0 so don't return content
            return ''
        end,
        -- supress the symbols, default still shows 'E: 1' etc.
        symbols = {error = '', warn = '', hint = '', info = ''},
        -- don't want any color output adding to the diagnostics
        colored = false,
        -- always show the slanty, it'll be empty if there are none for that type
        always_visible = true,
        -- only show when we have an lsp attached - this may need updating if I
        -- use other sources for diagnostics
        cond = conditions.has_lsp
    }
    return vim.tbl_extend("force", default_cfg, cfg)
end

local sections = {
    lualine_a = {
        diagnostic_section {
            sections = {'error'},
            color = {
                bg = lsp_colors("error"),
                fg = color({dark = "nord3_gui", light = "fg"})
            },
            -- custom var to be used by fmt function
            symbol = lsp_symbols.error
        }
    }
}

```

One thing that might be useful @shadmansaleh is the ability to still render a component even if the content is empty - perhaps a boolean option `allow_empty`? I've got a few copies of components that do this

```lua
local diagnostic_empty = require('lualine.components.diagnostics'):extend()
function diagnostic_empty:draw(default_highlight, is_focused)
    -- Copied from lualine.component and modified to allow empty status to render
    self.status = ''
    self.applied_separator = ''

    if self.options.cond ~= nil and self.options.cond() ~= true then
        return self.status
    end
    local status = self:update_status(is_focused)
    if self.options.fmt then status = self.options.fmt(status or '') end
    -- if type(status) == 'string' and #status > 0 then
    self.status = status
    self:apply_icon()
    self:apply_padding()
    self:apply_highlights(default_highlight)
    self:apply_section_separators()
    self:apply_separator()
    -- end
    return self.status
end
```

What do you think? Happy to put a PR together

shadmansaleh:
I'm afraid `allow_empty` also falls in niche use case, Sorry ;| 

The thing is it's not possible make a statusline that goes according to everyone's will . So the idea is we add stuff that are useful to many in this repo and just provide the flexibility so if an user wants to they can customize their statusline to their liking . That's why niche use cases aren't meant to be supported out of the box . Hopefully you can understand.

tapayne88:
> I'm afraid `allow_empty` also falls in niche use case, Sorry ;|
> 
> The thing is it's not possible make a statusline that goes according to everyone's will . So the idea is we add stuff that are useful to many in this repo and just provide the flexibility so if an user wants to they can customize their statusline to their liking . That's why niche use cases aren't meant to be supported out of the box . Hopefully you can understand.

Absolutely, appreciate it's a judgement call and you'll have a better idea than me of how others are using this project!

Does feel a little like it might have utility when used with separators like the slanted block but I'm happy with my workaround.

Thanks again for your help and guidance!

PR #496, opened 2021-12-18
feat(filename): add unnamed option

Saecki:
This makes the `[No Name]` text, which is shown when a buffer is unnamed, customizable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,10 +503,15 @@ sections = {
   lualine_a = {
     {
       'filename',
-      file_status = true,  -- displays file status (readonly status, modified status)
-      path = 0,            -- 0 = just filename, 1 = relative path, 2 = absolute path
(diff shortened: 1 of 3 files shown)


shadmansaleh:
Documentation needs to be updated in README.md . Otherwise seems alright .

Saecki:
Oh I updated the docs locally by running the script. I must have read over the CI part.

shadmansaleh (approved):

shadmansaleh:
lgtm . Thanks 👍🏾 

Don't worry about the warning . It's not an issue if changes doc/lualine.txt is modified by docgen .

ryanhuanli:
'buffer' and 'tab' components also have hard coded '[No Name]', so if we use unnamed= to change it for 'filename' here, they will mismatch and cause confusion, I think it's probabaly better make it a global option to use in all 3 places? thanks.

PR #497, opened 2021-12-18
fix: don't hide readonly symbol when modified

Saecki:
Currently if you're editing a readonly file and made any changes, only the modified symbol is shown. With this PR it will show both.

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -63,7 +63,8 @@ M.update_status = function(self)
   if self.options.file_status then
     if vim.bo.modified then
       data = data .. self.options.symbols.modified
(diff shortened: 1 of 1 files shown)


shadmansaleh:
How are you editing a readonly file ? neovim should prevent you from doing such a thing.

Saecki:
If the buffer is not modifiable yes. If it is readonly, it does warn you but you can choose to ignore that warning. In that case the readonly refers to the filesystem permissions for the file, so you can edit in vim but can't save it.

shadmansaleh:
I guess it can't hurt if it only gets triggered on a rare edge case .

shadmansaleh (approved):

Issue #523, opened 2022-01-03
Bug: attempt to index field 'diagnostic' (a nil value)


fidgetingbits:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

I am using commit lualine b18b7ee and NVIM v0.6.0-dev+43-g02bf251bb.

### How to reproduce the problem

I have the config entry below, and I typically run all of my terminals
in vim, so I start a new terminal directly via vim using this i3 command `bindsym
$mod+Return exec kitty nvim -c ':term'`:

At some point I started seeing a lualine error on terminal startup.

### Expected behaviour

No error on startup.

### Actual behaviour

```
Error detected while processing TermEnter Autocommands for "*":
E5108: Error executing lua ...line.nvim/lua/lualine/components/diagnostics/sources.lua:18: attempt to index field 'diagnostic' (a nil value)
```

### Minimal config to reproduce the issue

<details>
<summary>Config</summary>

let g:custom_title_prefix ='VIM MODE:%{mode()} RPC:%{v:servername} %{FugitiveStatusline()}'
let g:custom_title_suffix =' (%f) %t'
let g:custom_title=g:custom_title_prefix . g:custom_title_suffix

function SetTerminalTitleString(timer)
    if exists('b:term_title')
        " we don't set {b:term_title} directly in the title string, because the
        " once the b:term_title goes out of scope we don't want that buffer
        " still trying to access it, so we resolve the actual string here
        " instead
        let &titlestring = g:custom_title_prefix . ' TERM:' . b:term_title . g:custom_title_suffix
    endif
endf

function StopTerminalTitleString()
    if exists('b:terminal_title_timer')
        call timer_stop(b:terminal_title_timer)
        let &titlestring = g:custom_title
    endif
endf

augroup terminal
    autocmd TermEnter *
                \ set relativenumber |
                \ let &titlestring=g:custom_title |
                \ let b:terminal_title_timer = timer_start(500, 'SetTerminalTitleString', {'repeat': -1}) |
                \ redraw
augroup end
</details>

### Additional information

I confirmed that simply testing whether or not the diagnostics entry is nil
first is enough to avoid this error:

```
  nvim_diagnostic = function()
    if vim.diagnostic == nil then
        return 0, 0, 0, 0
    end
    local diagnostics = vim.diagnostic.get(0)
    local count = { 0, 0, 0, 0 }
    ...
```

[...]

PR #524, opened 2022-01-03
fix issue #523

fidgetingbits:


fidgetingbits:
invalid per #523 comments

PR #525, opened 2022-01-03
fix: use nvim-api to avoid win_getid exceptions

kylo252:
Fixes an issue where an exception that is caused by `win_getid` can sometimes get triggered. Using the direct API call seems to fix that.


shadmansaleh (approved):

shadmansaleh:
lgtm . Thanks :)

PR #526, opened 2022-01-03
refactor: use nvim api directly for bufnr

kylo252:
Remove some overhead by using `vim.api` directly

### DIFF
--- a/lua/lualine/components/branch/git_branch.lua
+++ b/lua/lualine/components/branch/git_branch.lua
@@ -35,7 +35,7 @@ end
 
 ---updates the current value of git_branch and sets up file watch on HEAD file
 local function update_branch()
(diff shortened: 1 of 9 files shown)


lua/lualine/components/tabs/init.lua
```diff
@@ -62,7 +62,7 @@ function M:update_status()
   end
   -- mark the first, last, current, before current, after current tabpages
   -- for rendering
-  local current = vim.fn.tabpagenr()
+  local current = vim.api.nvim_get_current_tabpage()
```
shadmansaleh, line 65: Same here, tabpagenr can't be replaced with get_current_tabpage.

shadmansaleh:
lgtm . Thanks again.

PR #544, opened 2022-01-10
don't use deprecated vim.lsp.diagnostic.get_count

onsails:


shadmansaleh:
You shouldn't be using a [deprecated](https://github.com/nvim-lualine/lualine.nvim/blob/bc5c2f5658821f98e55013eb3bec9d540c420470/lua/lualine/components/diagnostics/init.lua#L63-L79) diagnostics source  . `nvim_lsp` has been deprecated in lualine use `nvim_diagnostic` instead . No need to change this . This is only here to maintain competability with 0.5

onsails:
I see, thanks

PR #554, opened 2022-01-17
feat: Add monokai theme

hermitmaster:


shadmansaleh:
https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md#getting-started

> Currently we aren't accepting regular colorscheme based themes. We think they make more sense with colorschemes as they tend not to get updated once added here.

Also I think monokai was mistakenly named molokai . Check if [this](https://github.com/nvim-lualine/lualine.nvim/blob/master/THEMES.md#molokai) is what you're looking for.

PR #556, opened 2022-01-19
fix: use one theme colors from rakr/vim-one

idanko:
Good day @shadmansaleh. In my opinion, we use wrong One Color Theme versions, especially for light one: the level of saturation is lower and colors don't match. This MR intends to use [rakr/vim-one](https://github.com/rakr/vim-one) colors range as a baseline for better user experience.

After the changes, schema will look like:
image

Which matches both: light and dark variant.
Where:
* gray1 is mono_2
* gray2 is syntax_cursor
* gray3 is visual_grey

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -189,18 +189,18 @@ them based you your `background` option.
 
 ### onedark
 <p>
-<img width='700' src='https://user-images.githubusercontent.com/41551030/108649068-e5064880-74bc-11eb-9a8f-5e3b8f401efe.png'/>
-<img width='700' src='https://user-images.githubusercontent.com/41551030/108649071-e6377580-74bc-11eb-8715-99d5aaf9eed0.png'/>
-<img width='700' src='https://user-images.githubusercontent.com/41551030/108649075-e9326600-74bc-11eb-8a94-210575aca09e.png'/>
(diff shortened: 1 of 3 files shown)


shadmansaleh:
Good day :)

If I remember correctly onedark theme is based on [onedark.vim](https://github.com/joshdick/onedark.vim) not [vim-one](https://github.com/rakr/vim-one) . 

Since vim-one is an adaptation of onedark
> Adaptation of one-light and one-dark colorschemes for Vim

Do you think that's why you're seeing the differences ?

Also if you want you can customize themes as part of your config if you don't like some colors of a theme as shown [here](https://github.com/nvim-lualine/lualine.nvim#customizing-themes) .

idanko:
@shadmansaleh But most problems not with dark variant but with the light one (from blue to bg is identical with onedark.vim), I just updated dark variant for the sake of consistency (gray1, gray2, gary3) with light variant, but don't see any impact of this (can revert dark variant changes if requested).

shadmansaleh:
TBH I have no strong opinion on either case . And this version does look better in my opinion . Can you update the screenshots in [THEMES.md](https://github.com/nvim-lualine/lualine.nvim/blob/master/THEMES.md#onelight) then we can merge.

idanko:
@shadmansaleh done

shadmansaleh:
lgtm . Thanks :)

PR #568, opened 2022-02-02
feat: add lunarvim themes for lualine

abzcoding:
Instead of keeping these inside lunarvim, i though it might be better to add them to the lualine repo itself

### DIFF
--- a/lua/lualine/themes/darkplus.lua
+++ b/lua/lualine/themes/darkplus.lua
@@ -1,0 +1,32 @@
+-- stylua: ignore
+local colors = {
+  blue = "#569cd6",
+  green = "#6a9955",
+  purple = "#c586c0",
+  red = "#d16969",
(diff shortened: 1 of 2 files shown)


shadmansaleh:
As stated in [CONTRIBUTING.md](https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md#getting-started)

> Currently we are not accepting regular colorscheme based themes. We think they make more sense with colorschemes as they tend not to get updated once added here. But if you have some unique themes idea like auto or pywal open a pr or issue.

New theme like these are no longer being added . I think leaving these in luanarvim makes sense. As you guys are maintaining the themes and you can update them there when necessary.

abzcoding:
Thank you for clarifying the situation ,my bad for not reading the contributing guide 😅

PR #574, opened 2022-02-07
feature to rename tabs

Tokubara:
with
```vim
command! -nargs=1 TabRename call TabRename("<args>")

function! TabRename(tabname)
  let t:tabname = a:tabname
  normal gTgt
endfunction
```

### DIFF
--- a/lua/lualine/components/tabs/tab.lua
+++ b/lua/lualine/components/tabs/tab.lua
@@ -14,22 +14,27 @@ end
 --- of the tab.
 ---@return string
 function Tab:label()
-  local buflist = vim.fn.tabpagebuflist(self.tabnr)
-  local winnr = vim.fn.tabpagewinnr(self.tabnr)
-  local bufnr = buflist[winnr]
-  local file = vim.api.nvim_buf_get_name(bufnr)
-  local buftype = vim.fn.getbufvar(bufnr, '&buftype')
-  if buftype == 'help' then
-    return 'help:' .. vim.fn.fnamemodify(file, ':t:r')
-  elseif buftype == 'terminal' then
-    local match = string.match(vim.split(file, ' ')[1], 'term:.*:(%a+)')
-    return match ~= nil and match or vim.fn.fnamemodify(vim.env.SHELL, ':t')
-  elseif vim.fn.isdirectory(file) == 1 then
-    return vim.fn.fnamemodify(file, ':p:.')
-  elseif file == '' then
-    return '[No Name]'
+  local custom_tabname = vim.fn.gettabvar(self.tabnr, 'tabname')
+  if custom_tabname ~= '' then
+    return custom_tabname
+  else
+    local buflist = vim.fn.tabpagebuflist(self.tabnr)
+    local winnr = vim.fn.tabpagewinnr(self.tabnr)
+    local bufnr = buflist[winnr]
+    local file = vim.api.nvim_buf_get_name(bufnr)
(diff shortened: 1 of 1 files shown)


lua/lualine/components/tabs/tab.lua
```diff
@@ -14,6 +14,10 @@ end
 --- of the tab.
 ---@return string
 function Tab:label()
+  local custom_tabname = vim.fn.gettabvar(self.tabnr, 'tabname')
```
shadmansaleh, line 17: You should use `vim.t.tabname`.

shadmansaleh:
Add the TabRename as LualineTabRename [here](https://github.com/nvim-lualine/lualine.nvim/blob/dc20cbd0a99ff1345a8186ada1fb5fb2ca3e3fdf/lua/lualine/components/tabs/init.lua#L169) . Also instead of `:noramal gTgt` you should use `:redrawtabline` .

Tokubara:
Now with `set sessionoptions+=tabpages,globals`, tabnames can be saved and restored with session files.

lua/lualine/components/tabs/tab.lua
```diff
@@ -14,6 +14,21 @@ end
...
 ---@return string
 function Tab:label()
+  local custom_tabname
+  if vim.g.lualine_tabnames then
+    custom_tabname = vim.g.lualine_tabnames[tostring(self.tabnr)]
+  else
+    vim.cmd(
+    [[
+    if(exists("g:Lualine_tabnames_str"))
+      let g:lualine_tabnames = json_decode(g:Lualine_tabnames_str)
```
shadmansaleh, line 24: I don't like this . Why are we doing this ? We should just do tabpage variables like before .

Tokubara: With this, we can save tabnames with `:mksession` and restore. `mksession` can only save strings and numbers, so I have to  maintain another variable `g:Lualine_tabnames_str`.

shadmansaleh: Note: `shada` by default will store that variable too creating a mess .

If neovim doesn't provide an way to store tabpage variables in session, open an issue there . Lualine shouldn't work workaround it in this manner . Sessions aren't concern of lualine.

lua/lualine/components/tabs/tab.lua
Tokubara: I think I should use `vim.fn.gettabvar(self.tabnr, 'tabname')`. Otherwise, all tabs would have the  same name.

shadmansaleh: My bad it should be
```lua
vim.t[self.tabnr].tabname
```

shadmansaleh (approved):

shadmansaleh:
lgtm. Thanks for taking your time to implement this :)

Tokubara (commented):
I meet a situation(quite common) in which `:LualineTabRename` will lead to a crash.

1. create 2 tabs.
2. close first tab(`:tabc`).

Then tabline get crashed.

This happen because, when closing a tab, tabnr(tabpagenr) of a tab will change(2->1). But in tab.lua, `self.tabnr` does not change. So access to `vim.t` will be out of index.

shadmansaleh:
@Tokubara fixed in 98dc2dd3a2a07c251bcbd43048157aa56f901402

dakennguyen:
Hi, would it possible to save tab's names when using `mksession`?

PR #583, opened 2022-02-20
fix Palenight theme modal color assignments

MrQuasar173:
This changes around what  color each mode indicator is in the palenight theme. These colors are more consistent with palenight/material themes of other statuslines

### DIFF
--- a/lua/lualine/themes/palenight.lua
+++ b/lua/lualine/themes/palenight.lua
@@ -23,16 +23,16 @@ local colors = {
 
 return {
   normal = {
-    a = { fg = colors.black, bg = colors.purple, gui = 'bold' },
+    a = { fg = colors.black, bg = colors.cyan, gui = 'bold' },
     b = { fg = colors.purple, bg = colors.menu_grey },
     c = { fg = colors.comment_grey, bg = colors.black },
   },
   insert = {
-    a = { fg = colors.black, bg = colors.blue, gui = 'bold' },
+    a = { fg = colors.black, bg = colors.purple, gui = 'bold' },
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Are you sure ? As far as I see according to their repos [themes](https://github.com/drewtempelmeyer/palenight.vim/blob/master/autoload/lightline/colorscheme/palenight.vim) & [screenshots](https://github.com/drewtempelmeyer/palenight.vim/blob/master/images/screenshot.png) current colors are correct .

What are you using as reference to say it's wrong ?

MrQuasar173:
Palenight is a sub-theme of Material. The main Material theme for vim is [material.vim](https://github.com/kaicataldo/material.vim) and for lightline, airline, and other vimscript statuslines it uses cyan as normal mode,
purple as insert mode, green as replace mode, etc. I am not changing any of the actual color hexes. 

Here is a screenshot from material.vim's gh: [screenshot](https://raw.githubusercontent.com/kaicataldo/material.vim/main/screenshots/material-all-variants.png)

MrQuasar173:
(in the screenshot, the window behind the front one is palenight)

shadmansaleh:
lualines palenight them was submitted  based on  [palenight.vim](https://github.com/drewtempelmeyer/palenight.vim) not [material.vim](https://github.com/kaicataldo/material.vim) . Since it's correct according to that I will not be accepting this change. I don't see any reason to prefer on repos interpretation over other.

We let users to customize themes as part of their config (like shown [here](https://github.com/nvim-lualine/lualine.nvim#customizing-themes)) to avoid this kind of debates . You can try that :)

PR #584, opened 2022-02-20
Ayu adaptive background

mizlan:
Hi! I use Ayu and noticed this theme did not use the adaptive `background` setting. This was presumably because Ayu has 2 dark themes. But I have accounted for that here. I tested this and it works and is much better for light/dark switching :)

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -40,8 +40,11 @@ it'll try it's best to generate one.
 </p>
 
 ### ayu
-It's a combination of ayu_light, ayu_dark & ayu_mirage. It loads one of these
-them based you your `g:ayucolor` option.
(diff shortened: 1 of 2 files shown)


mizlan:
Hold on, this is bugged for some reason.

shadmansaleh:
LGTM. Thanks for your contribution .

shadmansaleh:
> Also, could there be an option to disable the "Use the first one that isn't in lualines repo" method? i.e., to force lualine's own theme instead

Not really. That doesn't make sense to do . Current priority is user_config > colorschemes > lualine .
 You can pull the contents of your theme  in your lusline config & pass it the table in theme option . Or copy the theme to your `~/.config/nvim/lua/lualine/themes` and lualine will load the themes present there instead of one in the colorscheme.

PR #592, opened 2022-03-03
feat: add floaterm extension

dcordb:
This PR adds a [vim-floaterm](https://github.com/voldikss/vim-floaterm) extension.

### DIFF
--- a/lua/lualine/extensions/floaterm.lua
+++ b/lua/lualine/extensions/floaterm.lua
@@ -1,0 +1,33 @@
+-- MIT license, see LICENSE for more details.
+-- Extension for floaterm.nvim
+
+-- based from here https://github.com/voldikss/vim-floaterm/issues/224#issuecomment-755233112
+local vim_func =
+[[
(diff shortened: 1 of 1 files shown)


lua/lualine/extensions/floaterm.lua
```diff
@@ -0,0 +1,33 @@
...
+      return printf('floaterm %s/%s', idx, cnt)
+  endif
+
+  return printf('floaterm (%s) %s/%s', name, idx, cnt)
+endfunction
+
+echo FloatermStatus()
+]]
+
+local function floaterm_statusline() return vim.api.nvim_exec(vim_func, true) end
```
shadmansaleh, line 25: I don't see any reason why this can't directly be done in a lua function.
Creating a viml function every time statusline updates isn't a good idea.

dcordb: Ok, I agree, but I need to learn some lua (and viml -> lua) to transform the FloatermStatus function.

PR #594, opened 2022-03-04
feat(diagnostics): add padding when using different background colors

camilledejoye:
When each "sub-section" of the diagnostic components have a different background the rendering is off when
multiple more than one of them is shown.

Before:
image

After:
image

I didn't want to add any new configuration so I decided to with detecting different background colors since it's this change of color which requires the extra padding.

### DIFF
--- a/lua/lualine/components/diagnostics/init.lua
+++ b/lua/lualine/components/diagnostics/init.lua
@@ -104,13 +104,17 @@ function M:update_status()
 
   -- format the counts with symbols and highlights
   if self.options.colored then
-    local colors = {}
+    local colors, bgs = {}, {}
     for name, hl in pairs(self.highlight_groups) do
       colors[name] = modules.highlight.component_format_highlight(hl)
+      bgs[name] = modules.utils.extract_highlight_colors(colors[name]:match('%%#(.-)#'), 'bg')
     end
+    local previous_section, padding
     for _, section in ipairs(self.options.sections) do
       if diagnostics_count[section] ~= nil and (always_visible or diagnostics_count[section] > 0) then
-        table.insert(result, colors[section] .. self.symbols[section] .. diagnostics_count[section])
+        padding = previous_section and (bgs[previous_section] ~= bgs[section]) and ' ' or ''
+        previous_section = section
+        table.insert(result, colors[section] .. padding .. self.symbols[section] .. diagnostics_count[section])
       end
     end
   else
--- a/lua/tests/spec/lualine_spec.lua
+++ b/lua/tests/spec/lualine_spec.lua
@@ -752,4 +752,103 @@ describe('Lualine', function()
       end)
     end)
   end)
+
+  describe('diagnostics', function()
+    local diagnostics_conf = vim.deepcopy(config)
+    diagnostics_conf.sections = {
+      lualine_a = { {
+        'diagnostics',
+        symbols = { error = 'E:', warn = 'W:', info = 'I:', hint = 'H:' },
+        diagnostics_color = {
+          error = { bg = "#a89984", fg = "#ff0000" },
+          warn = { bg = "#a89984", fg = "#ffa500" },
+          info = { bg = "#a89984", fg = "#add8e6" },
+          hint = { bg = "#a89984", fg = "#d3d3d3" },
+        },
+        sources = { function() return {} end },
+      } },
+      lualine_b = {},
+      lualine_c = {},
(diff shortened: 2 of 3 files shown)


lua/lualine/components/diagnostics/init.lua
```diff
@@ -104,13 +104,18 @@ function M:update_status()
 
   -- format the counts with symbols and highlights
   if self.options.colored then
-    local colors = {}
+    local colors, bgs = {}, {}
     for name, hl in pairs(self.highlight_groups) do
       colors[name] = modules.highlight.component_format_highlight(hl)
+      bgs[name] = self.options.diagnostics_color[name].bg or nil
```
shadmansaleh, line 110: color options can be string & function too. In that case it'll through error . Get the bg from the hl_group_name in colors[name] with extract_highlight_color function from utils module .

lua/lualine/components/diagnostics/init.lua
```diff
@@ -104,13 +104,18 @@ function M:update_status()
...
     end
+    local padding
+    local previous_section = nil
```
shadmansaleh, line 113: You can declare both padding & previous section in same line.

shadmansaleh:
I think similar change can also be added to `diff` component

lua/lualine/components/diagnostics/init.lua
camilledejoye: luacheck didn't "liked" it:
```lua
local previous_section, padding = nil, nil -- Complains about the value never being used since padding is redefined
local previous_section, padding = nil -- Complains that a value is missing and nil will be used instead
```

camilledejoye:
> I think similar change can also be added to `diff` component

I wouldn't know, I never used to it.

lua/lualine/components/diagnostics/init.lua
shadmansaleh: If you don't assign it'll be set to nil.

You just need .
```lua
local previous_section, padding
```

lua/lualine/components/diagnostics/init.lua
```diff
@@ -104,13 +104,18 @@ function M:update_status()
 
   -- format the counts with symbols and highlights
   if self.options.colored then
-    local colors = {}
+    local colors, bgs = {}, {}
     for name, hl in pairs(self.highlight_groups) do
       colors[name] = modules.highlight.component_format_highlight(hl)
+      bgs[name] = modules.utils.extract_highlight_colors(hl.name, 'bg') or nil
```
shadmansaleh, line 110: this is incorrect . hl is actually only meant to be consumed by `component_format_highlight` . It knows how to handle all cases regarding the highlight . hl.name is actually internal representation and isn't always same as the actual neovims highlight group name (because in some cases we need to create several highlight groups for different modes from same lualines hl)

It should be something like this
```lua
bgs[name] = modules.utils.extract_highlight_colors(colors[name]:match('%%#(.-)#', 'bg')
```

lua/lualine/components/diagnostics/init.lua
camilledejoye: Good to know ;)

shadmansaleh:
Can you rebase it on master and add some tests . With the new statusline module writing tests shouldn't be hard .

camilledejoye:
I tried my best but I'm not that comfortable with lua testing.
I didn't want to have to test the full feature neither so I just added the most common cases.

shadmansaleh:
> I tried my best but I'm not that comfortable with lua testing.

You can always ask for help if something confuses you :)

> I didn't want to have to test the full feature neither so I just added the most common cases.

These tests looks good to me . We don't have 100% coverage anyway . Slowly getting there though I think were somewhere between 75-80% . Adding tests for the feature you're adding is mostly good enough . Prevously testing lualine was a bit clunky so I didn't really enforce test requirements with prs . With statusline module I think most of that clunky ness is now solved so I'm thinking of having tests as requirement for prs .


Define the diagnostics_color option to specify highlights instead of relying on defaults . vim.diagnostics & related hl groups were added in nvim-0.6 . Because lualine prefers them as default colors in 0.5 aren't matching .

shadmansaleh (approved):

shadmansaleh:
LGTM. Thanks for taking the time to improve lualine :)

camilledejoye:
> LGTM. Thanks for taking the time to improve lualine :)

Thanks for taking the time to write it in the first place ;)

PR #595, opened 2022-03-05
feat: add windows component

camilledejoye:
This is more an attempt of improvement than a feature.
I reused everything from the `buffers` component to only list in the tab bar the buffers shown in a window inside the current tab.

I think it could be a good improvement to the builtin components, allowing users which want to leverage tabs to organize their buffers.
While it makes the `buffers` component easier to extend without having to copy/paste it locally it does not break anything and would be entirely up to the user to choose between the two.


camilledejoye:
Good point, I don't show the numbers so I never realized...
I'll check where it's done and see how to override it.

I also realized that we build a buffer in `buffers.lua` at some point but it will never provide the `winnr`...

There is definitely need for improvements, since you seems to agree with the concept I'll try to find the time to take a closer look at it and write the necessary documentation.

camilledejoye:
I just pushed the changes you required plus some additions that were missing.

I first developed this "feature" for my personal use and since my mouse is disabled I didn't thought about it at first.
If you spot any other thing I missed let me know.

I also improved `Buffer:name()` to handle quickfix and loclist since we show them when focused (both for `buffers` and `windows`).

I didn't updated the test yet, I wanted to at list test the windows number, because there is somme errors.

README.md
```diff
@@ -228,6 +228,7 @@ sections = {lualine_a = {'mode'}}
 
 - `branch` (git branch)
 - `buffers` (shows currently available buffers)
+- `windows` (shows currently available windows)
```
shadmansaleh, line 231: this list is actually sorted by name . so windows should be at bottom.

README.md
```diff
@@ -456,6 +457,43 @@ sections = {
 }
 ```
 
+#### windows component options
```
shadmansaleh, line 460: Same as above it's sorted by name.

lua/lualine.lua
```diff
@@ -271,14 +271,14 @@ local function set_statusline()
...
-    vim.go.statusline = nil
+    vim.go.statusline = ''
     vim.go.laststatus = 2
   end
 end
 
 -- lualine.statusline function
 --- Draw correct statusline for current winwow
----@param focused boolean : force the vale of is_focuased . useful for debugginf
+---@param focused boolean : force the vale of is_focuased . useful for debugging
```
shadmansaleh, line 281: Nitpick can fix the typo in value too

lua/tests/spec/lualine_spec.lua
```diff
@@ -751,6 +751,55 @@ describe('Lualine', function()
...
+        }
+        |{1: 1 d.txt }
+        {2:}
+        {3:                                                                                                              }|
+        ]===])
+
+        vim.cmd('tabprev')
+        tabline:expect([===[
+        highlights = {
+            1: lualine_a_buffers_active = { bg = "#a89984", bold = true, fg = "#282828" }
```
shadmansaleh, line 779: Windows component needs to use it's own highlight groups not the ones from buffers otherwise color option from windows and buffers components will effect each other.

shadmansaleh (changes requested):

shadmansaleh:
Sorry for the late reply. This mostly looks good. I've found few issues specially windows component using same highlight group as buffers component is a major one.

> lot of tests from lualine_spec.lua are not executed.

> It seems that embedding the describe in each other ended up masking the errors.

That's interesting . More reasons to switch to vusted for testing.

> The error come from the set_tabline and set_statusline function in lua/lualine.lua, because nil is assigned to vim.go.tabline and vim.go.statusline respectively.
nil is not permitted, it should be an empty string instead.

As far as I'm aware in this case nil and empty string is same so we should be good.

shadmansaleh:
I've applied the test fix in master as it was affecting other prs too . You can remove them from here.

camilledejoye:
> I've applied the test fix on master as it was affecting other prs too . You can remove them from here.

Great, I'll do it but probably not before Sunday :)

shadmansaleh:
> Great, I'll do it but ~probably not before Sunday~ :)
Actually, I found the time tonight.

Don't worry there's no hurry .

Also try to avoid force pushing as much as possible . It's harder to review if you force push as I need to look through prs entire diff again instead of just what you've changed after last review . Also In the end the pr will get squashed anyway so don't worry about having too many commits.

camilledejoye:
Not sure I really had any choice to force the push anyway since I needed to remove the first commit which was the quick fix for the specs and then rebase from master.
Let me know if there a way, I would now for next time :)

shadmansaleh:
> I needed to remove the first commit which was the quick fix for the specs and then rebase from master.

You can revert the commit instead of removing it. Instead of rebasing you can merge master to this branch to resolve conflicts or to pull changes from master . Both of them doesn't disrupt the history so no force push required.

I'll try to take a proper look at this soon.

One thing I haven't tested though It feels like calling `self.super.init()` from windows components init will try to reset buffers components colors (set by options) to their default values . Can you check if that's the case ? Maybe we are to do `self.super.super.init()` here 🤷‍♂️

camilledejoye:
> One thing I haven't tested though It feels like calling `self.super.init()` from windows components init will try to reset buffers components colors (set by buffers_color options) to their default values . Can you check if that's the case ? Maybe we are to do `self.super.super.init()` here

I can try to have a look after my work day.
I thought the changes was only applied to `self` anyway so it should not be same `buffers` component which is used by the `windows` than the one declared somewhere else.
Ultimately I don't see use cases to use both `windows` and `buffers` at the same time anyway but it would be best if there is no side effects.

camilledejoye:
On the left it's the `windows` component with default options: active is blue and inactive is same color as the tabline with my theme.
On the right it's the `buffers` component defined with (my awesome design skills):
```lua
{ 'buffers', mode = 2, buffers_color = {
  active = { fg = '#000000', bg = '#00ff00' }, -- red bg
  inactive = { fg = '#000000', bg = '#ff0000' }, -- green bg
} }
```
The result being:
image
I guess this confirms that there is no override of the buffers' highlights by the windows component.

shadmansaleh:
I've made some changes to this branch & resolved the conflicts . Take a look at them and remember to pull before pushing stuff.

windows component was creating extra buffer highlights and if you just set active without setting inactive in windows_color option it was throwing nil errors. Also windows component was respecting buffers_color option.

lua/lualine/components/windows/init.lua
```diff
@@ -0,0 +1,75 @@
+local Window = require('lualine.components.windows.window')
+local M = require('lualine.components.buffers'):extend()
+local highlight = require('lualine.highlight')
+
+local default_options = {
+  disabled_filetypes = {},
```
shadmansaleh, line 6: Is specifying these two here necessary ? They do get the same value from buffers component like other unspecified option defaults.

camilledejoye: I think we can remove them, I probably just added them because I was not sure about where those options are defined.
Since I needed them I declared them just to be safe, if they are always present then it's indeed unnecessary.

shadmansaleh: Ohh these are new options you added in windows . I mistakenly thought these were present in buffers component too . It's fine these are suppose to be here.

shadmansaleh:
LGTM. Thanks again .

camilledejoye:
My pleasure, it allows me to cleanup my dotfiles :laughing: 

Thanks for all your hard work!

PR #597, opened 2022-03-05
feat(themes): add automatic theme for nvim-base16 colorschemes

camilledejoye:
[base16](https://github.com/chriskempson/base16) is a huge repository of theme's templates which, as the name lead to think, is based on 16 colors.
Each of these templates is then used to generate themes for a lot of different tools, allowing to have a common colorscheme across your all system.

This is a proposal to add a dedicated theme which automatically defines the colors to use in lualine based on the colorscheme loaded by [RRethy/nvim-base16](https://github.com/RRethy/nvim-base16).
Which is the plugin allowing to integrate those themes inside Neovim, at least one the existing plugins. It seems to be the more maintained and used/stars.

Here are the preview I have added in the `THEMES.md` file:

image
image
image
image


shadmansaleh:
Adding a base16 based theme sounds like a good idea . I'll try to take a look at this tomorrow .

lua/lualine/themes/base16.lua
```diff
@@ -0,0 +1,92 @@
...
+  if not loaded then
+    add_notice(
+      'nvim-base16 is not currently present in your runtimepath, make sure it is properly installed,'
+      .. ' fallback to default colors.'
+    )
+
+    return nil
+  end
+
+  if not base16.colors and not vim.env.BASE16_THEME then
```
shadmansaleh, line 65: What sets `BASE16_THEME` ?

camilledejoye: I honestly have no idea, I found it when looking into nvim-base16 to check how to retrieves the colors: https://github.com/RRethy/nvim-base16/blob/6c6962d51dda76cf9bc1467aa8b09036396985d0/lua/base16-colorscheme.lua#L110

jacktuck:
I've been using this for a few weeks and tried a few base16 themes. No issues so far.

PR #617, opened 2022-03-18
fix(filename): use columns for shortening if using global statusline

fitrh:
After https://github.com/nvim-lualine/lualine.nvim/pull/613 , I think shortening target calculation should be based on `columns` option if using `globalstatus`, otherwise, `filename` shortening occurs when we open a new split

> Enabling global statusline, shortening based on window width

20220318_203631_full

---

> Global statusline, shortening based on `columns` option

20220318_203724_full


shadmansaleh (approved):

shadmansaleh:
LGTM . Thanks :)

PR #618, opened 2022-03-19
feat: add workspace diagnostics source

ybbond:
this method is from my own config to add workspace diagnostic to the component table.


shadmansaleh:
What do you mean by workspace diagnostics and how is it different from current nvim_diagnostic source?

fitrh:
@shadmansaleh `nvim_diagnostic` returns the diagnostics for current buffer (document diagnostics), passing `nil` as `bufnr` for `vim.diagnostic.get` will return diagnostics of all buffers (workspace diagnostics)

shadmansaleh:
I'm not sure if nvim lsp has a notation of workspaces yet. If all buffers mean all open buffers from different projects that would kind of annoying right ?

AnonymusRaccoon:
Passing nil to the bufnr does return the diagnostic for every file in the workspace, not opened ones only. This is also what is used in [trouble.nvim](https://github.com/folke/trouble.nvim/blob/main/lua/trouble/providers/diagnostic.lua#L10).

AnonymusRaccoon:
@shadmansaleh Is there a reason this hasn't been merged yet?

changyiyao:
> I'm not sure if nvim lsp has a notation of workspaces yet. Also even if nvim lsp had workspaces unlikely diagnostics module does. If all buffers mean all open buffers from different projects that wouldn't be very useful right ?

See this PR: https://github.com/nvim-lualine/lualine.nvim/pull/759.
nvim.diagnostic.get() without bufnr param will get workspace dianogstics from LSP server no matter whether the buffer is opened.

shadmansaleh:
Sorry I kind of forgot about this pr .I think we can merge it.

PR #620, opened 2022-03-22
Evil: Display up to 3 lsp names

jesstelford:
When multiple LSP clients are active on a buffer, display the first 3 names in full.

When more than 3 LSP clients are active, display a "+X" suffix (where "X" is the number of additional clients not displayed).

**4 LSP clients** _(new)_
Screenshot from 2022-03-22 16-29-49

**1 LSP client** _(existing behaviour)_
Screenshot from 2022-03-22 16-30-01

**No LSP clients** _(existing behaviour)_
Screenshot from 2022-03-22 16-30-44

### DIFF
--- a/examples/evil_lualine.lua
+++ b/examples/evil_lualine.lua
@@ -158,17 +158,29 @@ ins_left {
 ins_left {
   -- Lsp server name .
   function()
-    local msg = 'No Active Lsp'
     local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
     local clients = vim.lsp.get_active_clients()
     if next(clients) == nil then
-      return msg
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Evilline is just an example config . Users are meant to follow it's pattern and adapt it to their needs as you've done. This doesn't need to be in lualine it can be in your config . If you want you can share it in [wiki](https://github.com/nvim-lualine/lualine.nvim/wiki/Component-snippets) if you want to.

jesstelford:
Yeah, no worries 👍

I figured this might turn a slightly better default for those who just copy+paste.

Thanks for all the excellent work ♥️

PR #621, opened 2022-03-22
fix: don't add tests module to runtimepath

kylo252:
The `tests` folder was getting added automatically to `&rtp` since it was under `lua/`. This is problematic for other plugins that would like to also perform tests while lualine is loaded. 

Solution: move `tests` folder outside of `lua/` and explicitly tell Neovim to add it as a runtimepath.


kylo252:
To test the effect, try this before and after:
```lua
print(vim.inspect(require "tests.helpers"))
```

shadmansaleh:
As far as I remember we moved `tests/` in to `lua/` because hoob3rt was having trouble running locally on his machine with `tests/`  folder outside of `lua/` . We kind of just followed location telescopes test dir. Can you tell me why a plugin will run it's tests with other plugins loaded ?

kylo252:
> As far as I remember we moved tests/ in to `lua/` because hoob3rt was having trouble running locally on his machine with `tests/` folder outside of `lua/`

that's why I added that test-runner wrapper

> We kind of just followed location telescopes test dir. 

what I'm suggesting is inline with "core" plugins, see: https://github.com/nvim-treesitter/nvim-treesitter/tree/master/tests and https://github.com/neovim/nvim-lspconfig/tree/master/test

> Can you tell me why a plugin will run it's tests with other plugins loaded ?

in this case it's Lunarvim, but this is still polluting the global namespace regardless. I ended up doing this to fix the issue: [LunarVim/LunarVim@`94ed25c` (#2351)](https://github.com/LunarVim/LunarVim/pull/2351/commits/94ed25cacedb81c7f31cbb8ffa742a6aedc6133c)

shadmansaleh:
Ok . That's reasonable . Since all tests are passing I think we can merge it.

Issue #603, opened 2022-03-10
Feat: Enumerate buffer tabs in tabline


nkakouros:
### Requested feature

Similar to other buffer line extensions, it would be useful to have custom numbering for buffers in the buffer line, instead of using the internal buffer number of nvim. 

### Motivation

The buffer number increments for the life time of a vim session. This could lead to large numbers if the session is long lived. This makes it cumbersome to select the desired buffer and only really saves an `:ls` for the user. It also makes it hard to jump to the first buffer in the buffer line as you need to know the buffer number. If buffer tabs where numbered sequentially, the first tab would be `1` always.

I understand that this request is more complex than it appears, since in order to make use of these sequential numbers custom mappings will need to be provided by the plugin to select a specific buffer which will take care of the translation between sequential number and internal buffer number. But, it would greatly improve the usability.

shadmansaleh:
This is something we can definitely add. Any one interested in making a pr with this ?

Also can you elaborate on what you mean by custom mappings ? I think providing a vim command like we currently provide for naming tabs (LualineRenameTab) would be sufficient.

shadmansaleh:
> The tab selection mechanism will also need to be implemented

Not sure what you mean by tab selection ??

nkakouros:
The equivalent of `:buffer X` but using the custom numbering (instead of the internal buffer number).

PR #628, opened 2022-03-28
feat: show index in buffers component instead of bufnr

jqhr:
**Features:**
 Neovim's bufnr sometimes may very strange when it's number not as same as the order of buffers component. So I add buffer index, which index is as same as buffers component.
图片
In this picture, you can see the number of buffer component are different from bufnr.
I wont delete code about bufnr, I just add buffer index.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,8 +437,8 @@ sections = {
       show_modified_status = true, -- Shows indicator when the buffer is modified.
 
       mode = 0, -- 0: Shows buffer name
-                -- 1: Shows buffer index (bufnr)
-                -- 2: Shows buffer name + buffer index (bufnr)
+                -- 1: Shows buffer index
+                -- 2: Shows buffer name + buffer index
 
       max_length = vim.o.columns * 2 / 3, -- Maximum width of buffers component,
                                           -- it can also be a function that returns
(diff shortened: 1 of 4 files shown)


jqhr:
I think that feature was very useful.For example, add map for buffers. such as 
```viml
" buffer
function UpdateBufMap() 
  " add keymap
  let bufs = getbufinfo({ 'buflisted': v:true})
  for item in bufs
    let i = index(bufs, item)
    exec 'nmap <A-' . ( i + 1 ) . '> :b' . item.bufnr . '<CR>'
  endfor
endfunction
" 更新keymap
augroup update_buf_map
  autocmd!
  autocmd BufRead,BufDelete,BufNewFile * call UpdateBufMap() 
augroup END

```
That was very easy to add keymap for Alt prefix keys. Just like  switch tabs in chrome or firefox.

shadmansaleh:
Thanks . We were talking about this in #603. I think this should be the default behavior and we should expose a simple way to jump to specific buffer from index in buffers component. 

I've added them to this pr . Take a look if you have any objections.

Now we just need some tests Can you add some testing this feature in lualine_spec with other buffers tests ? Should be fairly easy with the statusline assertion we have.

PR #629, opened 2022-03-29
feat: hide filename extension for buffers component

mudox:
Add an option `hide_filename_extension` to `buffers` components.
The file kind icon already indicates filetype. Hidng extensions can save more spaces.

CleanShot 2022-03-29 at 18 31 47@2x


shadmansaleh:
Add a test for this in lualine_spec . Just follow the other buffers test there should be easy to add.

Issue #632, opened 2022-03-30
Warning about config issues recommends running :LualineNotices, but that spawns an empty window


Popkornium18:
Lualine warns about config issues during startup, but the recommended `:LualineNotices` only spawns an empty window.

I'm using Neovim 0.6.1 from the Arch Linux repos.

This is my lualine config, though I'm not looking for support with my config, this issue is about the empty `:LualineNotices` window.

```
local function getLspClient(msg)
  msg = msg or ' No Active Lsp'
  local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
  local clients = vim.lsp.get_active_clients()
  if next(clients) == nil then
    return msg
  end

  for _, client in ipairs(clients) do
    local filetypes = client.config.filetypes
    if filetypes and vim.fn.index(filetypes, buf_ft) ~= -1 then
      return '  ' .. client.name
    end
  end
  return msg
end

require('lualine').setup {
  options = {
    theme = 'onedark'
  },
  extensions = { 'quickfix' },
  sections = {
    lualine_a = { 'mode' },
    lualine_b = { 'branch' },
    lualine_c = {
      {
        'diff',
        colored = true,
        symbols = { added = '+ ', modified = '~ ', removed = '- ' },
      },
      {
        'diagnostics',
        sources = { 'nvim_diagnostic' },
        sections = { 'error', 'warn', 'info' },
        symbols = { error = ' ', warn = ' ', info = ' ' },
      },
      { 'filename', symbols = { modified = '  ', readonly = '  ' } },
    },
    lualine_y = { 'encoding', 'fileformat', { 'filetype', colored = true } },
    lualine_x = { getLspClient },
    lualine_z = { 'location' },
  }
}
```

I can provide further information if needed

tetzng:
I have the same issue.
When I run `:LualineNotices` I get an empty window.

```lua
require('lualine').setup {
  options = {
    icons_enabled = true,
    theme = 'auto',
    component_separators = { left = '', right = ''},
    section_separators = { left = '', right = ''},
    disabled_filetypes = {},
    always_divide_middle = true,
    globalstatus = true,
  },
  sections = {
    lualine_a = {'mode'},
    lualine_b = {'branch', 'diff', 'diagnostics'},
    lualine_c = {'filename'},
    lualine_x = {'encoding', 'fileformat', 'filetype'},
    lualine_y = {'progress'},
    lualine_z = {'location'}
  },
  inactive_sections = {
[...]

PR #635, opened 2022-04-02
Document workaround for LualineNotices and empty windows

jetm:
Lualine warns about config issues during startup, but the recommended
`:LualineNotices` only spawns an empty window.

The empty window is related to loading the colorscheme after calling
lualine setup. Load lualine after the colorscheme:

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,22 @@ use {
 
 You'll also need to have a patched font if you want icons.
 
+### LualineNotices and empty windows
+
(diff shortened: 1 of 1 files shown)


jetm:
#632

shadmansaleh:
This is not what I meant also miss leading as other things can cause the same issue too . I think I have an idea to fix it will try that later.

TODO(shadman) Make only last call to setup trigger LualineaNotice message to trigger by differing message  showing to couple of seconds after nvim startup.

PR #637, opened 2022-04-03
Adding hybrid color scheme

jonathf:
Great plugin.

I'd love for my favorite theme to be added, if you don't mind.

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -106,6 +106,15 @@ them based you your `background` option.
 <img width='700' src='https://user-images.githubusercontent.com/69089176/118902070-c2919900-b90c-11eb-9296-145a99ce52ce.png'/>
 </p>
 
+### hybrid
+<p>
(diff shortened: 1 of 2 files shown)


shadmansaleh:
Sorry we are currently not accepting new colorscheme based themes as stated in https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md

> Currently we aren't adding regular colorscheme based themes. We think they make more sense with colorschemes as they tend not to get updated once added here. But if you have some unique themes idea like [auto](https://github.com/nvim-lualine/lualine.nvim/blob/master/THEMES.md#auto) or [pywal](https://github.com/nvim-lualine/lualine.nvim/blob/master/THEMES.md#pywal) feel free to open an pr or issue.

You can open a pr in the colorschemes repo to see if they want to support lualine or not. Or use the theme locally as part of your config by placing it in ~/.config/nvim/lua/lualine/themes/ folder.

PR #649, opened 2022-04-13
feat: live filesize

dmun:
Hi! I changed the filesize component a bit so the filesize is always up-to-date by using vim.fn.wordcount().bytes. I removed the decimal for the bytes, since I don't think you can have half a byte, and I fixed a typo. Let me know what you think!

See preview:
[asciicast](https://asciinema.org/a/qyrJ9MVTdf7hPVh7qpFEcVo1S)

### DIFF
--- a/lua/lualine/components/filesize.lua
+++ b/lua/lualine/components/filesize.lua
@@ -1,15 +1,7 @@
 -- Copyright (c) 2020-2021 shadmansaleh
 -- MIT license, see LICENSE for more details.
 local function filesize()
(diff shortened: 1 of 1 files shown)


shadmansaleh:
lgtm . Thanks .

dmun:
> This may have a performance problem in very large file (60MB), it makes cursor move slow.

You are right! I should've known that it would become kinda laggy with big files, I created a new [PR](https://github.com/nvim-lualine/lualine.nvim/pull/675) to fix this.

PR #651, opened 2022-04-15
Adds option to start counting buffers from custom index

nkakouros:
This PR adds an option to the buffers component to start counting the buffers from a custom index. This is inspired by airline. This is useful for instance if the user wants to create mappings like `<leader>BUF_INDEX`. As it stands now, the user has to create mappings for indices starting from 1 as the counting starts from there. But, a mapping of `<leader>1` and one of `<leader>10` will "confilct" and there will be a delay in choosing buffer #1. This is quite unbearable as the numbers 1-9 are the ones that will always be available and there will always be a delay when switching buffers with such mappings unless a buffer with index >=10 is to be selected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,6 +441,7 @@ sections = {
                 -- 1: Shows buffer index
                 -- 2: Shows buffer name + buffer index
 
+      first_index = 1,  -- Start counting buffer indices from this number
       max_length = vim.o.columns * 2 / 3, -- Maximum width of buffers component,
                                           -- it can also be a function that returns
                                           -- the value of `max_length` dynamically.
--- a/lua/lualine/components/buffers/init.lua
+++ b/lua/lualine/components/buffers/init.lua
@@ -10,6 +10,7 @@ local default_options = {
   hide_filename_extension = false,
   show_modified_status = true,
   mode = 0,
+  first_index = 1,
   max_length = 0,
   filetype_names = {
     TelescopePrompt = 'Telescope',
@@ -48,7 +49,11 @@ function M:init(options)
     active = get_hl('lualine_' .. options.self.section, true),
     inactive = get_hl('lualine_' .. options.self.section, false),
   }
+
   self.options = vim.tbl_deep_extend('keep', self.options or {}, default_options)
+  self.options.index_offset = self.options.first_index - 1
+  M.options = self.options
+
   if self.options.component_name == 'buffers' then
     self.highlights = {
       active = highlight.create_component_highlight_group(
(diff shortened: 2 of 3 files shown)


shadmansaleh:
I'm still not sure of the usecase here. We aren't creating the mappings rather users are suppose to create mappings themselves they can just add 10 if they want to while creating there mappings . I don't think it makes sense to start indexing from anything other then 1.

nkakouros:
This was inspired by airline-tabline:

https://github.com/vim-airline/vim-airline/blob/be5bda1f1dbfa4e36b2adabaf3f423adfa66c336/doc/airline.txt#L1194-L1202

Agreed, they expose mappings. Lualine does not expose mappings, just a command (`LualineBuffersJump`). 

But, there is the case when the users want to create similar mappings, potentially because they are coming from airline, like me. In this case, they will be able to create mappings like `<leader>BUFINDEX`. They will for sure create mappings for buffer indices 1-9 because these are the first numbers that will be used by as buffer indices. It is not a stretch to assume that they will have more than 9 buffers open and showing in the buffer line. In this case, they will also create mappings for indices 10-99 (or more). In this case, mappings like `<leader>1` and `<leader>10` will conflict. Neovim will then wait for a timeout whenever the user presses `<leader>1` to select the first tab. This will happen even if there is no buffer with index 10 or higher. This is bad user experience.

My PR allows setting the starting index number to an arbitrary number. I was thinking that the user may want to use index numbers 10-99, but they may also want to use numbers 100-999. Of course, the PR could do what ailrine-tabline does, offer two alternatives, count between 0-9 or count between 10-99. 

A counter-argument can be that lualine does not offer mappings so this is none of its problem. But, the way the buffer indices are implemented (always starting counting from 1) can create annoyances to users or break their workflow.

Another counter argument is what you say, they can just create mappings like:

```
map <leader>43 :LualineBuffersJump34
" or
map <leader>21:LualineBuffersJump11
```

but I believe you can agree that this is very confusing and requires mental work from the user.

shadmansaleh:
Making lualine mimic airline isn't our goal.
A simple way to avoid the wait time with \<leader>1 would be to press \<space> or any unused key afterwords.

Besides I'd much prefer to let users easily use 1\<leader> in place of \<leader>1 rather than this .

I don't think anyone is using 999 even 99 buffers at the same time. If anyone is at that point a bufferline is absolutely useless for buffer navigation they should be looking at fuzzy finders.

nkakouros:
I will maintain my fork with this functionality.

shadmansaleh:
Sure you can always do that . Also pretty much all of lualine is overridable in runtime so . You could just have a custom component extending builtin buffers component as part of your config.

nkakouros:
That's an interesting approach I didn't know about! Could you spare some minutes to share a skeleton of what that would look like?

shadmansaleh:
You can take a look at windows component . It basically extends buffers component and just tweaks stuff that it needs to change.

PR #659, opened 2022-04-21
fix: Check minor version instead of static string

thraizz:
This pull request replaces the static check against "nvim-0.5" with a more-dynamic
check against the minor version.
Currently, the check fails, as nvim 0.7 was released a few days ago.

### DIFF
--- a/lua/lualine/components/diagnostics/sources.lua
+++ b/lua/lualine/components/diagnostics/sources.lua
@@ -6,7 +6,7 @@ local M = {}
 M.sources = {
   nvim_lsp = function()
     local error_count, warning_count, info_count, hint_count
-    if vim.fn.has('nvim-0.6') == 1 then
+    if 5 < tonumber(vim.version().minor) then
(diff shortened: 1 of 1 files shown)


shadmansaleh:
As clason has stated there's no reason for current check to fail. Can you say why you think so ?

Instead of thinking has('nvim-0.6') as whether user is using nvim version 0.6 think of it as whether user has features from nvim 0.6 which is true for all versions from 0.6 .

The minor version check you're doing will actually be broken if neovim realeses 1.0 while the current has check will keep working.

There are actually several checks like this in the codebase. And I'm currently strongly considering dropping support for nvim-0.5 & 0.6 .

thraizz:
Like many other issues with this  deprecation warning before, the real culprit was another plugin, nvim-lua/lsp_extensions.nvim.

I just happened to find these lines first and after editing them, the warning disappeared- because I had no LSP server for lua installed, so the extension was never used. I'll see myself out. 

Thanks for the great work and fast replies, have a good weekend.

PR #662, opened 2022-04-24
Use escape sequence instead of raw binaries

HE7086:
Raw binary characters confuse some programs like `file`, which thinks the file is `data` instead of `ASCII text`. This in turn messed up other programs that depending on file. Replace them with escape sequence solve the issue without any impact on functionality, and they can be correctly rendered on GitHub.


shadmansaleh:
Sounds like a good idea . Thanks.

PR #663, opened 2022-04-24
fix: make escape sequences compatible with Lua 5.1

zeertzjq:
Escape sequences starting with `\x` are supported by Lua 5.2 and LuaJIT, but not by Lua 5.1


shadmansaleh:
Wasn't aware of that . Thanks.

PR #666, opened 2022-04-26
feat: add option to invert component icon alignment

daephx:
I was about to open an issue for this but figured it seemed easy enough.
Here is the issue template I was going to provide.

Set: `icon_right = true`

### Requested feature

Allow component icons to be displayed on both the left and right side.

gimp-2 10_fO1cGKEU7J

### Motivation

Seems fairly simple and I like the look.

### DIFF
--- a/lua/lualine/component.lua
+++ b/lua/lualine/component.lua
@@ -115,16 +115,25 @@ function M:apply_icon()
   local icon = self.options.icon
   if self.options.icons_enabled and icon then
     if type(icon) == 'table' then
+      self.options.icon_alignment = icon.align
       icon = icon[1]
     end
-    if self.options.icon_color_highlight then
-      self.status = table.concat({
+    if self.options.icon_color_highlight and self.options.icon_alignment == 'right' then
+      self.status = table.concat {
+        self.status,
+        ' ',
+        self:format_hl(self.options.icon_color_highlight),
+        icon,
+        self:get_default_hl(),
+      }
+    elseif self.options.icon_color_highlight then
+      self.status = table.concat {
         self:format_hl(self.options.icon_color_highlight),
         icon,
         self:get_default_hl(),
         ' ',
         self.status,
-      })
+      }
     elseif self.options.icon_alignment == 'right' then
       self.status = table.concat({ self.status, icon }, ' ')
     else
--- a/lua/lualine/component.lua
+++ b/lua/lualine/component.lua
@@ -132,7 +132,7 @@ function M:apply_icon()
         ' ',
         self.status,
       }
-    elseif self.options.icon_alignment == 'right' then
+    elseif type(self.options.icon) == 'table' and self.options.icon.align == 'right' then
       self.status = table.concat({ self.status, icon }, ' ')
     else
       self.status = table.concat({ icon, self.status }, ' ')
--- a/lua/lualine/component.lua
+++ b/lua/lualine/component.lua
@@ -76,7 +76,9 @@ function M:apply_padding()
       -- When component has changed the highlight at begining
       -- we will add the padding after the highlight
       local pre_highlight = vim.fn.matchlist(self.status, [[\(%#.\{-\}#\)]])[2]
-      self.status = pre_highlight .. string.rep(' ', l_padding) .. self.status:sub(#pre_highlight + 1, #self.status)
+      self.status = pre_highlight
+        .. string.rep(' ', l_padding)
+        .. self.status:sub(#pre_highlight + 1, #self.status)
     else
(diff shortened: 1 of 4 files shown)


daephx:
Another thing I was looking at was was allow this to be set in `icon = table`.
But I'm not sure on the best way to handle this.

https://github.com/nvim-lualine/lualine.nvim/blob/4ba702e5d13da9c2041679b84467345834adaabc/lua/lualine/component.lua#L115-L117

Where you could get away with putting:

```diff
    if type(icon) == 'table' then
+     self.options.icon_right = icon.icon_right
      icon = icon[1]
    end
```
Or some variation on the name `icon = { '', align_right = true`?

shadmansaleh:
> So while the icon table options can be applied, the filetype component would need an option at a higher level from what I understand.

Why ?
I think we can make 
```lua
{'filetype', icon={align='right'}}
```
work . For filetype component icon gets ignored but it's still available in `self.options` table . We can just checif self.options.icon is a table and has align 'right' in it.

lua/lualine/component.lua
```diff
@@ -76,7 +76,9 @@ function M:apply_padding()
       -- When component has changed the highlight at begining
       -- we will add the padding after the highlight
       local pre_highlight = vim.fn.matchlist(self.status, [[\(%#.\{-\}#\)]])[2]
-      self.status = pre_highlight .. string.rep(' ', l_padding) .. self.status:sub(#pre_highlight + 1, #self.status)
+      self.status = pre_highlight
```
shadmansaleh, line 79: Did you change anything here other than formating ?

daephx: I Don't remember that line, so it probably got auto formatted and I didnt notice. My bad.

daephx:
> For filetype component icon gets ignored but it's still available in `self.options` table . We can just checif self.options.icon is a table and has align 'right' in it.

You're right, I just shied away from that to avoid checking `type() == 'table'` multiple times, cause it seems kind of verbose. Particularly when making sure the colors table still works.
```lua
if self.options.icon_color_highlight and type(self.options.icon) == 'table' and self.options.icon.align == 'right' then
      self.status = table.concat {
        self.status,
        ' ',
        self:format_hl(self.options.icon_color_highlight),
        icon,
        self:get_default_hl(),
      }
end
```

lua/lualine/component.lua
```diff
@@ -108,21 +108,32 @@ function M:apply_highlights(default_highlight)
   end
 end
 
----apply icon in front of component (prepemds component with icon)
+---apply icon to component (appends/prepemds component with icon)
 function M:apply_icon()
   local icon = self.options.icon
   if self.options.icons_enabled and icon then
     if type(icon) == 'table' then
+      self.options.icon.align = icon.align
```
shadmansaleh, line 116: What's this line suppose to do ? It seems you've just written 
```lua
x = x
```

lua/lualine/component.lua
```diff
@@ -108,21 +108,32 @@ function M:apply_highlights(default_highlight)
...
+      }
+    elseif self.options.icon_color_highlight then
       self.status = table.concat {
         self:format_hl(self.options.icon_color_highlight),
         icon,
         self:get_default_hl(),
         ' ',
         self.status,
       }
+    elseif self.options.icon_alignment == 'right' then
```
shadmansaleh, line 135: We don't have `options.icon_alignment` any more right ?

shadmansaleh:
> You're right, personally I shy away from that to avoid checking type(self.options.icon) == 'table' multiple times, cause it seems kind of verbose. Particularly when making sure the colors table still works.

Yeah the code is a bit verbose but since we already have color option in icon as table it makes sense to add other icon related options in the same table.

lua/lualine/component.lua
daephx: yup, wasn't on top of it and missed that one.

lua/lualine/component.lua
daephx: Nothing clearly, guess I made a poor assumption instead of properly reading.

lua/lualine/component.lua
daephx: I reverted the change. My neovim should utilizes project specific `.stylua.toml` settings so I'm surprised that occurred.

shadmansaleh:
This looks alright to me. Now we just need to document it in README with icon options & a test or two to ensure it works in future.

For test take a look at lualine_spec . You can just setup lualine with a config then assert statusline state with statusline module .

shadmansaleh:
lgtm . Thanks for your contribution :)

Issue #668, opened 2022-04-26
Feature request: Option to display buffer number instead of buffer index


MrcJkb:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem
Execute the following commands:
```vimscript
:edit first
:edit second
:bdelete first
:ls
```
Outputs:
```
2 %a   "second"                       line 1
```

### Expected behaviour
The lualine buffer section is consistent with the output of `:ls` (i.e. the lualine buffer number is `2`)

### Actual behaviour
The lualine buffer section is _not_ consistent with the output of `:ls` (i.e. the lualine number is `1`)

### Minimal config to reproduce the issue

```vimscript
call plug#begin("/home/mrcjk/tmp/.local/share/nvim/plugged")
Plug 'nvim-lualine/lualine.nvim'
call plug#end()
lua << END
require'lualine'.setup {
  tabline = {
    lualine_b = {{'buffers', mode = 2}}
  }
}
END
```

MrcJkb:
Nevermind. I just realised that `mode=2` is defined as "show buffer index", not buffer number.
It would be neat to have a config to display the buffer number instead of the index though (maybe via a separate option).
I will change this issue to a feature request...

PR #669, opened 2022-04-26
Add option to show buffer number + alternate file icon

MrcJkb:
This is a rough draft to address #668.
Still needs documentation (and maybe a test spec), which I will add if the feature request is approved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,7 +441,7 @@ sections = {
                 -- 1: Shows buffer index
                 -- 2: Shows buffer name + buffer index 
                 -- 3: Shows buffer number
-                -- 4: Shows buffer nmae + buffer number
+                -- 4: Shows buffer name + buffer number
 
       max_length = vim.o.columns * 2 / 3, -- Maximum width of buffers component,
                                           -- it can also be a function that returns
--- a/README.md
+++ b/README.md
@@ -439,7 +439,9 @@ sections = {
 
       mode = 0, -- 0: Shows buffer name
                 -- 1: Shows buffer index
-                -- 2: Shows buffer name + buffer index
+                -- 2: Shows buffer name + buffer index 
+                -- 3: Shows buffer number
+                -- 4: Shows buffer nmae + buffer number
 
       max_length = vim.o.columns * 2 / 3, -- Maximum width of buffers component,
                                           -- it can also be a function that returns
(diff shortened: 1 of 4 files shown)


shadmansaleh:
I like the idea of identifying alternate buffer .

I think we should just make mode 4 & 5 that are same as 2 & 3 but shows bufnr instead if buffer index .

shadmansaleh:
I like the idea of identifying alternate buffer .

I think we should just make mode 4 & 5 that are same as 2 & 3 but shows bufnr instead if buffer index .

MrcJkb:
> I like the idea of identifying alternate buffer .
> 
> I think we should just make mode 4 & 5 that are same as 2 & 3 but shows bufnr instead if buffer index .

Sure. What about the default behaviour?
Would you like to preserve the current default `mode = 2`?
I would suggest changing the default `mode` to 0, as that is what is set in the readme example. But it may surprise users who don't have `mode` set in their configs after an update...

Regarding the alternate buffer identifier: Would you prefer an option to enable/disable it?

shadmansaleh:
> What about the default behaviour?
Would you like to preserve the current default mode = 2?

Current default mode is 0 no reason to change that. It's set [here](https://github.com/nvim-lualine/lualine.nvim/blob/45d07fc026400c211337a7ce1cb90e7b7a397e31/lua/lualine/components/buffers/init.lua#L12) by the way . I think you thought it's 2 because of that functions  structure.

> Regarding the alternate buffer identifier: Would you prefer an option to enable/disable it?

You can expose it as an option under symbols table .
```lua
options.symbols = {
  modified = ....,
  alternate_file = ....,
}
```
Similar to how diff & diagnostics component exposes their icons . That way if someone doesn't want it they can just set it to empty string.

shadmansaleh:
You need to have plenary.nvim . Then run `make test` . See [here](https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md#devloper-tools)

MrcJkb:
I am getting inconsistent buffer numbers in my tests between NeoVim 0.5 and 0.6.
Are some tests excluded depending on the NeoVim version? It won't be possible to reset buffer indexing without quitting the session.

shadmansaleh:
> I am getting inconsistent buffer numbers in my tests between NeoVim 0.5 and 0.6.

This is something I've noticed too . You can try matching a pattern instead of exact number . Or you could get the bufnr for the bufname and test against that . That way you'll be varifying the bufnr too in the test though pattern match is easier to write as it's built into statusline test module :)

> Are some tests excluded depending on the NeoVim version?

Not currently . Though you could run different test code in the test based on which nvim version you're on , probably not a good idea in this case as it might break the test on nightly or a new version anytime.

MrcJkb:
I implemented the bufnr detection. It turns out that the inconsistency is because NeoVim 0.5 does not create an unnamed buffer. By silently deleting the alternate buffer after creating the first buffer `a.txt`, we can achieve consistent results.

shadmansaleh (approved):

shadmansaleh:
lgtm , thanks.

Issue #578, opened 2022-02-14
Bug: lualine doesn't show filetype icon for files without extension


kassio:
The 'filetype' component doesn't find the filetype icon sometimes:

<img width="1025" alt="_HOME_src_gdk_gitlab_a_rb" src="https://user-images.githubusercontent.com/120483/153911616-140a0326-e565-462c-9b6e-f3cb57da5f11.png">

<details>
<summary>My config</summary>

```lua
lualine.setup({
  extensions = {
    require('plugins.statusline.extensions.filetree'),
    require('plugins.statusline.extensions.neoterm'),
  },
  options = {
    theme = theme.colorscheme,
    icons_enabled = true,
    section_separators = '',
    component_separators = '',
  },
  sections = {
    lualine_a = { utils.bufnr() },
    lualine_b = {
      {
        'filetype',
        colored = true,
        icon_only = true,
        padding = 1,
      },
      {
        'filename',
        file_status = true,
        path = 1,
        shorting_target = 30,
        symbols = { modified = ' ', readonly = ' ', unnamed = ' [No Name] ' },
        padding = 0,
      },
    },
    lualine_c = {
      { vim.my.treesitter.gps.location },
    },
    lualine_x = {
      {
        'diagnostics',
        sources = { 'nvim_diagnostic' },
        symbols = {
          error = theme.signs.error,
          warn = theme.signs.warn,
          hint = theme.signs.hint,
          info = theme.signs.info,
        },
        color_error = theme.colors.error,
        color_warn = theme.colors.warn,
        color_hint = theme.colors.hint,
        color_info = theme.colors.info,
      },
    },
    lualine_y = { 'diff' },
    lualine_z = {
      { '[[%3l:%-3c]]' },
      utils.mode(),
    },
  },
  inactive_sections = {
    lualine_a = { utils.bufnr() },
    lualine_b = utils.filename({ icon_color = false }),
    lualine_c = {},
    lualine_x = {},
    lualine_y = {},
    lualine_z = {},
  },
})
```

</details>

### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem

I'm not sure, but it seems related to how the filetype is set, when it's set in `ftdetect` with `BufReadPos` it doesn't set the icon.

### Minimal config to reproduce the issue

<details>
<summary>Config</summary>

</details>

shadmansaleh:
[...]

PR #674, opened 2022-05-06
PoC tagline devicons and modifiability/modified markers

Slotos:
## Summary

_Obligatory request to not judge the code too harshly - first time dealing with lua in a serious manner. I'd spent more time figuring out patterns and execution flow than beautifying the code._

This is an attempt to implement the behaviour I'd gotten attached to with lualine back in the day - devicons, modification and modifiability markers in the tabaline, derived from currently active buffers on the tab pages.

image

```lua
    tabline = {
      lualine_a = {
        {
          'tabs',
          max_length = vim.o.columns,
          mode = 2,
          icons_enabled = true,
          modification_icons_enabled = true,
          icon = {
            align = 'right',
            use_default = true
          },
        }
      },
    },
```

Might I request assistance from someone who's experienced with lua in order to get this to a mergeable state?

## Caveats

- I'd taken a shortcut and made tab component extend filetype one in order to avoid introducing a competing standard No. 42. This probably should instead be replaced by extraction of `apply_icon` function into a utility module to be reused by specific components as they see fit.
- While fumbling around with highlight groups generation, I discovered a bug in `highlight.create_component_highlight_group` which I'd fixed in a separate commit so that it can be pulled in if need be. I don't think the bug surfaces often (it took some very bad code to trigger it), but it can lead to attempts to generate highlight groups with names longer than allowed.
- I'd implemented a simple icon lookup fallback, which might mostly resolve https://github.com/nvim-lualine/lualine.nvim/issues/578. It doesn't resolve cases with ruby file having its buffer's filetype set to something else, but that aligns with recommended behaviour in https://github.com/kyazdani42/nvim-web-devicons/pull/125. Again, a separate commit, although it's unlikely to be cherry-pickable.

### DIFF
--- a/lua/lualine/component.lua
+++ b/lua/lualine/component.lua
@@ -211,8 +211,8 @@ end
 ---Get stl formated hl group for hl_token
 ---@param hl_token table indentifier received from create_hl or create_component_highlight_group
 ---@return string stl formated hl group for hl_token
-function M:format_hl(hl_token)
-  return highlight.component_format_highlight(hl_token)
+function M:format_hl(hl_token, is_focused)
+  return highlight.component_format_highlight(hl_token, is_focused)
 end
 
 -- luacheck: push no unused args
--- a/lua/lualine/components/filetype.lua
+++ b/lua/lualine/components/filetype.lua
@@ -28,30 +28,43 @@ function M:apply_icon()
     return
   end
 
-  local icon, icon_highlight_group
+  local icon_char, icon, icon_highlight_group
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
-    local f_name, f_extension = vim.fn.expand('%:t'), vim.fn.expand('%:e')
-    f_extension = f_extension ~= '' and f_extension or vim.bo.filetype
-    icon, icon_highlight_group = devicons.get_icon(f_name, f_extension)
+    local f_name = self.f_name or vim.fn.expand('%:t')
+    local f_extension
 
-    if icon and self.options.colored then
+    if self.f_extension then
+      f_extension = self.f_extension
+    else
+      f_extension = vim.fn.expand('%:e')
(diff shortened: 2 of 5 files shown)


shadmansaleh:
The idea sounds good in general . I'll look at the code later .

> I'd spent more time figuring out patterns and execution flow than beautifying the code.

You don't need to beautify code . Code formating is handled by stylua. You can run it locally too take a look at CONTRIBUTING.md

> Might I request assistance from someone who's experienced with lua in order to get this to a mergeable state?

Sure you can ask for help when you need .


It'll be better if you open a separate pr with the bug fix . It can be merged faster and you won't need approval for CI . 

We should definitely extract interaction with nvim-web-devicons from filetype component to a separate module. extending filetype component to build tabs component doesn't make a lot of sence.

what does `use_default = true` do ?

Slotos:
> It'll be better if you open a separate pr with the bug fix

Will do a bit later today 👍 

> You don't need to beautify code

I'd added quite a bit of calls that were not there before, and readability definitely suffered. Some of `self` properties are being set up in a... opportunistic manner (e.g. `self.bufnr` in `Tab:render`), which I failed to find a better place for due to uncertainties about execution flow. Although after a night's sleep I think `init` is perfectly suited for that.

> what does use_default = true do ?

Enables fallback to default icon for `nvim-web-devicons` if specific icon is not resolved. Initially it was simply a third argument to `get_icon`, but with icon extraction fallbacks it had to use a hacky `get_icon('', '', { default = true })` approach,  since `nvim-web-devicons` doesn't export `get_default_icon` function (I'll write a PR for that, it's a good compliment to existing `set_default_icon`).

I'll make sure tests run locally and will work on extracting icon routines into a utility module.

Slotos:
Closing this PR until I have a better solution to be presented in a separate PR.

PR #675, opened 2022-05-06
fix: large file lag

dmun:
Since wordcount() also provides your current cursor position, it's getting called for every single cursor movement. This is kind of a problem for big files since it takes longer to get the data. Typing is also kinda laggy because of it. I'm basically removing the live filesize again, since it was kinda trivial anyway.


shadmansaleh:
We can try implementing this asynchronously with luv & coroutines to avoid performance issues .

PR #676, opened 2022-05-06
Support tilde as home path in the absolute file paths

jacks0n:


doc/lualine.txt
```diff
@@ -167,10 +167,10 @@ how to do it <./CONTRIBUTING.md>.
 
 >
     local custom_gruvbox = require'lualine.themes.gruvbox'
-    
+
```
shadmansaleh, line 170: don't make changes to this file

lua/lualine/components/filename.lua
```diff
@@ -43,6 +43,10 @@ M.update_status = function(self)
   elseif self.options.path == 2 then
     -- absolute path
     data = vim.fn.expand('%:p')
+  elseif self.options.path == 3 then
+    -- absolute path, with tilde
+    local home_dir = os.getenv('HOME')
+    data = '~/' .. string.sub(vim.fn.expand('%:p'), #home_dir + 2)
```
shadmansaleh, line 49: This doesn't make sense. What happens when your home is /home/asdf and you've opened a file in /usr/share . Use builtin filename modifications with expand instead (something like this `expend('%:p:~)`).

shadmansaleh (changes requested):

shadmansaleh:
lgtm, thanks for your contribution.

shadmansaleh (approved):

PR #677, opened 2022-05-08
Use single counter when creating component highlights

Slotos:
`highlight.create_component_highlight_group` right now keeps
on adding new suffixes after already tried ones, which
leads to exceeding of highlight group name limit quite fast.

This commit fixes the problem by never changing the base of
highlight group name when iterating with a counter.

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -304,16 +304,17 @@ end
 function M.create_component_highlight_group(color, highlight_tag, options, apply_no_default)
   local section = options.self.section
   local tag_id = 0
+  local highlight_tag_counted = highlight_tag
   while
-    M.highlight_exists(table.concat({ 'lualine', section, highlight_tag }, '_'))
-    or (section and M.highlight_exists(table.concat({ 'lualine', section, highlight_tag, 'normal' }, '_')))
+    M.highlight_exists(table.concat({ 'lualine', section, highlight_tag_counted }, '_'))
+    or (section and M.highlight_exists(table.concat({ 'lualine', section, highlight_tag_counted, 'normal' }, '_')))
   do
-    highlight_tag = highlight_tag .. '_' .. tostring(tag_id)
+    highlight_tag_counted = highlight_tag .. '_' .. tostring(tag_id)
     tag_id = tag_id + 1
   end
 
   if type(color) == 'string' then
-    local highlight_group_name = table.concat({ 'lualine', section, highlight_tag }, '_')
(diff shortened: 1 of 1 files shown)


lua/lualine/highlight.lua
```diff
@@ -304,16 +304,17 @@ end
...
+    or (section and M.highlight_exists(table.concat({ 'lualine', section, highlight_tag_counted, 'normal' }, '_')))
   do
-    highlight_tag = highlight_tag .. '_' .. tostring(tag_id)
+    highlight_tag_counted = highlight_tag .. '_' .. tostring(tag_id)
     tag_id = tag_id + 1
   end
 
   if type(color) == 'string' then
-    local highlight_group_name = table.concat({ 'lualine', section, highlight_tag }, '_')
+    local highlight_group_name = table.concat({ 'lualine', section, highlight_tag_counted }, '_')
```
shadmansaleh, line 317: We can probably just put highlight_tag_counted value in hightlight_tag. That way codechanges from this line can be avoided . And counted variable isn't necessary after this and the counted part below this seems unnecessary. Name conflict resolution  can be done int this part of the code and the rest of the function can be unaware of that . 

```lua
highlight_tag = highlight_tag_counted
```
what do you think?

Slotos: Should we perchance extract unique suffix search into its own function then?

```lua
local function append_unique_suffix_to_highlight_group(highlight_tag, section) -- naming is hard
  local tag_id = 0
  while
    M.highlight_exists(table.concat({ 'lualine', section, highlight_tag }, '_'))
    or (section and M.highlight_exists(table.concat({ 'lualine', section, highlight_tag, 'normal' }, '_')))
  do
    highlight_tag = highlight_tag .. '_' .. tostring(tag_id)
    tag_id = tag_id + 1
  end
  return highlight_tag
end

function M.create_component_highlight_group(color, highlight_tag, options, apply_no_default)
  local section = options.self.section
  highlight_tag = append_unique_suffix_to_highlight_group(highlight_tag, section) -- could yield a new variable here, to explicitly show that we aren't using incoming value as is
  ...
```

I personally prefer any long form argument manipulation to yield a new variable or be extracted into an early explicitly named function call. Function signature and its first few lines should provide all the information about its parameters expected values IMO.
As a newcomer to the codebase frantically looking into the source of "highlight name too long" error, this block of code wasn't immediately noticeable - it doesn't set highlights itself, while the latter blocks that do seem to be using incoming argument as is.

shadmansaleh: Extracting it to a separate function sounds like a better idea . Sure you can do that .

PR #689, opened 2022-05-22
feat: add winbar support

diegodox:
Support winbar added to nvim in this [PR](https://github.com/neovim/neovim/pull/18562).

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -281,7 +281,6 @@ local function set_statusline()
 end
 
 --- Sets &winbar option to lualine
---- adds auto command to redraw lualine on VimResized event
 local function set_winbar()
   if next(config.winbar) ~= nil or next(config.inactive_winbar) ~= nil then
     vim.go.winbar = "%{%v:lua.require'lualine'.winbar()%}"
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -280,6 +280,36 @@ local function set_statusline()
   end
 end
 
+--- Sets &winbar option to lualine
+--- adds auto command to redraw lualine on VimResized event
+local function set_winbar()
+  if next(config.winbar) ~= nil or next(config.inactive_winbar) ~= nil then
+    vim.go.winbar = "%{%v:lua.require'lualine'.winbar()%}"
+  end
+end
+
+-- lualine.winbar function
+--- Draw correct winbar for current window
+---@param focused boolean : force the value of is_focused . useful for debugging
+---@return string winbar string
+local function winbar_dispatch(focused)
+  local retval
+  local current_ft = vim.bo.filetype
+  local is_focused = focused ~= nil and focused or modules.utils.is_focused()
+  for _, ft in pairs(config.options.disabled_filetypes) do
+    -- disable on specific filetypes
+    if ft == current_ft then
+      return ''
+    end
+  end
+  if is_focused then
+    retval = statusline(config.winbar, is_focused)
+  else
+    retval = statusline(config.inactive_winbar, is_focused)
+  end
+  return retval
+end
+
 -- lualine.statusline function
 --- Draw correct statusline for current window
 ---@param focused boolean : force the value of is_focused . useful for debugging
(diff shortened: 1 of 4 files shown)


shadmansaleh:
I'll wait on it and see first what winbar can best be used for . Also what ideas community can come up with for winbar . For now I don't think using winbar just as a statusline is a good idea . I'll prefer to have winbar to be buffer specific unlike global . A winbar specific plugin can probably pull it of better than if just treat is as part of statusline in lualine.

So I'm putting it on hold . Not entirely rejecting it nor accepting the idea.

diegodox:
@shadmansaleh
I agree with you.
I implement this just for me to try how I feel when use `winbar` as status bar. But I found #688, so I decided to create a PR for who interested in this.

btw I think wibar is a nice place to have information like `nvim-gps`, `lsp-signature`'s statusline and  lsp status.

>I've only noticed it to be used in vimspector

I did not know `vimspector`, it looks like winbar is made for it.

lkhphuc:
@diegodox exactly what I want. There is a lot of buffer specific info in my global statusline at the moment.
Just putting the `file name > class > method > matchparen` of the current cursor position is probably all I want.

diegodox:
PS: I found winbar is only reasonable place to show filename of each buffer (especially inactive buffer) with global status line.

rockyzhang24:
The most useful points of winbar are:
1. Full path of the filename (`%F`)
2. LSP symbol path (`aerial.nvim` has API to output this)
3. Window number (`vim.api.nvim_win_get_number()`)

Setting them is very easy and no need further supports. I believe a fancy `statusline` is necessary but a plus fancy `winbar` at the top will be distracting. Also, a verbose and complicated `winbar` will get truncated and making it meaningless at all.

Screen Shot 2022-05-23 at 15 53 56

bellini666:
Sorry for being totally off topic here guys...

But @rockyzhang24, do you mind sharing your font and color scheme from that screenshot?

rockyzhang24:
@bellini666 I will move it to the discussion page. Sorry about that.

> do you mind sharing your font and color scheme from that screenshot?

Font: JetBrainsMono
Color scheme: `nightfox.nvim` and I configured some colors that are not easily distinguished (See this please if you need: https://github.com/rockyzhang24/dotfiles/blob/master/.config/nvim/lua/plugin/nightfox.lua)

barrett-ruth:
Furthermore, with the creation of [nvim-navic](https://github.com/SmiteshP/nvim-navic), I think the community has certainly made up it's mind about the utility of the winbar.

alex-popov-tech:
imho statusline and winbar aim different purposes, statusline for more global things, which are shared across splits, and winbar are about some specific splint only 
CleanShot 2022-06-14 at 22 23 24@2x

joeveiga:
@alex-popov-tech I think the ideal approach would be to be flexible enough for users to display whatever information they want on each. Like we do now with lualine sections, but having an extra set of section keys in the configuration for winbar sections.

In a perfect world users would be able to add lualine plugins currently used for statusline to the winbar, and vice-versa.

As you point out, a popular configuration could be winbar for buffer specific info and statusline for global/project level info. Another one could be, for example, statusline for the focused buffer (with a more detailed view) and winbar for inactive buffers.

barrett-ruth:
@diegodox I'd like to add that because you set the global winbar to `winbar()` it becomes difficult to disable lualine winbar on certain files in a different window. For example, disabling the winbar on `NvimTree` sidebar becomes basically impossible. There are some discussion about this (e.g. [here](https://github.com/neovim/neovim/issues/18660)

However, I may just be doing something stupid and it is indeed possible with your commits (thanks!).

shadmansaleh:
@diegodox  I've pulled the changed from this branch to #736 .

Would appreciate if you guys interested in the feature can test it.

diegodox:
Need advice on implementation approach, is autocmd welcome to set winbar window option? or more simpler way exists?

diegodox:
@shadmansaleh I will look and test that branch, thank you.

shadmansaleh:
> Thank you for pointing out vim.go.winbar will not respect vim.wo.winbar. I'm planning to fix that (use vim.wo.winbar instead of vim.go.winbar). And sorry for late reply.
But sorry again, I have no time to impl that right now. Anyone who want to impl that feature are welcome. :)

fixing that issue is practically the purpose of #736 . So lualine can keep running even when options are unset.

shadmansaleh:
Merged in 53aa3d82d9d9329880b4197f278fcb74051a6af5

PR #691, opened 2022-05-24
Add always_visible option for the diff component

arthurpi:


arthurpi:
I personally dislike having partial git status showing, for example: `+1 -2`, and I prefer to see the full form `+1 ~0 -2`.

This PR adds an option to do exactly this. If _any_ stat is non-zero, we print every stat.
The new option default is the existing behavior, so no changes will be observed unless users opt-in.

Another valid approach could be to always print every stat, even if all are zeroes, as long as we are in a git repo.
But not showing anything at all in zero-diff scenario is I think nicer.

README.md
```diff
@@ -513,12 +513,13 @@ sections = {
   lualine_a = {
     {
       'diff',
-      colored = true, -- Displays a colored diff status if set to true
+      colored = true,          -- Displays a colored diff status if set to true.
+      show_all_if_any = false, -- Shows every value if any is non-zero.
```
shadmansaleh, line 517: I think we should try another name . This one seems confusing . But I'm not sure what it should be named either.

PR #696, opened 2022-05-27
docs: cleanup formatting and comment typos

daephx:
- Fix typo in `LualineNotices`
- Rename misspelled function parameter [diff](https://github.com/nvim-lualine/lualine.nvim/pull/696/files#diff-f8f8ed0fbbfef26340dbb80180776b9d89d79ff7134c2de3be7e9f897bf19959L73-R77)
- Cleanup minor typos and wording in comments
- Cleanup minor typos and wording in docs, i.e. README, CONTRIBUTING, THEMES

Initially I spotted a small misspelling in the `LualineNotices` output: `paris` -> `pairs`

https://github.com/nvim-lualine/lualine.nvim/blob/c12b1673107c181e32ce54f2dc4c76a2a884d7ba/lua/lualine/utils/loader.lua#L113

But I took the plunge on improving spelling and readability where I spotted issues across most files.
A few changes might be a bit pedantic, like `spellcapcheck` errors but excluding those, I think a bit of cleanup here and there makes the project more presentable and easier to navigate.

## Themes

A lot of theme files had the same misspelled comments, I believe `Line 85` of the theme conversion tool has a single typo that likely propagates to any themes that are created using the script.

`Genarated` -> `Generated`

[lightline2lualine_theme_converter.lua](https://gist.github.com/shadmansaleh/000871c9a608a012721c6acc6d7a19b9)

https://github.com/nvim-lualine/lualine.nvim/blob/c12b1673107c181e32ce54f2dc4c76a2a884d7ba/lua/lualine/themes/ayu_dark.lua#L3

I can't make a PR on a gist, so fixing this should prevent new themes from introducing more misspellings.


README.md
```diff
@@ -705,7 +708,7 @@ For example:
 :LualineRenameTab Project_K
 ````
 It's useful when you're using rendering mode 2/3 in tabs.
-To unname a tablage run `:LualineRenameTab` without argument.
+To rename/unname a tabpage run `:LualineRenameTab` without argument.
```
shadmansaleh, line 711: To remove name ?

If additional argument isn't provided `:LualineRenameTab` removes the currently set name.

lua/lualine.lua
```diff
@@ -219,8 +219,8 @@ end
 
 --- sets up theme by defining hl groups and setting theme cache in highlight.lua
 --- uses options.theme option for theme if it's a string loads theme of that name
---- if it's a table directlybuses it .
---- when theme load fails this fallsback to 'auto' theme if even that fails
+--- if it's a table it directly buses it.
```
shadmansaleh, line 222: it'll be
> directly passes it .

shadmansaleh (commented):
Overall looks great . I think these few things can be changed.

lua/lualine.lua
daephx: So the line should say?: `-- if it's a table, it'll be directly bused.`

README.md
daephx: Good catch! Fixed

daephx:
Pushed the suggestions and a few more things I spotted. 
Let me know if there's anything else 👍

lua/lualine.lua
shadmansaleh: lets make it
```
-- if it's a table, it's directly used as theme.
```

daephx: My brain kinf of hurt reading: `uses options.theme option for theme if it's a string loads theme`
Here's a revision of the whole comment block, what do you think?

```
--- Sets up theme by defining hl groups and setting theme cache in 'highlight.lua'.
--- Uses 'options.theme' variable to apply the theme:
--- - If the value is a string, it'll load a theme of that name.
--- - If it's a table, it's directly used as theme.
--- If loading the theme fails, this falls back to 'auto' theme.
--- If the 'auto' theme also fails, this falls back to 'gruvbox' theme.
--- Also sets up auto command to reload lualine on ColorScheme or background changes.
```

shadmansaleh: That seems alright to me you can use it.

daephx:
Well, that was a decent excuse to read through a bit on how the plugin functions.
For now, I don't see anything else worth worrying about.

shadmansaleh:
Thanks a lot. The docs definitely needed some work.

PR #729, opened 2022-07-06
Add option to sort/filter buffers by :ls flags

ChHaeni:
Buffers can be sorted according to the time they were last accessed (as printed by `:ls t`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,7 @@ sections = {
       show_filename_only = true,   -- Shows shortened relative path when set to false.
       hide_filename_extension = false,   -- Hide filename extension when set to true.
       show_modified_status = true, -- Shows indicator when the buffer is modified.
+      sort_last_accessed = false, -- Sorts buffers by the time they were last accessed when set to true.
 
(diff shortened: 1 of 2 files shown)


jhthorsen:
I think this is a great addition, but could it be possible to change "sort_last_accessed" to "sort_by" instead, where the user could set it to "t", "ta", or something else?

ChHaeni:
I'm open to any other option naming other than `ls_flags` and/or option description.

ChHaeni:
Many thanks to Jan Henning Thorsen for his suggestion to extend the sorting to any :ls flag!

jhthorsen:
Sorry for deleting my message, but I did not like the way the buffers were “jumping” around when the sorting changed.

This is unrelated to this PR, but I would rather prefer if I could manually reorder the buffers like you can with bufferline.nvim, but I haven’t found the time to implement that.

ChHaeni:
Closing this PR since #887 addresses buffer sorting in a more generic way.

PR #735, opened 2022-07-11
Use the original gruvbox colors

f1rstlady:
Lualine's gruvbox theme differs from lightline's gruvbox theme.  Hence,
it is updated (using the lightline2lualine theme converter) to resemble
the original theme.

The original lightline theme may be found at
https://github.com/morhetz/gruvbox/blob/master/autoload/lightline/colorscheme/gruvbox.vim

I have already created screenshots of the updated theme to update the
screenshots in the list of themes, but unfortunately I do not know how to add
them via `user-images.githubusercontent.com`.

### DIFF
--- a/lua/lualine/themes/gruvbox_dark.lua
+++ b/lua/lualine/themes/gruvbox_dark.lua
@@ -2,47 +2,47 @@
 -- MIT license, see LICENSE for more details.
 -- stylua: ignore
 local colors = {
-  black        = '#282828',
-  white        = '#ebdbb2',
-  red          = '#fb4934',
-  green        = '#b8bb26',
-  blue         = '#83a598',
-  yellow       = '#fe8019',
-  gray         = '#a89984',
-  darkgray     = '#3c3836',
-  lightgray    = '#504945',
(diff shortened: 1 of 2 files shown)


f1rstlady:
I have no clue why the three testing pipelines fail, especially since I only changed a theme. If anyone can help me with this regard, it would be appreciated.

shadmansaleh:
I don't think it's necessary to replicate lightlines theme here. I think originally @hoob3rt had made the gruvbox theme by mixing both lightline & airlines theme. Most of the the themes are usually provided by colorschemes and it's easy enough to have a custom theme on users side for example you can just pass them table in theme option or just place the file under ~/.config/nvim/lua/lualine/themes and lualine will load this theme instead of the one in the repo. So I think we should leave it as it is.

The tests are failing because gruvbox were basically first theme made for lualine and we've been using it as default theme for the tests . So colors are no longer matching since you changes the theme spec.

f1rstlady:
Ok. Thanks for your detailed explanation, regarding both the themes and the test failure!

Issue #692, opened 2022-05-25
Feat: Winbar support


mrjones2014:
### Requested feature

Just like how Lualine supports both statusline (including global statusline) and tabline, it should also support the new `winbar`. I would love to show the buffer list in each window, with each window showing the active buffer as the buffer active in it's specific window. I currently have a fork of bufferline.nvim doing this but its jank and buggy.

### Motivation

Lualine already has a built-in `buffers` component, and already supports statusline and tabline. winbar feels like a very natural extension to that. I would love to be able to use Lualine's API to build out my `winbar` configuration.

shadmansaleh:
See my response [here](https://github.com/nvim-lualine/lualine.nvim/pull/689#issuecomment-1133937406). Relevent discussion https://github.com/nvim-lualine/lualine.nvim/discussions/688

Issue #395, opened 2021-10-22
Feat: Hide the tabline conditionally


snorkysnark:
### Requested feature
In global options, add a function that is called on redraw to determine whether the tabline should be visible. For example, you can only show it if there is more than one tab/more than one buffer, etc.
```lua
require('lualine').setup{
  tabline = {
    ...
  },
  tabline_visible = function()
    return vim.fn.tabpagenr('$') > 1
  end
}
```

### Motivation

Currently, this plugin forces the tabline to be always visible, decreasing the available screen space. Since the option is set during the first redraw, I haven't found a good way to override it with autocmds.
As people have different tabline configurations (with buffers and other custom components), you need a more flexible alternative to vim's showtabline=1 (that shows and hides the tabline based on tab count)

shadmansaleh:
For now I'm dropping this idea. As it's too complicated to add for too little gain for now.

amerlyq:
I wanted to achieve the same effect as airline -- to hide tabline when there is only single buffer:
```
let g:airline#extensions#tabline#buffer_min_count = 2
```
But I don't understand how to use your workaround above to achieve the same thing.
@shadmansaleh do you have any other suggestion?

Maybe simple corner case of "hide when there is only one buffer/tab" can be much more easily hacked, 
than the proper solution of "evaluate conditional function inside of tabline" ?

AlxHnr:
This hack does not work in neovim 0.7.0:

> ```viml
> au OptionSet showtabline :set showtabline=1
> ```

For some reason OptionSet is not fired when lualine does it. I can set `showtabline=1` manually, but when switching colorschemes it reappears. Is there any reason why we can't add an option to simply leave `showtabline` alone [here](https://github.com/nvim-lualine/lualine.nvim/blob/dfac736a015f1c295eae517d9af0db87413a88ac/lua/lualine.lua#L256-L264)?

mario-amazing:
@shadmansaleh, because of not working OptionSet in neovim 0.7.0, can we reopen this feature request?

matheussampaio:
Maybe an easier approach is to let users control if we want [Lualine to set `showtabline=2`](https://github.com/nvim-lualine/lualine.nvim/blob/3cf45404d4ab5e3b5da283877f57b676cb78d41d/lua/lualine.lua#L508)?

If we wrap that in an condition, it would allow users to set `showtabline=1` in their config and get the default behavior of only displaying the tabline if at least two tabs exists:

```
[...]

PR #736, opened 2022-07-12
feat: refresh lualine based on timer + winbar support. (experimental)

### LINKED ISSUE #744: Feat: global statusline should still displays for all filetypes, including `ignored_filetypes`
### Requested feature
When global statusline is used, statusline should works for all buffer, including those in `disabled_filetypes` tables.

### Motivation

`disabled_filetypes` is used to prevent rendering a crammed statusline in overly small windows, this doesn't apply with global statusline.

- One option is just to not disabled any filetypes, but that means winbar (if it is ever officially supported #736 ) will not ignore some filetypes with small windows either.
- Another option is to have separate options, i.e `statusline_disabled_filetypes` and `winbar_disabled_filetypes` (again if it's ever supported)
- Or just ignore `ignored_filetypes` altogether when using global statusline.

shadmansaleh:
### Needs testing

Try testing this branch and if you find any issue let me know in comment

### What it is
With this lualine will control it's own refresh instead of relying on nvim.

If we refresh this way we should be able to change out the statusline, winbar, tabline options and still be running . We should be able to address https://github.com/nvim-lualine/lualine.nvim/pull/689#issuecomment-1157194479 & #395 with this style of update.

### Major changes
-  lualine now refresh based on timer + some auto-commands.
- added `require('lualine').refresh` function to force refresh lualine . It can refresh all statusline, tabline, winbars or can granularly refresh specific statusline/tabline/winbar of specific windows see the code docs for more info.
-  winbar support. (originally form #689 by @diegodox)
- winbar support in extensions.
-  newly added options
    - options.refresh.statusline
    - options.refresh.tabline
    - options.refresh.winbar 
    -  options.disabled_filetypes.statusline
    -  options.disabled_filetypes.winbar
    -  winbar
    -  inactive_winbar
-  various bug fixes.

### known Issues/todos:
- ~~Refresh time is set to 1s . It seems when changing wins inactive & active stls don't switch immediately instead takes that 1s to update which is quite noticeable . Possible solutions: utilize autocommads to refresh on such events.~~
- ~~We need to handle errors in the refresh funcitnon so if we say fail to update 3 times in a row we should stop the timer~~.
- ~~looks like nvim wants to put winbars on floating windows too. but it might fail if there aren't enough room like telescopes prompt . that's an issue for us. wonder if we can determine if a win is float or not and skip floats 🤦‍♂️~~ (skiping winbar on winheight <= 1 & floating windows)
- ~~try merging statusline_dispatch and winbar_dispatch togather they have a lot of duplication~~
- ~~statusline doesn't seem to update regularly in cmdline mode~~
- ~~nvim-0,5 & 0.6 is broken with nvim_opts module . (may we should just have compat tags for older nvim version and be done with them . keeping support for them is getting annoying )~~
- ~~live update seems broken for gobalstatus & separators options.~~
- yes we have that intro disappearing due to redraw issue again.

closes #744 #692
supersedes #689

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,10 @@ require('lualine').setup {
     theme = 'auto',
     component_separators = { left = '', right = ''},
     section_separators = { left = '', right = ''},
-    disabled_filetypes = {},
+    disabled_filetypes = {
+      statusline = {},
+      winbar = {},
+    },
     always_divide_middle = true,
     globalstatus = false,
     refresh = {
@@ -343,7 +346,10 @@ options = {
   theme = 'auto', -- lualine theme
   component_separators = { left = '', right = '' },
   section_separators = { left = '', right = '' },
-  disabled_filetypes = {},     -- Filetypes to disable lualine for.
+  disabled_filetypes = {     -- Filetypes to disable lualine for.
+      statusline = {},       -- only ignores the ft for statusline.
+      winbar = {},           -- only ignores the ft for winbar.
+  },
 
   always_divide_middle = true, -- When set to true, left sections i.e. 'a','b' and 'c'
                                -- can't take over the entire statusline even
--- a/README.md
+++ b/README.md
@@ -124,6 +124,11 @@ require('lualine').setup {
     disabled_filetypes = {},
     always_divide_middle = true,
     globalstatus = false,
+    refresh = {
+      statusline = 1000,
+      tabline = 1000,
+      winbar = 1000,
+    }
   },
   sections = {
     lualine_a = {'mode'},
@@ -142,6 +147,8 @@ require('lualine').setup {
     lualine_z = {}
   },
   tabline = {},
+  winbar = {},
+  inactive_winbar = {},
   extensions = {}
 }
 ```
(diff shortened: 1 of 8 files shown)


shadmansaleh:
I don't think that's the case here. We are respecting disabled filetypes but It seems telescope is creating 6 windows but it only marks 2 of those with filetype TelescopeResults and TelescopePrompt. I'm not using floaterm but toggleterm seems to work correctly with disabled_filetype. Maybe floaterm does something similar to telescope.

lkhphuc:
> Also it seems floating windows don't use global value of winbar option so in previous pr floating windows didn't have winbar at all even without you adding the filetypes to disabled list

I wonder if this is desired behaviour in most case.

shadmansaleh:
> I wonder if this is desired behaviour in most case.

lkhphuc:
Another thing I notice is that this with this branch, my `showtabline` is set to 1, even though I have `vim.o.showtabline=0` in my config.

lkhphuc:
It fixes the issue. All seems good to me now.

shadmansaleh:
What kind of window layout do you have when this happens . Since you're saying it happens on key press do is it possible some small completion popups or similar are appearing ? Though we should already be ignoring popups.

lkhphuc:
I have figured out one reproducible case for the error above, it's due to the small popup window when using https://github.com/stevearc/dressing.nvim 
Everytime I trigger a rename function, I got the "not enough room" above.

shadmansaleh:
I'm not sure but it seems the original window is actually the one throwing error not the floating window . Which is odd because it does have enough space to draw the winbar . Do you think you can reproduce this without any other plugin ?

lkhphuc:
I use https://github.com/andymass/vim-matchup with this config to get a popup window display on the right like this for very long function: 
```
  use {"andymass/vim-matchup",
    config = function ()
      vim.g.matchup_matchparen_offscreen = {method= 'popup', position='cursor'}
    end
  }
```
Screenshot 2022-07-20 at 18 49 46

Now if I press anything, enter insert, command mode etc I will get this "not enough room" error. I will have to move the cursor elsewhere first.

shadmansaleh:
@lkhphuc I think a bug in neovim is causing the issue you're facing . I've opened an issue in neovim's repo https://github.com/neovim/neovim/issues/19464 lets what the devs think about it.

shadmansaleh:
I think this branch is quite stable . I can probably merge it already . I anyone want can take it for a spin and see if there are any inconsistencies.

willnorris:
This seems to be causing a problem with `set wildmode=list`... the command line completion list displays for just  a second, and then disappears.  Reverting to 8d956c1 fixes the behavior, so it definitely seems to be caused by this.

shadmansaleh:
@willnorris  fixed in 6fbc35b

PR #748, opened 2022-07-22
feat(buffer): add lualine_buffer_silence option

NavePnow:
Purpose: ignore the error when jump out of index
Currently I used global option to handle variable, but I would like to use options inside buffer to handle it.


shadmansaleh:
Instead of this add bang support to the command so `LualineBuffersJump!` ignores the error while regular `LualineBuffersJump` throws error . That's how vim commands usually work.

NavePnow:
Appreciate your help!

PR #759, opened 2022-07-27
feat(diagnostic): add an option to show diagnostics from all buffers

changyiyao:
Make it behave like visual studio code when using nvim_diagnostic or nvim_lsp diagnostic component


shadmansaleh:
This feature has been merged with #618 .I prefer to have it as separate source rather then an option because respecting this option will probably be hard for all the different kind of source that emit diagnostics

PR #769, opened 2022-07-30
Added support for import.nvim as extension

miversen33:
See title lol. 
I have added [`import.nvim`](https://github.com/miversen33/import.nvim/pull/6), but cannot merge it until the lualine extension has been accepted :)

Note: This extension includes an example of how to implement the [`on_click`](https://github.com/nvim-lualine/lualine.nvim/pull/742) functionality, I am unsure if that has been documented fully or if other plugins/extensions are currently using it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -821,6 +821,7 @@ extensions = {'quickfix'}
 - fern
 - fugitive
 - fzf
+- import
 - man
 - mundo
 - neo-tree
--- a/lua/lualine/extensions/import.lua
+++ b/lua/lualine/extensions/import.lua
@@ -1,0 +1,6 @@
+local M = {}
+
+M.sections = { lualine_a = {'filetype'}}
+M.filetypes = {'ImportManager'}
+
+return M


shadmansaleh:
> I have added https://github.com/miversen33/import.nvim/pull/6, but cannot merge it until the lualine extension has been accepted :)

You actually don't need to just like components and themes extensions are loaded from external sources too. So putting the file `lua/lualine/extensions/import.lua` in your repo is same as putting it here .

>  I am unsure if that has been documented fully

It's a documented lualine feature so it's not getting removed . but since it's only available on nvim-0.8 unless minimum requirement of your plugin is nightly you should probably add a version check.

miversen33:
Ahh my bad, I was reading [here](https://github.com/nvim-lualine/lualine.nvim#custom-extensions) and assumed that meant if you have a specific plugin/extension that you want to be included in lualine, to submit a PR for it.

> You actually don't need to just like components and themes extensions are loaded from external sources too. So putting the file lua/lualine/extensions/import.lua in your repo is same as putting it here .
Huh thats nifty! Well in that case, this PR is not needed :) 

Thanks!

PR #780, opened 2022-08-04
More user-friendly LualineBuffersJump

kioplato:
These two commits attempt to make LualineBuffersJump more user-friendly and provide more functionality by returning a boolean value depending on successful or unsuccessful buffer jumps.

It passes the CI: https://github.com/kioplato/lualine.nvim/actions/runs/2797484866


shadmansaleh:
Is this to prevent `LualineBuffersJump` from throwing errors ? If you're running it as command you won't get the return value anyway and if you're calling it as a function from lua you can get this info just by calling with `pcall`.

If it's to prevent the error . I'd prefer to add `<bang>` support instead so  regular `LualineBuffersJump` throws error but `LualineBuffersJump!` ignores the error.

kioplato:
I didn't accurately describe my changes. I'm sorry.

The first commit is to prevent `LualineBuffersJump` from throwing errors when attempting to jump to a non-existent buffer number. Since `LualineBuffersJump` uses `buffer_jump` to achieve this, I replaced the error call with `echomsg` inside `buffer_jump`.

However, `<bang>` support could solve this, as you suggested, in the same way `q!` ignores the errors (is this the case?). I tried to implement it but the documentation is lacking. I only found `-bang` attribute for `command` but there's nowhere any description about what this actually does. Any pointers on this on how should I proceed?

The second commit makes `buffer_jump` return a boolean value, not `LualineBufferJump`. `pcall` could achieve the same thing in Lua and didn't know about its existence. This renders my second commit unnecessary.

shadmansaleh:
Create the command like this.
```vim
command! -nargs=1 -bang LualineBuffersJump call v:lua.require'lualine.components.buffers'.buffer_jump(<f-args>, <bang>)
```

Add 2nd arg to buffer_jump and check if it equals to '!' to know if command was called with bang. When it's called with bang just exit the funciton instead of calling error().

kioplato:
I implemented bang support. 

The second argument needs to be surrounded in quotation marks because when `<bang>` is replaced by `!`, a syntax error occurs.
```
command! -nargs=1 -bang LualineBuffersJump call v:lua.require'lualine.components.buffers'.buffer_jump(<f-args>, "<bang>")
```

shadmansaleh (approved):
lgtm. Thanks for your contribution:)

PR #782, opened 2022-08-05
Extension for the Drex directory explorer plugin

f1rstlady:
Taken from [here](https://github.com/TheBlob42/drex.nvim/wiki/Configuration#lualine-setup).

### DIFF
--- a/README.md
+++ b/README.md
@@ -818,6 +818,7 @@ extensions = {'quickfix'}
 
 - aerial
 - chadtree
+- drex
 - fern
 - fugitive
 - fzf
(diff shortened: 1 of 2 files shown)


shadmansaleh:
Can you open a pr in drex.nvim putting the same `lua/lualine/extensions/drex.lua` in the same location in their repo and see if they want to support a lualine extension ?

It'll be same whether it placed here or their . Since this is calling their internal functions it's better if it's on their repo that way they'll know if they even introduce breaking changes to this .

f1rstlady:
Good idea! [Here](https://github.com/TheBlob42/drex.nvim/pull/26) it is.

f1rstlady:
The pr got accepted, i.e. the extension has been added to the drex repo. Hence I will close this one.

PR #785, opened 2022-08-06
[filename] support detect new file status

hiberabyss:
Add `[New]` symbol for new file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ sections = {
     {
       'filename',
       file_status = true,      -- Displays file status (readonly status, modified status)
-      newfile_status = false   -- Whether show new file status, default disabled
+      newfile_status = false   -- Display new file status (new file means no write after created)
       path = 0,                -- 0: Just the filename
                                -- 1: Relative path
                                -- 2: Absolute path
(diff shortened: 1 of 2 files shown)


shadmansaleh:
In what case does unnamed and newfile differ ?

hiberabyss:
> In what case does unnamed and newfile differ ?

For example, when edit a new file via `:e /tmp/newfile.txt`, the status will be `[New]` before `:w`.

shadmansaleh:
Ok can you document this option with other filename options in README.md ?

hiberabyss:
> Ok can you document this option with other filename options in README.md ?

Readme updated, please help to review.

README.md
```diff
@@ -593,6 +593,7 @@ sections = {
     {
       'filename',
       file_status = true,      -- Displays file status (readonly status, modified status)
+      newfile_status = false   -- Whether show new file status, default disabled
```
shadmansaleh, line 596: I think you should define what you mean by 'new file status'. Also default disabled isn't necessary as =false already states that

hiberabyss: Updated.

shadmansaleh (approved):
lgtm .

PR #797, opened 2022-08-09
fix: Remove base16-specific actions in "auto" theme

### LINKED ISSUE #796: Bug: My base16 theme is overridden even though I'm using the "auto" theme
### Self Checks

- [X] I'm using the latest lualine.
- [X] I didn't find the issue in exsisting issues or prs.

In your `nvim` config, add this line:
```vim
let g:colors_name = "base16-<theme_name>"
```

If [RRethy/nvim-base16](https://github.com/RRethy/nvim-base16) is installed, and `<theme_name>` corresponds to a theme defined in the plugin, there should be no issues.

If it is installed but `<theme_name>` does not correspond to an existing theme, Lualine will complain.

If it is not installed, Lualine will complain.

I found this issue because I am using [flavours](https://github.com/Misterio77/flavours) to have the same base16 theme in all my applications. I have a `flavours.vim` file in my `colors/` directory that takes the original [base16-vim](https://github.com/chriskempson/base16-vim/blob/master/templates/default.mustache) template and fills it with the right colors.
You can see that [it sets the theme name to `base16-<theme_name>`](https://github.com/chriskempson/base16-vim/blob/master/templates/default.mustache#L148) which causes the lualine issue.

### Expected behaviour
My theme gets loaded regardless of its name, like it did before https://github.com/nvim-lualine/lualine.nvim/pull/597 was merged.

### Actual behaviour
I get a warning to go to the lualine notices, and when I do I see the message:
```
theme(base16): nvim-base16 is not currently present in your runtimepath,
make sure it is properly installed,
fallback to default colors.
```

```vim
call plug#begin("<test_dir>/.local/share/nvim/plugged")
  Plug 'nvim-lualine/lualine.nvim'
call plug#end()

" With this line you'll get the LualineNotices thing
let g:colors_name = "base16-test_theme"
" With this line you will not get it
" let g:colors_name = "test_theme"


lua << END
require'lualine'.setup {
  options = {
    -- Auto theme should just default to whatever colors are found
    theme = 'auto'
  }
}
END

" Instructions:
" -------------------------------------------------------------
" Load this config with nvim_isolated_conf.sh -l <test_dir>
" Remember to run :PlugInstall after changing plugin section
```

augustebaum:
Closes #796.

The test suite seems broken:
```
Starting...Scheduling: tests//spec/config_spec.lua
Scheduling: tests//spec/component_spec.lua
Scheduling: tests//spec/utils_spec.lua
Scheduling: tests//spec/lualine_spec.lua
make: *** [test] Error 1
```
That's all I get.

### DIFF
--- a/lua/lualine/themes/auto.lua
+++ b/lua/lualine/themes/auto.lua
@@ -5,11 +5,6 @@ local loader = require('lualine.utils.loader')
 
 local color_name = vim.g.colors_name
 if color_name then
-  -- All base16 colorschemes share the same theme
-  if 'base16' == color_name:sub(1, 6) then
-    color_name = 'base16'
-  end
-
   -- Check if there's a theme for current colorscheme
   -- If there is load that instead of generating a new one
   local ok, theme = pcall(loader.load_theme, color_name)


shadmansaleh:
Instead of removing base-16 auto detection lets add an additional check to see if f [RRethy/nvim-base16](https://github.com/RRethy/nvim-base16) is installed . You should find a check for that in base-16 theme in lualine. I think it'd still be a good default to use base-16 theme when f [RRethy/nvim-base16](https://github.com/RRethy/nvim-base16) is installed.

PR #810, opened 2022-08-22
fix component(filetype) proper icon based on filetype

### LINKED ISSUE #578: Bug: lualine doesn't show filetype icon for files without extension

zappolowski:
With https://github.com/kyazdani42/nvim-web-devicons/pull/125 a filetype based lookup was implemented upstream which can be used to get an icon when a file doesn't have an extension.

fixes #578

### DIFF
--- a/lua/lualine/components/filetype.lua
+++ b/lua/lualine/components/filetype.lua
@@ -31,9 +31,7 @@ function M:apply_icon()
   local icon, icon_highlight_group
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
-    local f_name, f_extension = vim.fn.expand('%:t'), vim.fn.expand('%:e')
-    f_extension = f_extension ~= '' and f_extension or vim.bo.filetype
-    icon, icon_highlight_group = devicons.get_icon(f_name, f_extension)
+    icon, icon_highlight_group = devicons.get_icon(vim.fn.expand('%:t'))
 
     if icon == nil and icon_highlight_group == nil then
       icon = ''
--- a/lua/lualine/components/filetype.lua
+++ b/lua/lualine/components/filetype.lua
@@ -32,6 +32,9 @@ function M:apply_icon()
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
     icon, icon_highlight_group = devicons.get_icon(vim.fn.expand('%:t'))
+    if icon == nil then
+      icon, icon_highlight_group = devicons.get_icon_by_filetype(vim.bo.filetype)
+    end
 
     if icon == nil and icon_highlight_group == nil then
       icon = ''
--- a/tests/spec/component_spec.lua
+++ b/tests/spec/component_spec.lua
@@ -424,6 +424,42 @@ describe('Filetype component', function()
     hl.create_component_highlight_group:revert()
     vim.fn.expand:revert()
   end)
+
+  it('uses filetype lookup when file has no extension', function()
+    stub(vim.fn, 'expand')
+    vim.fn.expand.on_call_with('%:t').returns('test')
+
+    local hl = require('lualine.highlight')
+    stub(hl, 'create_component_highlight_group')
+
+    local utils = require('lualine.utils.utils')
+    stub(utils, 'extract_highlight_colors')
+
+    local devicons = require('nvim-web-devicons')
+    stub(devicons, 'get_icon')
+    devicons.get_icon.on_call_with('test').returns(nil)
+    stub(devicons, 'get_icon_by_filetype')
+    devicons.get_icon_by_filetype.on_call_with('lua').returns('*', 'test_highlight_group')
+
+    local opts = build_component_opts {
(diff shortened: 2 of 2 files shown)


shadmansaleh:
That's a good news. Thanks for the pr.

tests/spec/component_spec.lua
```diff
@@ -258,14 +258,21 @@ end)
 
 describe('Filetype component', function()
   local filetype
+  local orig_expand
 
   before_each(function()
     filetype = vim.bo.filetype
     vim.bo.filetype = 'lua'
+    orig_expand = vim.fn.expand
+    vim.fn.expand = function(fmt)
```
shadmansaleh, line 267: You can use `stubs` from testing framework instead of overwriting the function yourself. There are examples in the test suit and you can take a look at the lua busted docs for more info.

lua/lualine/components/filetype.lua
```diff
@@ -31,9 +31,12 @@ function M:apply_icon()
   local icon, icon_highlight_group
   local ok, devicons = pcall(require, 'nvim-web-devicons')
   if ok then
-    local f_name, f_extension = vim.fn.expand('%:t'), vim.fn.expand('%:e')
-    f_extension = f_extension ~= '' and f_extension or vim.bo.filetype
-    icon, icon_highlight_group = devicons.get_icon(f_name, f_extension)
```
shadmansaleh, line 36: Do we need to remove these ? Can't we use get_icon_by_filetype when get_icon returns nil as fallback?

tests/spec/component_spec.lua
zappolowski: While at it shall I use `stub` for accessing `nvim-web-devicons` as well? If yes, should I put it on another (refactoring) PR to focus on file type icons in this one?

lua/lualine/components/filetype.lua
zappolowski: This might work ... playing around with `get_icon` this might be a valid simplifcation:
```lua
icon, icon_highlight_group = devicons.get_icon(vim.fn.expand('%:t'))
if icon == nil then
  icon, icon_highlight_group = devicons.get_icon_by_filetype(vim.bo.filetype)
end
```
as `get_icon` determines the extension from the name if no extension (2nd argument) is given.

tests/spec/component_spec.lua
shadmansaleh: Sure if you feel like it you can do that too.

lua/lualine/components/filetype.lua
shadmansaleh: That looks cleaner . You can change it to this.

I was thinking about keeping the older filetype based check because the webdeviconcs pr stated the list added wasn't exhaustive . But now looking at it I think in the cases where older filetype check would matter ar already supported by `get_icon_by_filetype` . So we can remove that.

shadmansaleh:
I don't exactly remember was there something blocking it's merge ? where were we ? Does this version work properly ?

zappolowski:
I've kept the branch up to date and tests are still passing. As far as I can tell, the discussions are resolved.

PR #812, opened 2022-08-24
fix component(filename) add separator between filename and icons

### LINKED ISSUE #794: Filename component: symbols space
So hey........... .-.

The default vim statusline adds one space between the filename and the first symbol

default: `file.xyz [+][RO]`
lualine:  `file.xyz[+][RO]`

zappolowski:
This makes it behave like the default statusline by have a space between the filename and the following icons (but not between the icons).

While at it, also added new test case to ensure that representation is correct, when both states (modified and readonly) are active at the same time.

fixes #794

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -80,19 +80,21 @@ M.update_status = function(self)
     end
   end
 
+  local symbols = {}
   if self.options.file_status then
     if vim.bo.modified then
-      data = data .. self.options.symbols.modified
+      table.insert(symbols, self.options.symbols.modified)
     end
     if vim.bo.modifiable == false or vim.bo.readonly == true then
-      data = data .. self.options.symbols.readonly
+      table.insert(symbols, self.options.symbols.readonly)
     end
   end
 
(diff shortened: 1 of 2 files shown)


lua/lualine/components/filename.lua
```diff
@@ -80,19 +80,21 @@ M.update_status = function(self)
...
     end
   end
 
   if self.options.newfile_status and is_new_file() then
-    data = data .. self.options.symbols.newfile
+    table.insert(symbols, self.options.symbols.newfile)
   end
-  return data
+
+  return data .. (#symbols > 0 and ' ' .. vim.fn.join(symbols, '') or '')
```
shadmansaleh, line 97: Use `table.concat(symbols)` instead of `vim.fn.join`. That way we won't have type conversion back and forth.

Otherwise looks good to me

zappolowski:
One idea came up: should we make the separator configurable?

shadmansaleh:
> One idea came up: should we make the separator configurable?

Doesn't seem very useful to be honest. Lets leave it for now.

Issue #820, opened 2022-08-30
Bug: E539: Illegal character in filename


liuerfire:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem

set `path` to any value (1, 2 or 3) except the default 0 in filename section.

```
require('lualine').setup {
  sections = {
    lualine_c = {
      { 'filename', path = 3, },
    },
  },
}
```
Install https://github.com/mfussenegger/nvim-jdtls and jump to the definition of a class. 

### Expected behaviour
no errors

### Actual behaviour

![image](https://user-images.githubusercontent.com/745867/187383416-b9d95fa1-ea6f-49ed-b82f-93b5921c540d.png)

### Aditional information

The full path of this file is:

![image](https://user-images.githubusercontent.com/745867/187383259-0fb3c477-0ab6-46f6-b214-a875f1f326ed.png)

liuerfire:
Maybe it's jdtls's bug, which shouldn't escape '/' to %5C?

zeitchef:
I'm getting the exact same error

ramanenka:
I've got the same error and here is how I've solved it:
```lua
lualine_c = {
  function()
    local fn = vim.fn.expand('%:~:.')
    if vim.startswith(fn, "jdt://") then
      fn = string.sub(fn, 0, string.find(fn, "?") - 1)
    end
    if fn == '' then
      fn = '[No Name]'
    end
    if vim.bo.modified then
      fn = fn .. ' [+]'
    end
    if vim.bo.modifiable == false or vim.bo.readonly == true then
      fn = fn .. ' [-]'
    end
    local tfn = vim.fn.expand('%')
    if tfn ~= '' and vim.bo.buftype == '' and vim.fn.filereadable(tfn) == 0 then
      fn = fn .. ' [New]'
    end
    return fn
  end
},
```
For some reason writing the same logic in an `fmt` function didn't prevent the error from happening.

kmoschcau:
@ramanenka I figured out why `fmt` isn't working. What you get in that function is probably already shortened and looks like this:
`j//c/j/j/N/%/h/k/.g/c/m/f/j/j/2/8/j/g/m/=/o/%%5C/WEB-INF%%5C/lib=/%%3Cjavax.persistence(NoResultException.class`
So your prefix check doesn't even work in there.

1Jo1:
[...]

PR #821, opened 2022-08-30
fix: remove url-encoding chars in filename

liuerfire:
fixes https://github.com/nvim-lualine/lualine.nvim/issues/820

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -62,6 +62,7 @@ M.update_status = function(self)
     data = vim.fn.expand('%:t')
   end
 
+  data = data:gsub("%%5C", "")
   data = modules.utils.stl_escape(data)
 
   if data == '' then


shadmansaleh:
That's strange % in there should be escaped by `utils.stl_escape()` in following line it it shouldn't throw error.

liuerfire:
I'm not sure this should be fixed by lualine or by jdtls :cry: But this commit did fix #820.

Feel free to close this MR 'cause I don't think this is a good solution either 😄

1Jo1:
at least this annoying bug is fixed :)

why not just like escaping `%`
```
function M.stl_escape(str)
  if type(str) ~= 'string' then
    return str
  end
  return str:gsub('%%', '%%%%'):gsub('%%', '')
end
```

1Jo1:
> That's strange % in there should be escaped by `utils.stl_escape()` in following line it it shouldn't throw error.

 `str:gsub('%%', '%%%%'):` it doesn't escaped '%', it actually % will be added

shadmansaleh:
> str:gsub('%%', '%%%%') it doesn't escaped %, it actually % will be added

In statusline %% is displayed as % just like how in most regular strings \\ is displayed as \

1Jo1:
> > str:gsub('%%', '%%%%') it doesn't escaped %, it actually % will be added
> 
> In statusline %% is displayed as % just like how in most regular strings \ is displayed as \

sorry I'm really confused, I just tried lua in terminal:
```
Lua 5.4.4  Copyright (C) 1994-2022 Lua.org, PUC-Rio
> 
> string.gsub('%%', '%%', '%%%%')
%%%%  2
> 
```
according to you, result should be `%` right? 
maybe I'm just not familiar with Lua/Neovim or I miss something else?

1Jo1:
what do you think? @shadmansaleh

PR #826, opened 2022-09-04
feat: Add One theme

laomafeima:
It's a combination of onelight and onedark. It loads either of
them based you your `background` option.


shadmansaleh:
I don't see value in adding it. Also as stated in [CONTRIBUTING.md](https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md#contributing-to-lualinenvim) we're currently not accepting themes like these. You can put it in your ~/.config/nvim/lua/lualine/themes directory and use it as a regular theme. lualine will load themes placed in &rtp/lua/lualine/themes even if it's not from this repo.

PR #827, opened 2022-09-04
fix(extensions): disable winbar

NFrid:
Currently if we have winbar configured it will exist on the loaded extensions' buffers as well. I guess they shouldn't...

### DIFF
--- a/lua/lualine/extensions/aerial.lua
+++ b/lua/lualine/extensions/aerial.lua
@@ -4,6 +4,9 @@ local M = {}
 
 M.sections = { lualine_a = { 'filetype' } }
 
+M.winbar = {}
+M.inactive_winbar = {}
+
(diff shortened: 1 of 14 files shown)


shadmansaleh:
I'm not sure if we should disable winbar on builtin extensions. We don't disable tabline just change the statusline configuration to have different config. From that point it makes sense to have default winbar. Still if users wants winbar disabled on extensions they can modify the extension table before passing it to lualine setup.

NFrid:
there's no way to define an extension for a filetype while still using a built-in for that same filetype :(
we need to "join" them together in `get_extension_sections`

well I tried to solve that but found out that I don't really like a built-in extensions anyways and made just a winbar disabler one for the filetypes I needed

so thank you I guess for helping me figure it out :)

PR #828, opened 2022-09-04
Add overseer extension.

dcordb:
This PR adds a simple [overseer.nvim](https://github.com/stevearc/overseer.nvim) extension.

### DIFF
--- a/lua/lualine/extensions/overseer.lua
+++ b/lua/lualine/extensions/overseer.lua
@@ -1,0 +1,12 @@
+-- Copyright (c) 2020-2021 hoob3rt
+-- MIT license, see LICENSE for more details.
+
+local M = {}
+
+M.sections = {
+  lualine_a = { function() return "OverseerList" end },
+}
+
+M.filetypes = { "OverseerList" }
+
+return M
--- a/lua/lualine/extensions/overseer.lua
+++ b/lua/lualine/extensions/overseer.lua
@@ -4,9 +4,13 @@
 local M = {}
 
 M.sections = {
-  lualine_a = { function() return "OverseerList" end },
+  lualine_a = {
+    function()
+      return 'OverseerList'
+    end,
+  },
 }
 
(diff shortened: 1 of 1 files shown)


dcordb:
cc @stevearc @dzfrias

stevearc:
While you're adding overseer configuration to the lualine repo, this would be a good place to ask if it makes more sense for [this overseer component](https://github.com/stevearc/overseer.nvim/blob/master/lua/lualine/components/overseer.lua) to live in the overseer repo or in the lualine repo. Thoughts?

dcordb:
@stevearc Yes I think it makes the more sense to have that component on lualine repo. For example, if someone wants to have overseer integrated with another status line plugin, it does not make sense to add it to the overseer repo, you would eventually end up with many integrations in there. For this, I think the implementation of each component for each status line should be in the status line repo.

shadmansaleh:
> While you're adding overseer configuration to the lualine repo, this would be a good place to ask if it makes more sense for [this overseer component](https://github.com/stevearc/overseer.nvim/blob/master/lua/lualine/components/overseer.lua) to live in the overseer repo or in the lualine repo. Thoughts?

The issue with overseer component is it's highly dependent on overseer itself and uses it's internal modules. And I just won't know if breaking changes are made to them. Generally I try to avoid codes that depend on other plugins directly. Also plugin specific component like these only makes sense when the plugin is actually installed. But adding it to this repo would mean it'd be available even when overseer isn't installed. This was the reason why we have 3rd party components.

> if someone wants to have overseer integrated with another status line plugin, it does not make sense to add it to the overseer repo, you would eventually end up with many integrations in there

In this case I think what most plugins does is provide a common function returning a string as statusline integration. Since basically all statusline plugins render these functions in the same manner that's the easiest way to go. What providing is lualine specific component does is it allows better rendering specially makes handling colors much easier. In the end it's upto the plugin authors to decide how they'll provide support.

PR #834, opened 2022-09-10
Support extracting `special` color from highlight group

antoineco:
Besides `background` and `foreground`, (Neo)vim highlight groups have a `special` attribute that can be used for setting the color of underlines. (See `:h nvim_set_hl()`)

I happen to be working with a color scheme which sets the highlight style of `DiagnosticHint`, etc. to a colored `undercurl` instead of a foreground color:

```lua
c = vim.api.nvim_get_hl_by_name('DiagnosticHint', true)

for attr in pairs(c) do
  print(attr)
end

--[[ print
special
undercurl
--]]
```

This PR provides access to this attribute via the `extract_highlight_colors()` function.

Example:

```lua
warn = { fg = utils.extract_color_from_hllist('sp', {'DiagnosticWarn'}, '#ffa500') }
```


shadmansaleh:
merged with #836

PR #836, opened 2022-09-11
Support extracting a fallback color from a highlight group

### LINKED ISSUE #681: Bug: Colour for diagnostics mismatch
### Self Checks

- [x] I didn't find the issue in exsisting issues or prs.

Install lualine with packer and call lualine.setup. 
The colors for diagnostics seem to be incorrect
image
image
image

Taken from lualine source, it seems that it should respect those hi groups.
```Lua
function M.apply_default_colors(opts)
  local default_diagnostics_color = {
    error = {
      fg = utils.extract_color_from_hllist(
        'fg',
        { 'DiagnosticError', 'LspDiagnosticsDefaultError', 'DiffDelete' },
        '#e32636'
      ),
    },
    warn = { 
      fg = utils.extract_color_from_hllist(
        'fg',
        { 'DiagnosticWarn', 'LspDiagnosticsDefaultWarning', 'DiffText' },
        '#ffa500' 
      ),
    },
    info = {
      fg = utils.extract_color_from_hllist(
        'fg',
        { 'DiagnosticInfo', 'LspDiagnosticsDefaultInformation', 'Normal' },
        '#ffffff'
      ),
    },
    hint = {
      fg = utils.extract_color_from_hllist(
        'fg',
        { 'DiagnosticHint', 'LspDiagnosticsDefaultHint', 'DiffChange' },
        '#273faf'
      ),
    },
  }
  opts.diagnostics_color = vim.tbl_deep_extend('keep', opts.diagnostics_color or {}, default_diagnostics_color)
end
```

### Expected behaviour
Color should be respected

### Actual behaviour
Instead lualine renders weird dark grey.

No configuration has been applied just calling lualine.setup()

### Aditional information
using everforest theme.

antoineco:
> **Note**
> Marking as draft because requires #834 to properly handle the `sp` highlight scope.
> Right now this PR contains both commits to prevent tests from failing.

---

### What

Allows passing a list of scopes to `extract_color_from_hllist()`, where each scope is tried sequentially until a matching color is found in the highlight group (return on first match).

The change is fully backwards compatible; if the `scope` parameter is a string, the function can handle that and only tries that one scope.

### Why

My main motivation is to allow a sensible fallback to the <ins>underline</ins> color in `config.(default_)diagnostics_color` in case any of the `DiagnosticError`, `DiagnosticWarn`, `DiagnosticHint` or `DiagnosticInfo` highlight groups do not have a foreground color set.

In many IDEs, it is a [common practice](https://user-images.githubusercontent.com/7727148/38758132-7284e58a-3f24-11e8-8e39-627e6916e9ec.PNG) to highlight diagnostics by underlining the context, instead of changing its foreground color. Due to Vim's legacy of supporting terminals without capabilities such as _italics_, **bold** or <ins>underline</ins>, this has never become a widespread pattern in our favourite text editor. Nevertheless, some (fairly popular) color schemes — such as [everforest](https://github.com/sainnhe/everforest), [gruvbox-material](https://github.com/sainnhe/gruvbox-material) or [sonokai](https://github.com/sainnhe/sonokai) — do apply underline styles to these highlight groups, and right now Lualine can't handle this practice very well:

image
image

With this change, Lualine is now able to pick a sensible color for diagnostic signs, when diagnostic highlight groups use the underline highlight attribute without a foreground color:

image

Also, fixes #681

### DIFF
--- a/lua/lualine/components/diagnostics/config.lua
+++ b/lua/lualine/components/diagnostics/config.lua
@@ -26,28 +26,28 @@ function M.apply_default_colors(opts)
   local default_diagnostics_color = {
     error = {
       fg = utils.extract_color_from_hllist(
-        'fg',
+        { 'fg', 'sp' },
         { 'DiagnosticError', 'LspDiagnosticsDefaultError', 'DiffDelete' },
         '#e32636'
       ),
     },
     warn = {
       fg = utils.extract_color_from_hllist(
-        'fg',
+        { 'fg', 'sp' },
         { 'DiagnosticWarn', 'LspDiagnosticsDefaultWarning', 'DiffText' },
         '#ffa500'
       ),
     },
     info = {
       fg = utils.extract_color_from_hllist(
-        'fg',
+        { 'fg', 'sp' },
         { 'DiagnosticInfo', 'LspDiagnosticsDefaultInformation', 'Normal' },
         '#ffffff'
       ),
     },
     hint = {
       fg = utils.extract_color_from_hllist(
-        'fg',
+        { 'fg', 'sp' },
         { 'DiagnosticHint', 'LspDiagnosticsDefaultHint', 'DiffChange' },
         '#273faf'
       ),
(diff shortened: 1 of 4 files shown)


shadmansaleh:
Looks good to me. Why did you mark it as draft ?

Thanks for letting me know what's special is used for. I definitely misunderstood this highlight property till now.

antoineco:
Mainly because I though it would be easier to review in two parts. #834 first, then this.
If you're fine merging all at once as one feature, let's do it :)

Thanks for the review

shadmansaleh:
> If you're fine merging all at once as one feature, let's do it :)

Sure these 2 are pretty close and not too big either we can just merge without squashing .

PR #839, opened 2022-09-12
Changing the name of the shorting_target attr

felix-khlmnn:


felix-khlmnn:
I'm not really sure how to set up the checks to account for new names etc. Help would be greatly appreciated.

Issue #860, opened 2022-10-07
fzf-lua extension: filetype not recognised


gennaro-tedesco:
### Motivation

More than a feature a request for advice: I would like to write an extension for [fzf-lua](https://github.com/ibhagwan/fzf-lua) and I have seen that a [current fzf](https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/extensions/fzf.lua) extension already exists: as far as I understand such extension solely checks the filetype of the current window and returns a string component in one of the lualine sections. Whilst this works with fzf.vim, it fails with fzf-lua _although the filetype is the same_ (see [cross post](https://github.com/ibhagwan/fzf-lua/issues/537#issuecomment-1271860920) discussion). By failing I mean that nothing is being shown although explicitly invoked.

Perhaps my understanding of the extension mechanics is incorrect, and besides the filetype other checks are performed? Could you provide more explanation on the topic? :)

shadmansaleh:
Your assumption is right it's just filetype check.

It's not working for fzf-lua because fzf-lua gets rendered on a floating windows and floating windows don't have a  statusline. If set inactive_statusline in extension it'll appear all the other windows apart from the fzf one. If you use globalstatus it'll be rendered properly as with gloalstatusline we can render statusline for floating windows too in the same statusline spot. I think you can  try setting the winbar in the extension and see if you like it. floating windows can have winbars but not statusline it's a limitation in neovim currently.

gennaro-tedesco:
Thank you for your thorough explanation: I will experiment with inactive sections and winbar and see what comes comes out of it! :)

PR #863, opened 2022-10-08
Add fzf-lua extension

gennaro-tedesco:
This PR adds a statusline extension for [fzf-lua](https://github.com/ibhagwan/fzf-lua), in particular it addresses the discussion we had in https://github.com/nvim-lualine/lualine.nvim/issues/860.

See [cross-post](https://github.com/ibhagwan/fzf-lua/issues/537) for additional reference.

Below is a screen recording showing what it looks like in action :)

https://user-images.githubusercontent.com/15387611/194726388-8ee88056-6d15-4e78-8e22-a6d3b989590d.mov

### DIFF
--- a/lua/lualine/extensions/fzf.lua
+++ b/lua/lualine/extensions/fzf.lua
@@ -1,5 +1,34 @@
--- Copyright (c) 2020-2021 hoob3rt
--- MIT license, see LICENSE for more details.
+--[[
+lualine extension for fzf filetypes:
+works with both https://github.com/junegunn/fzf.vim and https://github.com/ibhagwan/fzf-lua
+
+-- fzf-lua must be set-up in split mode
+]]
+
+local fzf_lua, _ = pcall(require, 'fzf-lua')
+
+local function fzf_picker()
+  if not fzf_lua then
+    return ''
+  end
+
+  local info_string = vim.inspect(require('fzf-lua').get_info()['fnc'])
+  return info_string:gsub('"', '')
+end
+
+local function fzf_element()
+  if not fzf_lua then
+    return ''
+  end
+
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Since they are both working on fzf filetype why not make fzf extension render like this when fzf-lua in installed instead of fzf.vim?

gennaro-tedesco:
I have included the logic for `fzf-lua` in the original extension file and tested with both `fzf.vim` and `fzf-lua`.

Moreover I have rebased the latest commits onto the initial ones so that you now only see one commit with the latest changes.

nyngwang:
@gennaro-tedesco In my case nothing changes(the lualine is not attached to fzf-lua floating windows) with the following settings:

```lua
config = function ()
  local navic = require('nvim-navic')
  require('lualine').setup {
    options = {
      theme = 'palenight', -- Purple
    },
    sections = {
      lualine_c = { 'filename', { navic.get_location, cond = navic.is_available } },
      lualine_x = { 'encoding', 'filetype' },
    },
    extensions = { 'fzf' },
  }
end
```

gennaro-tedesco:
The statuslines are only available for split views, not for floating windows: see [this neovim](https://github.com/neovim/neovim/issues/16753) issue: it is being worked on and contributed at the moment.

By the way the same happens for the standard fzf.vim too: unless you use it in a split (horizontal or vertical) you will not see the corresponding lualine element. I have left a comment in the source code to remark in fact that the lualine elements will only be displayed in those cases.

nyngwang:
> The statuslines are only available for split views [...]

Oops! I missed the point :P Now I got what you meant. Thanks for your PR still, it makes users know which commands they have called and it's really helpful IMOO.

PR #875, opened 2022-10-20
Theme: Powerline_Cyan added

edr3x:
### Powerline_Cyan

- This is my fav tmux theme so ported that theme in lualine to make visually uniform.


shadmansaleh:
As stated in [CONTRIBUTING.md](https://github.com/nvim-lualine/lualine.nvim/blob/master/CONTRIBUTING.md#contributing-to-lualinenvim) we're currently not accepting themes like these. You can put it in your ~/.config/nvim/lua/lualine/themes directory and use it as a regular theme or if you want you can release it as a separate repo. lualine will load themes placed in &rtp/lua/lualine/themes even if it's not from this repo.

PR #880, opened 2022-10-24
Fixes bug with string interpolation from 3rd party functions

halfdan:
This fixes a bug I experienced when using lualine with lsp-status and Elixir.

```elixir
  defp build_slack_message(%{asset: asset, project: project, parent: parent} = comment) do
    timecode = comment |> timestamp(parent) |> Core.Timecode.generate(asset.fps)
    user = Comments.author(comment)
 ```

A function like the above would return the function head as a string and result in a Lua formatting error "E121: variable asset not defined". By properly escaping the % sign the bug is fixed.

### DIFF
--- a/lua/lualine/components/special/eval_func_component.lua
+++ b/lua/lualine/components/special/eval_func_component.lua
@@ -20,6 +20,8 @@ function M:update_status()
       status = M.vim_function(component)
     end
   end
+  
+  status = string.gsub(status, "%%", "%%%%")
   return status
 end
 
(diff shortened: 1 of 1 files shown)


shadmansaleh:
This won't be accepted. This will break all kinds of statusline formatters being emitted by function including highlight specifiers. Basically with this displaying anything but simple strings will be impossible from a function.

So the function is responsible for returning proper output for statusline. So the issue you're facing will have to be fixed by lsp-status so they don't return unescaped % directly from buffer. Or you can have a wrapper function as part of your config that escapes the output of lsp-status specifically.

PR #884, opened 2022-10-31
`padding` as a function and show separators for empty components

rbrugo:
This PR adds two small features:
- you can now assign to `padding` a function instead of a value
- a new option, `show_separators_if_empty`, defaulted to `false` to preserve retro-compatibility. When it is set to `true` on a component, draws only its separators (as if it was folded) instead of skipping it.

For example, in a setup with three components, if `show_separators_if_empty = false` (default):
image
image

And if `show_separators_if_empty = true`:
image
image


lua/lualine/component.lua
```diff
@@ -274,14 +283,22 @@ function M:draw(default_highlight, is_focused)
...
-    self:apply_on_click()
-    self:apply_highlights(default_highlight)
-    self:apply_section_separators()
-    self:apply_separator()
+  local show_separators_if_empty = self.options.show_separators_if_empty
+  if type(show_separators_if_empty) == 'function' then
+    show_separators_if_empty = show_separators_if_empty()
+  end
+  if type(status) == 'string' then
+    if #status > 0 or show_separators_if_empty then
```
shadmansaleh, line 291: Add this condition to the previous if condition. We don't need to increase the indentation level here since there's no else case.

```lua
if type(status) == 'string' and (#status > 0 or show_separators_if_empty) then
```

lua/lualine/component.lua
```diff
@@ -274,14 +283,22 @@ function M:draw(default_highlight, is_focused)
...
+      self.status = status
+      if #status > 0 then
```
shadmansaleh, line 293: We checked the status is larger then 0 while entering this block so this condition is redundent.

shadmansaleh (changes requested):

lua/lualine/component.lua
rbrugo: If you're talking about line 293, the condition is not redundant since in 291 we tested if the length is 0 _or_ the flag is `true` - we may arrive to that line even if the line length is, in fact, 0 (and that's the whole point of the feature, being able to render the separator even if the status length is equal to 0).
If you prefer I could save the length in a variable to avoid computing it two times?

shadmansaleh: > we may arrive to that line even if the line length is, in fact, 0

You're right. I missed that.

> If you prefer I could save the length in a variable to avoid computing it two times

That won't be necessary.

rbrugo:
Is there anything else I should change?

lua/lualine/component.lua
```diff
@@ -274,11 +283,17 @@ function M:draw(default_highlight, is_focused)
   if self.options.fmt then
     status = self.options.fmt(status or '')
   end
-  if type(status) == 'string' and #status > 0 then
+  local show_separators_if_empty = self.options.show_separators_if_empty
+  if type(show_separators_if_empty) == 'function' then
```
shadmansaleh, line 287: when is it useful to have this as a function? It doesn't seem like something that will change much from the setup call. And if you need to check a condition to set it you can just use that function to get the value during setup. like

```lua
require'lualine'.setup {
  options = {
    show_separators_if_empty = some_condition_check()
   }
}
```

rbrugo: You're not wrong. It was just so easy to check for that case that I went for it.
I can remove that if you want; in that case, do you prefer the `'function'` case to be silently ignored or should I `print` an error to the user?

shadmansaleh: ignore it. function is implicitly true.

rbrugo:
Any news on this?

PR #905, opened 2022-11-30
Protect call inside `set` to avoid `dressing` winbar error

thedaviddelta:
The recent winbar feature produces an error while using both the `lualine` and the `dressing` packages.
This is because when `dressing` input creates a virtual window, `lualine` tries to set a value to its empty winbar.
It cannot be filtered out in any way as this `set` seems to happen before the filetype is even set for the new window.
The only way I found to solve this issue is to wrap the `nvim_win_set_option` with a protected call, so it doesn't report an error and freeze the execution for a second.
I don't think there should be a problem for this, as this error is not handled anyway, only reported to the user, and it doesn't actually break the mentioned window.

### DIFF
--- a/lua/lualine/utils/nvim_opts.lua
+++ b/lua/lualine/utils/nvim_opts.lua
@@ -74,7 +74,7 @@ function M.set(name, val, opts)
     set_opt(name, val, function(nm)
       return vim.api.nvim_win_get_option(opts.window, nm)
     end, function(nm, vl)
-      vim.api.nvim_win_set_option(opts.window, nm, vl)
+      pcall(vim.api.nvim_win_set_option, opts.window, nm, vl)
     end, options.window[opts.window])
   end
 end


shadmansaleh:
It's an upstream bug https://github.com/neovim/neovim/issues/19464

thedaviddelta:
Should we then just wait for upstream fix?

shadmansaleh:
> Should we then just wait for upstream fix?

Yeah.  You can try fixing it upstream. Wrapping the set call with pcall isn't the correct solution to this. It'll ignore all kinds of errors while setting the options. It'll just make it harder to find bugs if errors occur at that stage. I won't merge it as a fix.

thedaviddelta:
Yes I suppose it's not a good fix code-wise because it just inhibits the error.
But I'll just keep the "fix" in my own branch until it gets fixed upstream so I can avoid the issue.
Thanks!

PR #915, opened 2022-12-20
Passing self as second arg to self.options.fmt

daliusd:
I found this undocumented option and it is really useful if I want
to show if buffer is modified in tabline. However I need `tabnr`
for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,11 @@ sections = {
                    --   padding = { left = left_padding, right = right_padding }
 
       fmt = nil,   -- Format function, formats the component's output.
+                   -- This function receives two arguments:
+                   -- - string that is going to be displayed and
+                   --   that can be changed, enhanced and etc.
+                   -- - context object with information you might
+                   --   need. E.g. tabnr if used with tabs.
       on_click = nil, -- takes a function that is called when component is clicked with mouse.
                    -- the function receives several arguments
                    -- - number of clicks incase of multiple clicks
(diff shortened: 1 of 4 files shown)


shadmansaleh:
why not pass the entire self table? 

here too https://github.com/nvim-lualine/lualine.nvim/blob/bfa0d99ba6f98d077dd91779841f1c88b7b5c165/lua/lualine/component.lua#L275

daliusd:
> why not pass the entire self table?

Maybe. I don't know. As long it is not documented I assume anything might work. Personally, both would work for me.

E.g. if `self` is passed then I would need to change this code to show modified files in tabs only minimally:

```lua
          fmt = function(name, tabnr)
            local buflist = vim.fn.tabpagebuflist(tabnr)
            local winnr = vim.fn.tabpagewinnr(tabnr)
            local bufnr = buflist[winnr]
            local mod = vim.fn.getbufvar(bufnr, '&mod')

            return name .. (mod == 1 and ' +' or '')
          end
```

> https://github.com/nvim-lualine/lualine.nvim/blob/bfa0d99ba6f98d077dd91779841f1c88b7b5c165/lua/lualine/component.lua#L275

I assume here can be changed as well.

Was this change request?

daliusd:
OK. I have changed it everywhere and found one more place.

shadmansaleh:
one more thing add a note about component itself being passed as second argument in the docs

https://github.com/nvim-lualine/lualine.nvim/blob/bfa0d99ba6f98d077dd91779841f1c88b7b5c165/README.md?plain=1#L454

daliusd:
> one more thing add a note about component itself being passed as second argument in the docs
> 
> https://github.com/nvim-lualine/lualine.nvim/blob/bfa0d99ba6f98d077dd91779841f1c88b7b5c165/README.md?plain=1#L454

OK. Done. I have added example in `tabs` section as well.

PR #923, opened 2022-12-29
Add missing coma to README.md - filename component

QingTian1927:
Please correct me if I'm wrong but I think there should be a coma here


shadmansaleh:
There should be a coma here thanks.

PR #933, opened 2023-01-08
added colours to qf extension to distinguish quickfix and location list

gennaro-tedesco:
This PR introduces colours to distinguish between quickfix and location list (see screenshot below): I have been using it locally myself for a time and thought to propose it as a PR, should it be of interest for others too.

At the moment I have configured hardcoded highlight colours, however ideally the users could pass them as arguments: I have not been able to understand how (and if) it is possible; I would imagine something along the lines of
```lua

lualine.setup({
    extensions = { 
        quickfix.setup{ qf = '#...', ll = '#...'}, 
        ...
    }
}
```
perhaps you can direct me to how to achieve it (and I could improve the PR).

https://user-images.githubusercontent.com/15387611/211216863-5ba09dac-45e7-48e8-8b8b-8f03139f05b4.mov

### DIFF
--- a/lua/lualine/extensions/quickfix.lua
+++ b/lua/lualine/extensions/quickfix.lua
@@ -18,7 +18,7 @@ end
 
 local M = {}
 
-function M.init()
+function M.get_colours()
   -- Make sure ft wf doesn't create a custom statusline
   vim.g.qf_disable_statusline = true
   return {
@@ -32,7 +32,7 @@ M.sections = {
     {
       label,
       color = function()
-        return is_loclist() and { bg = M.init()['ll'] } or { bg = M.init()['qf'] }
+        return is_loclist() and { bg = M.get_colours()['ll'] } or { bg = M.get_colours()['qf'] }
       end,
     },
   },
--- a/lua/lualine/extensions/quickfix.lua
+++ b/lua/lualine/extensions/quickfix.lua
@@ -21,10 +21,21 @@ local M = {}
 function M.init()
   -- Make sure ft wf doesn't create a custom statusline
   vim.g.qf_disable_statusline = true
+  return {
+    ll = vim.api.nvim_get_hl_by_name('Constant', false).foreground,
+    qf = vim.api.nvim_get_hl_by_name('Identifier', false).foreground,
+  }
 end
 
 M.sections = {
-  lualine_a = { label },
+  lualine_a = {
+    {
+      label,
+      color = function()
+        return is_loclist() and { bg = M.init()['ll'] } or { bg = M.init()['qf'] }
+      end,
+    },
+  },
   lualine_b = { title },
   lualine_z = { 'location' },
 }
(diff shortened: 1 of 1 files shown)


lua/lualine/extensions/quickfix.lua
```diff
@@ -21,10 +21,21 @@ local M = {}
 function M.init()
```
shadmansaleh, line 21: init is a special fuction used to initialize the extension. use a different function to implement this

gennaro-tedesco: You are right, I have renamed the method to `M.get_colours()`.

lua/lualine/extensions/quickfix.lua
```diff
@@ -18,13 +18,24 @@ end
 
 local M = {}
 
-function M.init()
+function M.get_colours()
   -- Make sure ft wf doesn't create a custom statusline
   vim.g.qf_disable_statusline = true
+  return {
```
shadmansaleh, line 24: the qf_disable_statusline still only needs to be called on init. I asked you to move the part you added to different location.

gennaro-tedesco: Oops sorry, my bad, I did not understand. I have now moved the colours assignment to its own location, decoupling it from `M.init()`.

petobens:
@gennaro-tedesco I have 2 questions:
i) how do i change the colours for the qf extension? In particular I want to have a grey "inactive" color
ii) (see the gif below) add/remove a custom separator? i.e eithet add/remove the `>` separator after "Location List"

Peek 2023-04-01 09-11

gennaro-tedesco:
> i) how do i change the colours for the qf extension? In particular I want to have a grey "inactive" color 

as per my initial comment in the PR, the colours are hardcoded at the moment: I have not understood well enough the mechanism to pass them as parameters (perhaps the lualine author can provide some recommendations here?)

> ii) add/remove a custom separator? i.e eithet add/remove the `>` separator after "Location List"

same as above, it seems that customising colours elements of extensions changes the component separator: I do not really know why :(

petobens:
@shadmansaleh if you have any pointers they be greatly appreciated :)

shadmansaleh:
I think changing the color based if it's location list or quickfix in lualine was a bad idea. it doesn't feel like a out of the box behavior lualine should have. it's simple enough to have change color if anyone wants to in user-config . picking colors that'd work with all themes is pretty complicated. on top of that hardcoding it makes no sense. I'll revert this commit.

it was rather short sight be me to accept it in current form.

anyone still wanting the current behavior can do something like the following. even set their own colors in `qf_colors`

```lua
local quickfix_custom = require('lualine.extensions.quickfix');
quickfix_custom.sections.lualine_a[1].color = function()
  local is_loclist = vim.fn.getloclist(0, { filewinid = 1 }).filewinid ~= 0
  local qf_colours = {
    ll = vim.api.nvim_get_hl_by_name('Constant', false).foreground,
    qf = vim.api.nvim_get_hl_by_name('Identifier', false).foreground,
  }
  return  { bg = is_loclist and qf_colours['ll'] or qf_colours['qf']}
end

require('lualine').setup {
     extensions = { quickfix_custom, 'fugitive', 'nvim-dap-ui'}
}
```

PR #934, opened 2023-01-08
feat: Add support for Night Owl theme

oxfist:
Adding support for Sarah Drasner's [Night Owl Theme](https://github.com/sdras/night-owl-vscode-theme)

image
image
image
image


shadmansaleh:
We're no longing accepting colorscheme based themes in this repo after we externalized them to colorschemes. You can place the exact same theme file at the exact same place to the colorscheme ropo and have lualine load it seamlessly. This way colorschemes can have control over the looks. Most of the colorschemes nowadays provides their own version of lualine theme in this fashion [tokyonight](https://github.com/folke/tokyonight.nvim/tree/main/lua/lualine/themes) for example.

Also you can use any theme file locally by placing them in your configs usually ~/.config/nvim/lua/lualine/themes/. or even passing the theme table directly theme option while calling lualine setup would work too.

PR #945, opened 2023-01-16
feat: Add name+parent path option for component(filename)

mikaraunio:
Adds a `path` option to display the filename with its parent directory, ie. `dir/filename`.

Useful when you don't need the whole path, but your project has many files with the same name, and you want to easily see if you're working on `base/models.py`, `auth/models.py` or `products/models.py`.


lua/lualine/components/filename.lua
```diff
@@ -50,6 +50,19 @@ local function shorten_path(path, sep, max_len)
   return table.concat(segments, sep)
 end
 
+local function filename_and_parent(path)
+  local segments = vim.split(path, '/')
```
shadmansaleh, line 54: on windows path separator would be forword slash right?

mikaraunio: You're right, the separator shouldn't be hardcoded. I think using `path_separator` which is already being set for `shorten_path` could be the smart move here. Will look at it soon.

mikaraunio: Can't test on Windows, but now uses `path_separator` like `shorten_path`.

shadmansaleh:
looks good to me. thanks for your contribution.

PR #966, opened 2023-02-13
Add support for Moonbow theme

arturgoms:
Add support for [Moonbow](https://github.com/arturgoms/moonbow.nvim) theme.

normal
visual
replace
insert


shadmansaleh:
We're no longing accepting colorscheme based themes in this repo after we externalized them to colorschemes. You can place the exact same file at the exact same place to the colorscheme ropo and have lualine load it seamlessly. This way colorschemes can have control over the looks. Most of the colorschemes nowadays provides their own version of lualine theme in this fashion [tokyonight](https://github.com/folke/tokyonight.nvim/tree/main/lua/lualine/themes) for example.

PR #971, opened 2023-02-16
Fix dynamic coloring of tabline/winbar components

willothy:
I noticed that the color of the buffers component wasn't changing with the other components in my lualine setup as I changed modes:<br>
lua-tabs-before

I attempted to use a function to change the color based on mode, but encountered this visual issue where only part of the background of the active buffer changed:<br>
lualine-before-with-fn

I fixed that issue in highlight.lua, and added a `get_mode_suffix` function to simplify retrieving the mode suffix from outside of the highlight module.

I also added a `use_mode_colors` option to the buffers, tabs, and windows components (false by default), which automatically sets the colors of those components. 

How it looks now:<br>
lualine-after

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,6 +501,9 @@ sections = {
         alpha = 'Alpha'
       }, -- Shows specific buffer name for that filetype ( { `filetype` = `buffer_name`, ... } )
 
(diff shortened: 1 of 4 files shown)


shadmansaleh:
lgtm. thanks for your contribution

PR #989, opened 2023-03-09
Render terminal mode as insert

mireq:
I think, that terminal mode should be rendered as insert if insert mode is activated. This patch sets insert mode  properties to terminal mode in powerline theme. I can update other themes it this behavior is ok. 

term_insert
term_normal

### DIFF
--- a/lua/lualine/themes/powerline.lua
+++ b/lua/lualine/themes/powerline.lua
@@ -19,7 +19,7 @@ local Colors = {
   gray10         = '#f0f0f0',
 }
 
-return {
+local M = {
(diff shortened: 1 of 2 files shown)


oxfist:
Hey @mireq! Tests are failing. Docs say running `make test` should allow you to reproduce locally.

mireq:
@oxfist tests failing for another reason, it's not this commit. Next pull request has exactly same error (https://github.com/nvim-lualine/lualine.nvim/actions/runs/4378537914/jobs/7663406347)

PR #990, opened 2023-03-09
Add overseer extension to README

sho-87:
Overseer is a task runner plugin: https://github.com/stevearc/overseer.nvim

I was looking to see if lualine had an overseer extension so was digging around in the extensions dir and found it there, but it wasn't listed in the README, so just a short 1 line commit to update the text

### DIFF
--- a/README.md
+++ b/README.md
@@ -849,6 +849,7 @@ extensions = {'quickfix'}
 - nerdtree
 - nvim-dap-ui
 - nvim-tree
+- overseer
 - quickfix
 - symbols-outline
 - toggleterm


oxfist:
Not familiar with `overseer`, I'm guessing a short explanation could help with approval

sho-87:
Overseer is a task runner plugin: https://github.com/stevearc/overseer.nvim

I was looking to see if lualine had an overseer extension so was digging around in the extensions dir and found it there, but it wasn't listed in the README, so just a short 1 line commit to update the text

PR #991, opened 2023-03-10
feat: add dap server status to nvim-dap-ui extension

wassimk:
The current status of a DAP debug server can be retrieved via `require('dap').status()`. It's helpful always to see it on screen while debugging. So here it is added to lualine.

Here is what it'll look like _(bottom right)_:

### Non-PR Specific Question
The [`nvim-dap-ui`](https://github.com/rcarriga/nvim-dap-ui) plugin filetypes are all the windows around the main code window in the screenshot. I have a question about this. You can only see the dap status if you are focused on one of those filetypes. Is it possible to update a lualine status at any time? The [`nvim-dap`](https://github.com/mfussenegger/nvim-dap) plugin has an event system you can hook into. So, if it's possible to update `lualine_x` we could call the update method each time a DAP server event fires. How that fits into this extension code, I'm not sure.

### DIFF
--- a/lua/lualine/extensions/nvim-dap-ui.lua
+++ b/lua/lualine/extensions/nvim-dap-ui.lua
@@ -1,9 +1,27 @@
 -- MIT license, see LICENSE for more details.
 -- Extension for nvim-dap-ui
+
+local dap_lua, _ = pcall(require, 'dap')
+
+local function dap_status()
+  if not dap_lua then
+    return ''
+  end
+
+  local status = require('dap').status()
+
+  if status == nil or status == '' then
+    return ''
+  else
+    return 'DAP: ' .. status
+  end
+end
+
 local M = {}
 
 M.sections = {
   lualine_a = { { 'filename', file_status = false } },
+  lualine_x = { dap_status },
 }
 
 M.filetypes = {


shadmansaleh:
I think going to those windows to see the status is kind of inconvenient. The dap.status is meant to be directly used with statusline. You can use it directly in lualine too.

```lua
require'lualine'.setup {
  lualine_x = {require'dap'.status}
}
```

wassimk:
Thank you @shadmansaleh for the reply! I agree it's not as convenient!

I wrote it like this thinking the dap status always being in the statusline would be a performance concern. Since I use DAP a fraction of the time. Your reply suggests this is not really a concern. 

I wonder, is it possible to inject something into `lualine_x` and then remove it? For example, with [`nvim-dap-ui`](https://github.com/rcarriga/nvim-dap-ui#usage) you can execute code during debug start and stop.

```lua
local dap, dapui = require("dap"), require("dapui")
dap.listeners.after.event_initialized["dapui_config"] = function()
  -- inject dap status into lualine
  dapui.open()
end
dap.listeners.before.event_terminated["dapui_config"] = function()
  -- remove dap status from lualine
  dapui.close()
end
dap.listeners.before.event_exited["dapui_config"] = function()
  -- remove dap status from lualine
  dapui.close()
end
```

wassimk:
Thank you for the time to reply, @shadmansaleh. This is helpful!

PR #998, opened 2023-03-22
Allow specifying theme as function

sscherfke:
Use case: I have a custom theme that dynamically adjusts to the OS' dark/light mode. If the theme can be a function, that function can return the proper color values each time it is called.

Alternative: Create a `lua/lualine/themes/mytheme.lua` next to `lua/lualine/myconfig/mytheme.lua` which only does `require("myconfig.mytheme").lualine_theme()` and configure `lualine.setup({options = {theme = "mytheme"}})`.  This is IMHO more convoluted and a lot less readable/understandable.

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -251,6 +251,9 @@ local function setup_theme()
     elseif type(theme_name) == 'table' then
       -- use the provided theme as-is
       return config.options.theme
+    elseif type(theme_name) == 'function' then
+      -- call function and use returned (dynamic) theme as-is
+      return config.options.theme()
     end
     if theme_name ~= 'auto' then
       notify_theme_error(theme_name)


shadmansaleh:
why to just call the function in  the config?

say you have mytheme function that specifies what the current theme should be in current state. With this you'll be passing it like this 
```lua
require'lualine'.setup {
  options = {
    theme = mytheme,
  }
}
```

but you can already just call the function while setting the theme right?
```lua
require'lualine'.setup {
  options = {
    theme = mytheme(),
  }
}
```
should have the same effect

sscherfke:
In this case, `mytheme()` returns different colors depending on whether the background is dark or light.  With this PR , `mytheme` would be re-evaluated with each switch between dark/light mode.  If you just do `theme = mytheme()`, this would not be the case and the status line colors would not be adjusted.

shadmansaleh:
Ok that's a valid use case.

PR #1001, opened 2023-03-27
Fix #796 and close #797

### LINKED ISSUE #796: Bug: My base16 theme is overridden even though I'm using the "auto" theme

awanwar:
If there are errors loading nvim-base16, the setup_base16() call will throw an error instead of returning defaults. This has the effect of falling back to the 'auto' theme, allowing the colorscheme to remain more consistent with the existing theme as opposed to using a default that doesn't match the exising theme.

This solution also addresses the discussion in #797.

### DIFF
--- a/lua/lualine/themes/base16.lua
+++ b/lua/lualine/themes/base16.lua
@@ -36,45 +36,38 @@ local function setup(colors)
   return theme
 end
 
-local function setup_default()
-  return setup {
-    bg = '#282a2e',
-    alt_bg = '#373b41',
-    dark_fg = '#969896',
-    fg = '#b4b7b4',
-    light_fg = '#c5c8c6',
-    normal = '#81a2be',
-    insert = '#b5bd68',
-    visual = '#b294bb',
-    replace = '#de935f',
-  }
-end
-
 local function setup_base16()
   local loaded, base16 = pcall(require, 'base16-colorscheme')
 
   if not loaded then
     add_notice(
-      'nvim-base16 is not currently present in your runtimepath, make sure it is properly installed,'
-        .. ' fallback to default colors.'
+      'nvim-base16 is not currently present in your runtimepath. If this is unintended, make sure it is properly installed.'
     )
 
-    return nil
+    error('nvim-base16 is not currently present in your runtimepath, make sure it is properly installed.')
   end
 
   if not base16.colors and not vim.env.BASE16_THEME then
     add_notice(
       'nvim-base16 is not loaded yet, you should update your configuration to load it before lualine'
-        .. ' so that the colors from your colorscheme can be used, fallback to "tomorrow-night" theme.'
+        .. ' so that the colors from your colorscheme can be used.'
+    )
+
+    error(
+      'nvim-base16 is not loaded yet, you should update your configuration to load it before lualine'
+        .. ' so that the colors from your colorscheme can be used.'
     )
   elseif not base16.colors and not base16.colorschemes[vim.env.BASE16_THEME] then
     add_notice(
-      'The colorscheme "%s" defined by the environment variable "BASE16_THEME" is not handled by'
-        .. ' nvim-base16, fallback to "tomorrow-night" theme.'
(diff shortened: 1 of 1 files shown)


awanwar:
Doesn't seem like this commit fails the tests. Every PR since about 3 weeks ago seem to fail the these two tests with the same issue.

lua/lualine/themes/base16.lua
```diff
@@ -36,45 +36,38 @@ local function setup(colors)
...
+      'The colorscheme "%s" defined by the environment variable "BASE16_THEME" is not handled by' .. ' nvim-base16.'
+    )
+
+    error(
+      'The colorscheme "%s" defined by the environment variable "BASE16_THEME" is not handled by' .. ' nvim-base16.'
     )
   end
 
-  local colors = base16.colors or base16.colorschemes[vim.env.BASE16_THEME or 'tomorrow-night']
+  local colors = base16.colors or base16.colorschemes[vim.env.BASE16_THEME]
```
shadmansaleh, line 70: I don't get what you're doing here. If the base16 plugin is not present this color fetch will fail. that's why above checks are here. The bug and the pr you linked was about adding support to another base16 plugin.

awanwar: If the supported base16 plugin is not installed, the auto theme should try to detect what colors are used. What the current code does is `base16.lua`provides a default theme which the check in line 16 in `lua/lualine/themes/auto.lua` thinks is okay, and so the theme returns that default instead of trying to detect existing colors. Instead, if the supported plugin is not there but a base16 theme is installed, the auto theme should try to detect the colors in the theme instead of using a default theme that may not match the colors of the existing base16 theme.

For this fix in particular, if the supported plugin is not there or is configured incorrectly, the attempt to load the theme in `base16.lua` should fail. This in turn fails the check in line 16 in `auto.lua`, and allows the auto-detect process to proceed.

As a side note, if the user decides to specify the theme as `base16` (as opposed to `auto`), but `base16.lua` cannot load the theme for any particular reason, `lualine` should fall back to auto-detecting colors, just as it would do for any other non-base16 theme that fails to load.

PR #1011, opened 2023-04-09
feat: refresh on BufModifiedSet

camoz:
Add the event BufModifiedSet to default_refresh_events so that it is instantly visible if the buffer content is modified or not.


camoz:
(I find this useful when jumping around in history with `u`/`Ctrl-r` so that it is immediately obvious when you are at the current state of the file, i.e. the buffer modified indicator updates immediately.)

camoz:
I have squashed the commits now.

Is there anything I can/should do to get this merged?

PR #1013, opened 2023-04-09
feat: add option `always_show_tabs` to control whether show `tabline` when there are only one tab.

milanglacier:


wwkayak:
Works good... thanks

milanglacier:
> Works good... thanks

Thanks for your feedback... Hope we can merge this patch ASAP...

deponian:
In the meantime you can use this snippet until this PR is merged:
```lua
require('lualine').hide({place = {'tabline'}})
local lualine_tmp = vim.api.nvim_create_augroup('lualine_tmp', { clear = true })
vim.api.nvim_create_autocmd({ 'TabNew', 'TabClosed' }, {
  group = lualine_tmp,
  callback = function()
    if vim.fn.tabpagenr('$') == 1 then
      require('lualine').hide({place = {'tabline'}})
    else
      require('lualine').hide({place = {'tabline'}, unhide = true})
    end
  end,
})
```
Place this right after `require('lualine').setup{ ... }`

milanglacier:
> In the meantime you can use this snippet until this PR is merged:
> 
> ```lua
> require('lualine').hide({place = {'tabline'}})
> local lualine_tmp = vim.api.nvim_create_augroup('lualine_tmp', { clear = true })
> vim.api.nvim_create_autocmd({ 'TabNew', 'TabClosed' }, {
>   group = lualine_tmp,
>   callback = function()
>     if vim.fn.tabpagenr('$') == 1 then
>       require('lualine').hide({place = {'tabline'}})
>     else
>       require('lualine').hide({place = {'tabline'}, unhide = true})
>     end
>   end,
> })
> ```
> 
> Place this right after `require('lualine').setup{ ... }`

Actually, this snippet doesn't work, because when `lualine.hide` is called, the `showtabline` will still be `2`, so the effect of `lualine.hide` is simply that lualine will not try to decorate the tabline and the tabline will fallback to the default appearance.

Here is my workaround way to make `showtabline` always be 1:

```lua
        tabline = {
            lualine_a = {
                { 'filetype', icon_only = true },
            },
            lualine_b = {
                { 'tabs', mode = 2, max_length = vim.o.columns },
                {
                    function()
                        vim.o.showtabline = 1
                        return ''
                        --HACK: lualine will set &showtabline to 2 if you have configured
                        --lualine for displaying tabline. We want to restore the default
                        --behavior here.
                    end,
                },
            },
        },

```

It might be expensive to constantly call `vim.o.showtabline = 1` but I absolutely don't feel any latency caused by this call at all... So I am satisfied with this workaround.

deponian:
> Actually, this snippet doesn't work, because when `lualine.hide` is called, the `showtabline` will still be `2`, so the effect of `lualine.hide` is simply that lualine will not try to decorate the tabline and the tabline will fallback to the default appearance.

You are not quite right. `hide()` restores previous setting for `tabline` and `showtabline` options. See here: https://github.com/nvim-lualine/lualine.nvim/blob/05d78e9fd0cdfb4545974a5aa14b1be95a86e9c9/lua/lualine.lua#L488-L489

So if you have `showtabline=1` (which is default) then my snippet is correct.

milanglacier:
> what’s the status of this?

I hope it can be merged ASAP, or at least get some feedback.

Since all the tests (except for those which are resulted in CI bugs) have passed.

So currently I have to use my own workaround to bypass it.

milanglacier:
Any updates on this?

README.md
```diff
@@ -127,6 +127,7 @@ require('lualine').setup {
     },
     ignore_focus = {},
     always_divide_middle = true,
+    always_show_tabs = true,
```
shadmansaleh, line 130: I think always_show_tabline will be a better name. because tabline not necessarily shows just the tabs.

shadmansaleh:
lgtm. Thanks. Also sorry for the late reply.

PR #1052, opened 2023-05-25
diagnostic: do not show if disabled

sstallion:
I've been experimenting with a new workflow by keeping diagnostics disabled globally unless a project-specific configuration enables them. I found that lualine does not check `vim.diagnostic.is_disabled()` before showing diagnostic information. This PR adds a simple check to verify that diagnostics are indeed enabled before updating the statusline.

Thanks for a great plugin!


sstallion:
Final round of changes are done. I made one more modification to check if diagnostics are disabled for the current buffer rather than the global disable.

sstallion:
Looks like this has been fixed already. Thanks!

PR #1058, opened 2023-06-09
docs(plugin): add docs for installing lualine.nvim with lazy.nvim

sadiksaifi:


README.md
```diff
@@ -68,6 +68,17 @@ Last Updated On: 18-04-2022
 
 ## Installation
 
+### [lazy.nvim](https://github.com/folke/lazy.nvim)
+
+```lua
+return {
+  'nvim-lualine/lualine.nvim',
+  event = { "BufRead", "BufNewFile", "VimEnter" },
```
coffebar, line 76: I don't understand something here.
Why not just omit event line? The event will happen in any case, won't it?

if not:

Can BufRead be fired without VimEnter? 
Maybe VimEnter is enough here?

README.md
```diff
@@ -68,6 +68,17 @@ Last Updated On: 18-04-2022
...
+  dependencies = { 'nvim-tree/nvim-web-devicons', event = "VeryLazy" }
+  opts = {} -- lazy equivalent to setup({}) function
```
coffebar, line 78: I don't think this comment is appropriate here.
`-- lazy equivalent to setup({}) function`
This is an instruction for a specific plugin manager, not a "lazy vs packer" comparison.

If you look at the current instructions for the packer and do it in the same way, then it should be only 2 lines:
```lua
  {
    "nvim-lualine/lualine.nvim",
    dependencies = { "nvim-tree/nvim-web-devicons" },
  },
```

coffebar (changes requested):

Issue #888, opened 2022-11-13
Bug: FugitiveHead messes up menu tab completion


rkulla:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### Expected behaviour

:Packer[tab]  cycles through whole menu

### Actual behaviour

:Packer[tab] doesn't cycle through whole menu. (:Packer is just an example but really any :Command[tab] has same issue)

### Minimal config to reproduce the issue

lualine_a = {'FugitiveHead'},

### Aditional information

This works fine: lualine_a = {'branch'},

It's only when I use FugitiveHead that this happens. I tried most combinations of vim-fugitive being in start/ or opt/ and/or lualine being in start/ or /opt.

rkulla:
It also happens with or without adding `extensions = {'fugitive'}`

shadmansaleh:
I can't reproduce it. The completion cycling is working correctly for me with `FugitiveHead`. To be honest, here we are doing nothing but calling the `FugitiveHead()` viml function that vim-fugitive provides. It seems unlikely that this is a lualine bug.

https://github.com/nvim-lualine/lualine.nvim/blob/3325d5d43a7a2bc9baeef2b7e58e1d915278beaf/lua/lualine/components/special/eval_func_component.lua#L19-L21

rkulla:
Interesting.  It was the only way I could reproduce it but it must be something else conflicting with it then. Thanks.

ViViDboarder:
I did some debugging and was able to narrow this down to a combination of using `FugitiveHead` along with the change introduced in this commit for `vim-fugitive`: https://github.com/tpope/vim-fugitive/commit/4b0f2b604

I found that commit by running git bisect on the Fugitive tree between known working good and bad versions.

ViViDboarder:
I walked through changes within that diff and it turns out when `expand()` is called on line 401, it produces the bug. The expand call itself works and doesn't appear to error, however if I replace `expand('~')` with `/home/vividboarder`, the bug goes away. If I wrap the known directory with `expand` again (`expand('/home/vividboarder')`) it returns. So I'm fairly confident it's to do with the `expand(...)` function being called in this context.

To avoid the issue for now, you can pin vim-fugitive just before this commit or at the previous tag, v3.6.

ViViDboarder:
[...]

PR #1061, opened 2023-06-10
Skip running refresh when in command mode

ViViDboarder:
This works around the issue documented in #888 and stemming from [nevoim](https://github.com/neovim/neovim/issues/10121).


ViViDboarder:
Looks like the bug is going to be fixed in the an upcoming Neovim release. Might still be worth it to reduce necessary cycles.

PR #1063, opened 2023-06-17
add buffers filter

peter-lyr:
I think it will be useful for me to handle my requirement.

I want to show those buffers in the same git root directory. But all the buflisted buffer will be shown in tabline by default. So I dive into the lualine.nvim repo, and I made some changes to make it work.

Maybe it will be slow down, and maybe there is another way to solve this problem.

Thx!

https://github.com/nvim-lualine/lualine.nvim/assets/55801796/4441b5c5-88a7-40e9-bffd-7825d9c23c8d

### DIFF
--- a/lua/lualine/components/buffers/init.lua
+++ b/lua/lualine/components/buffers/init.lua
@@ -28,6 +28,7 @@ local default_options = {
     alternate_file = '#',
     directory = '',
   },
+  buffer_filter = nil,
 }
 
 -- This function is duplicated in tabs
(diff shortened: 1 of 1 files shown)


phi-mah:
This looks pretty similar to the PR I submitted half a year ago: #887
Hopefully one of the two / a combination thereof  will make it through

peter-lyr:
Is it possible to merge into the master branch? Otherwise I have to use your repo or my repo instead of `nvim-lualine/lualin.nvim`.

PR #1071, opened 2023-07-02
feat: wifi module to show connection status, essid and bit rate

Pheon-Dev:
Picture


shadmansaleh:
This looks pretty cool. But I'm not convinced this is something that needs to be shown in a text editor's statusline. Also this'd be pretty platform specific. 

I think it'd make more sense if you turned it into a lualine plugin. checkout here https://github.com/nvim-lualine/lualine.nvim/wiki/Plugins

Issue #886, opened 2022-11-06
Bug: flicker when running keymap


Rahlir:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem
Map a key to a command. Good example can be seen when using `delimitmate` plugin where there are imaps for  `<space>` and `<bs>`. You can also add your own maps, such as `map m :echo "hello"<cr>`, but for the very simple keymaps the flicker doesn't happen every time, only sometimes.

### Expected behaviour
When you run that keymap, no flicker of lualine should occur.

### Actual behaviour
Lualine flickers, mode seems to switch for a split second to command and this redraws the lualine.

### Aditional information
This bug is most apparent when other plugins are involved (such as delimitmate). However, this is clearly an issue with lualine, not other plugins. Of course those keymaps do cause vim to switch to command mode and back, but I think lualine should be able to filter out these super short switches caused by keymaps.

Rahlir:
Initially I thought this was a different issue, but on second look it seems that this is identical to the issue #852. Nevertheless, I won't close this just yet since I think the additional context I described might be helpful: I am pretty sure the flickering described in #852 is actually not happening simply because new line is added but because the user have a map for <cr> that is run when adding a new line.

folke:
You should use `:h <cmd>` for your mappings instead. `<cmd>` exists for this exact reason.

segeljakt:
It happens when running `<Plug>` keymaps as well, e.g.:
```lua
vim.keymap.set('', 'w', '<Plug>CamelCaseMotion_w', opt_plug)
vim.keymap.set('', 'b', '<Plug>CamelCaseMotion_b', opt_plug)
vim.keymap.set('', 'e', '<Plug>CamelCaseMotion_e', opt_plug)
```
This cannot be solved by `<Cmd>`

dabbertorres:
I noticed this with andymass/vim-matchup. I disabled the default mappings for that plugin and made the mappings myself using `<cmd>` (instead of `:<c-u>` like vim-matchup does) as suggested by @folke. The flickering between Insert and Command mode stopped, but the statusline still flickers. So better, but not gone.

[...]

PR #1076, opened 2023-08-08
fix: remove CursorMoved refresh events (#886)

HawkinsT:
This reverts https://github.com/nvim-lualine/lualine.nvim/commit/664c688974a7f183f6d6edae70c6aa117dcbd009 and https://github.com/nvim-lualine/lualine.nvim/commit/b87e47e1dee3cf96fc787f24e282ed5b52ed2fd2 which were originally introduced to make lualine refresh more frequently, but also introduced a bug that causes lualine to flicker under certain circumstances, as noted by https://github.com/nvim-lualine/lualine.nvim/issues/886 .

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -27,7 +27,7 @@ local refresh_real_curwin
 
 -- The events on which lualine redraws itself
 local default_refresh_events =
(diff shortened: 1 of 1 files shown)


HawkinsT:
Finally fixed by 0978a6c.

PR #1093, opened 2023-09-08
Set quickfix colours dynamically

f1rstlady:
The quickfix extension sets the colour of section A statically on startup. To handle highlighting changes, it has to be set dynamically.


shadmansaleh:
closed with #1118

PR #1094, opened 2023-09-12
Check if diagnostics are disabled.

sidequestboy:
Introduce a check for each diagnostic namespace whether that namespace has been disabled via `vim.diagnostic.disable` before counting it.

Also tidy the three neovim diagnostic counters with a common function.

Applies to "nvim_lsp", "nvim_workspace_diagnostic", "nvim_diagnostic".

I began looking at the test suite but this would be the first test of the diagnostic component - should we do a whole LSP setup for these tests?


sidequestboy:
There's actually some bugs here still - my bad I'll close and reopen

Issue #1096, opened 2023-09-13
Bug: lualine causes .git/index.lock file to persist, blocking subsequent git operations


svermeulen:
### Self Checks

- [x ] I'm using the latest lualine.
- [ x] I didn't find the issue in existing issues or PRs.

### Expected behaviour
lualine should not block unrelated git operations

### Actual behaviour
sometimes, when opening certain files inside a git repo, an index.lock file appears at .git/index.lock, and stays there

### Minimal config to reproduce the issue
N/A

I really tried to repro with a minimal .vimrc but I could not.  However - I am still pretty confident it's related to something lualine is doing, since removing lualine completely prevents the issue from happening.

Feel free to close the issue though.  Even as a closed issue it might be valuable to anyone else having this problem, since it took me awhile to narrow this one down.

### Additional information

My guess is that there is a git operation that lualine runs that crashes and then leaves the index.lock file there.

Should note also that I'm not using fugitive.vim, not sure if relevant

The file I noticed it with most has unix style endings, and this occurred on windows, though that is probably not relevant

scanny:
Solution might be the same as mentioned in this `vim-fugitive` issue, a number of other plugins have added the `--no-optional-locks` option to `git status` that is mentioned there. https://github.com/tpope/vim-fugitive/issues/1624#issuecomment-731840747

scanny:
Confirmed that only removing the `"diff"` item from lualine causes this problem to go away and come back. I have the `fugitive` extension installed, but it works fine without the diff item.

svermeulen:
Thanks!  Your mention of `diff` item led me through the code, where I discovered that git_diff.lua was sometimes force-killing the git diff operation

shadmansaleh:
I'd like to get it fixed. Meanwhile, it's likely everyone already has other more git-focused plugins more thoroughly tracking diff status than lualine. You can let diff component use those as source of info instead of lualines implementation. I'm using gitsigns as source for diff. Here's an example how to do it https://github.com/nvim-lualine/lualine.nvim/wiki/Component-snippets#using-external-source-for-diff

PR #1100, opened 2023-09-21
Bug fix for issue #1096, where git index.lock file is created but not deleted

svermeulen:
This was happening because in some cases the git diff operation was being force-killed, preventing it from clearing the index.lock file it created, and blocking all further git operations.

In my case I was getting this all the time because I had another plugin that triggered some code in BufEnter, which I think caused some buffer writes, triggering BufWritePost, causing lualine to execute multiple consecutive calls to update_git_diff

### DIFF
--- a/lua/lualine/components/diff/git_diff.lua
+++ b/lua/lualine/components/diff/git_diff.lua
@@ -13,6 +13,7 @@ local git_diff = nil
 local diff_output_cache = {}
 -- variable to store git_diff job
 local diff_job = nil
+local is_running = false
 
 local active_bufnr = '0'
 local diff_cache = {} -- Stores last known value of diff of a buffer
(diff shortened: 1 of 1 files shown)


lua/lualine/components/diff/git_diff.lua
```diff
@@ -123,13 +125,17 @@ end
 
 ---update git_diff variable
 function M.update_git_diff()
-  if M.diff_args then
+  -- It's important here that we check is_running, since
+  -- calling stop() on a git operation can have index.lock remain
+  -- and block other git operations
+  if M.diff_args and not is_running then
```
shadmansaleh, line 131: won't this lead to bugs like diff component not updating properly?

svermeulen: Possibly, yeah, so maybe it should use a flag to auto run it again in on_exit, when it is still running in update_git_diff

PR #1101, opened 2023-09-23
Use guibg in the quickfix extension

coot:


shadmansaleh:
guibg doesn't really do anything here. We take userinput as bg and it's always applied as GUI colors. cterms colors are inferred by lualine. check out the docs on color option

PR #1102, opened 2023-09-26
feat(use_mode_colors) add the option to toggle mode colors off for everything

Sam-programs:
simply adds `use_mode_colors` to everything not just the `tab` component

i found another bug while making this which is if  `use_mode_colors` is set to true for the `tab` component and the theme doesn't have a highlight for a vim mode (eg. terminal) it fallsback to having no color ?
which is why i had to specify each and every vim mode in the commits

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ require('lualine').setup {
   options = {
     icons_enabled = true,
     theme = 'auto',
+    use_mode_colors = true,
     component_separators = { left = '', right = ''},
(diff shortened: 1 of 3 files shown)


shadmansaleh:
What's the purpose of this actually?

Sam-programs:
well i didn't like the fact that colors change when i switch modes and i thought maybe others would like to have 1 color for all modes too

PR #1103, opened 2023-09-27
feat: Add LualineBuffersNext and LualineBuffersPrev to switch buffers

luv2code:
This PR adds two new commands for switching the current buffer to the one on the right (next) or the left (prev). If the selection will cycle.

I ran `make precommit_check` and `make format` and everything is good.

Here is an example of use that binds the commands to get browser-tab-like buffer switching:
```lua
vim.keymap.set({"n", "v"}, "<Leader><Tab>", ":LualineBuffersNext<cr>")
vim.keymap.set({"n", "v"}, "<Leader><S-Tab>", ":LualineBuffersPrev<cr>")
```

Thanks for considering my pull request.


luv2code:
I'm a dummy. `:bnext` and `:bprevious` work fine for this. no extra lualine code needed.

PR #1114, opened 2023-10-12
Fall back to GIT_DIR if git dir is not found

pedropombeiro:
In repositories that use a git dir which is not accessible by parent path traversal, it is nice to be able to read `GIT_DIR` and use it to show the current branch.

A project where this is important is https://yadm.io/, which is a dotfiles manager that stores its git dir in `~/.local/share/yadm/repo.git`.

I've tested this change locally and it works as expected:

image

### DIFF
--- a/lua/lualine/components/branch/git_branch.lua
+++ b/lua/lualine/components/branch/git_branch.lua
@@ -56,14 +56,27 @@ local function update_branch()
   branch_cache[vim.api.nvim_get_current_buf()] = current_git_branch
 end
 
+---updates the current value of current_git_branch and sets up file watch on HEAD file if value changed
+local function update_current_git_dir(git_dir)
+  if current_git_dir ~= git_dir then
+    current_git_dir = git_dir
+    update_branch()
+  end
+end
+
 ---returns full path to git directory for dir_path or current directory
 ---@param dir_path string|nil
----@return string
+---@return string|nil
 function M.find_git_dir(dir_path)
+  local git_dir = vim.env.GIT_DIR
+  if git_dir then
+    update_current_git_dir(git_dir)
+    return git_dir
+  end
+
   -- get file dir so we can search from that dir
   local file_dir = dir_path or vim.fn.expand('%:p:h')
   local root_dir = file_dir
-  local git_dir
   -- Search upward for .git file or folder
   while root_dir do
     if git_dir_cache[root_dir] then
@@ -101,9 +114,8 @@ function M.find_git_dir(dir_path)
   end
 
   git_dir_cache[file_dir] = git_dir
-  if dir_path == nil and current_git_dir ~= git_dir then
-    current_git_dir = git_dir
-    update_branch()
+  if dir_path == nil then
+    update_current_git_dir(git_dir)
   end
   return git_dir
 end


shadmansaleh:
image

I think if we'll support GIT_DIR it should come up before we walk directories not after

pedropombeiro:
I've tried a commit to check for it first, but it seems my attempt was too simplistic. The code in the `while` loop is trying to add a `.git` extension to the path, but this should not happen on `GIT_DIR` values.

pedropombeiro:
Should be good now

lua/lualine/components/branch/git_branch.lua
```diff
@@ -62,42 +62,48 @@ end
...
-          git_dir = git_dir and git_dir:match('gitdir: (.+)$')
-          file:close()
-        end
-        -- submodule / relative file path
-        if git_dir and git_dir:sub(1, 1) ~= sep and not git_dir:match('^%a:.*$') then
-          git_dir = git_path:match('(.*).git') .. git_dir
-        end
+  if vim.env.GIT_DIR then
+    git_dir = vim.env.GIT_DIR
+    file_dir = git_dir
```
shadmansaleh, line 68: why are you setting file_dir to git_dir ?? this is just used to cache git directory for the folder current file is in so we don't make the search every time git component runs.

also instead of putting main search logic in an else block. 

check if GIT_DIR is set. use that value to check if we need to update state. then early return from the function once the proceeding is done.

something like this
```lua
 if vim.env.GIT_DIR then
    git_dir = vim.env.GIT_DIR
    -- you can probably refactor this logic into a separate function so code repetition isn't needed
    git_dir_cache[file_dir] = git_dir
    if dir_path == nil and current_git_dir ~= git_dir then
      current_git_dir = git_dir
      update_branch()
    end
    return git_dir
  end
    
```

pedropombeiro: I've made the change as you suggested, and also removed the `git_dir_cache` assignment, since it wouldn't be used anyway when `GIT_DIR` is set. Let me know what you think.

lua/lualine/components/branch/git_branch.lua
```diff
@@ -56,14 +56,27 @@ local function update_branch()
...
+local function update_current_git_dir(git_dir)
+  if current_git_dir ~= git_dir then
+    current_git_dir = git_dir
+    update_branch()
+  end
+end
+
 ---returns full path to git directory for dir_path or current directory
 ---@param dir_path string|nil
----@return string
```
pedropombeiro, line 61: Line 120 can return a nil value, so the linter was complaining about this annotation.

shadmansaleh:
looks good to me. thanks.

smartinellimarco:
@pedropombeiro Should the yadm repo branch be picked up automatically by lualine? I'm not able to reproduce

pedropombeiro:
> @pedropombeiro Should the yadm repo branch be picked up automatically by lualine? I'm not able to reproduce

@smartinellimarco I have this code in my lua config to set `GIT_DIR` automatically: https://github.com/pedropombeiro/dotfiles/blob/c64afd8acead686d4235c61088a2803b9e4359fb/.config/nvim/lua/core/yadm-helper.lua#L10

PR #1115, opened 2023-10-16
added a semlink to examples in lua accessible directory

yechielw:


shadmansaleh:
it's intentionally like this. examples are given to inspire user configuration not as a pre config

PR #1118, opened 2023-10-18
Revert "added colours to qf extension to distinguish quickfix and location list"

shadmansaleh:
Reverts nvim-lualine/lualine.nvim#933


shadmansaleh:
context: https://github.com/nvim-lualine/lualine.nvim/pull/933#issuecomment-1767659028

PR #1119, opened 2023-10-18
Added extension for telescope

leisurelicht:
image


nezia1:
Hey, I'm confused on where the telescope integration went. I can't find it anywhere (it disappeared from the README), and trying to use it, it's saying it cannot find it.

PR #1125, opened 2023-10-24
feat: allow enabling extensions per buftype

martinparadiso:
Context and motivation
======================

I was writing an extension for the special terminal buffer 
and realized in has no ``filetype`` like the quickfix window,
instead filetype is empty and the `buftype` is 'terminal'. 
Currently here is no direct way to set an special bufferline 
for the terminal buffer.

Proposed change
===============

Add a ``buftypes`` table to the extension table, allowing to
filter extensions by buftype; and the corresponding
implementation.

Some basic test cases were added.

Example extension
=================

Example of an extension for terminal buffers:

<details>

<summary>Source code</summary>

```lua
local terminal_extension = {
    sections = {

        lualine_a = { function()
            return 'TERMINAL'
        end },

        lualine_b = {
            function()
                local name = vim.api.nvim_buf_get_name(0)
                local match = string.match(vim.split(name, ' ')[1], 'term:.*:(%a+)')
                return match ~= nil and match or vim.fn.fnamemodify(vim.env.SHELL, ':t')
            end },
        lualine_c = {
            function()
                return vim.b.term_title
            end
        },
        lualine_x = {},
        lualine_y = {

            function()
                local dev, _ = require('nvim-web-devicons').get_icon('zsh')
                return dev
            end
        },
        lualine_z = {},
    },
    buftypes = { 'terminal' },
}
```

</details>

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -887,7 +887,7 @@ You can find a bigger list [here](https://github.com/rockerBOO/awesome-neovim#ta
 ### Extensions
 
 lualine extensions change statusline appearance for a window/buffer with
-specified filetypes.
+specified filetypes or buffertypes.
 
 By default no extensions are loaded to improve performance.
 You can load extensions with:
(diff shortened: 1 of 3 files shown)


shadmansaleh:
Can you include the terminal extension too? It seems nice to have plus will be an example implementation.

martinparadiso:
Done, should be working with and without web-devicons.

Issue #921, opened 2022-12-27
Bug: lualine should not override user options unless explicitly requested


ZyX-II:
### Self Checks

- [x] I'm using the latest lualine. (fffbcb829288c3ca366f17cdb8d46aefb5c1d6f3)
- [x] I didn't find the issue in exsisting issues or prs.
       Specifically, #395 does mention the issue regarding `showtabline`, but it contains more generic feature request while this issue is more limited in scope; and existence of #670 suggests that current behaviour of ignoring `laststatus` should be classified as a bug.

### How to reproduce the problem

1. Run neovim with `-u exampleConfig.vim` (replace paths in first line with appropriate ones).
2. Type `:set showtabline? laststatus?`.

### Expected behaviour

Neovim reports both options as being equal to 1.

### Actual behaviour

They are both set to 2.

### Minimal config to reproduce the issue

```
set runtimepath=C:/tools/neovim/nvim-win64/share/nvim/runtime,P:/ath/to/lualine
set laststatus=1
set showtabline=1
lua << EOF
require('lualine').setup({
    tabline = { lualine_a = { 'tabs' } },
})
EOF
```

### Additional information

Workaround using OptionSet from #395 used as-is not good idea because it prevents setting this option to anything else. It is obviously possible to create workaround for this issue of the workaround, but having options to configure options is something strange. Also just setting options after setup call for some reason works in the minimal config above, but not in my actual config.

shadmansaleh:
I disagree with this. Most people using vim/nvim aren't aware of these options or what they do.

And people are usually looking for statuslline & tabline to be always visible. For example, if you're using your tabline to show your gitstatus it doesn't make sense for it only to be visible when there is more than 1 tab. That makes sense when you're using tabline to display only tabs like the builtin tabline . Same goes for statusline I don't think anyone wants their statusline to disappear when they have only one window open.

We are already taking over statusline & tabline options. I don't think there's anything wrong with us taking over a few more related options to provide better default behavior.

As for #395 I do intend to fix it properly. My idea is to make tabline disappear when the components in tabline all return empty. This lets the components decide when they want to show themselves. And the users can use `cond` option to provide their own logic when they want tabline to be shown, instead of relying on just the number of tabs. winbar already works like this. I just didn't get time to do the same for tabline yet.

ZyX-I:
> I disagree with this. Most people using vim/nvim aren't aware of these options or what they do.
>
[...]

Issue #904, opened 2022-11-29
Feat: Show tabline only when more than one tab


chrisgrieser:
### Requested feature
similar to vim's default behavior, the tabline should only be shown when there is more than one tab. (or there should be an option for that)

### Motivation
no reason to "waste" a line when only having one tab.

---

I tried using `cond` like this, but it seems that "tabs" creates a tabline, even if it is empty.
```lua
tabline = {
		lualine_a = {{
			"tabs",
			mode = 1,
			section_separators = "",
			cond = function () return fn.tabpagenr("$") > 1 end,
		}},
	},
```

aerosol:
https://github.com/nvim-lualine/lualine.nvim/issues/395#issuecomment-1312371694

theJian:
I opened a proposal to add a new option for this, @chrisgrieser , let me know what you think.

PR #1137, opened 2023-11-08
Add option tabline_always_visible

theJian:
#921 
#904 

This pr added an option `tabline_always_visible`, when setting it to `false`, tabline will be hidden if there are less than two tab pages.


milanglacier:
This PR serves the same purpose  as #1013

PR #1144, opened 2023-11-14
fixes #1105 : check for LSP progress, report working/linted file

### LINKED ISSUE #1105: Feat: diagnostics: show OK logo if no error (+ WIP logo)
### Requested feature
The diagnostics widget does not print anything
  - while the diagnostics are being produced
  - when the lint is done and successful

It could print a "work-in-progress" logo while waiting for the diagnostics, and an "OK" logo when completed successfully.

### Motivation
This would allow users to know in a glimpse
  - wether a linter is in progress
  - wether a linter is done

Indeed, as of now the UI is in the same state, wether a lint was successful or no lint was done whatsoever.

lafleurdeboum:
This small PR adds a handler to `vim.lsp.handlers`, checking if a language server reported that it was working on the current file and is not done yet. If it's the case, a "Working" icon is appended to the `diagnostics` component. Then when the language server reports that it's done working, if no error, warning, hint or info is declared, the "Working" icon is replaced by an "OK" icon.

This behaviour is useful when working on heavyweight linters, like `rust-analyzer`, that can take a minute to lint a project. The user gets the bare useful information with very little overhead. Unfortunately, several language servers don't report their progress (thinking about [vim-language-server](https://github.com/iamcco/vim-language-server) and [bash-language-server](https://github.com/bash-lsp/bash-language-server). In this case, I found no way to make sure that a linting is currently running. Indeed, the `vim.diagnostic` API lets _any_ process diagnose what they please. My intention is to ask `vim-language-server` and `bash-language-server` to declare a minimal work progress, since this behaviour seems quite standard.

I tested my PR against `luacheck` and it succeeded. I also tried to respect the project's customs (like indentation). Any comment greatly appreciated !

### DIFF
--- a/lua/lualine/components/diagnostics/init.lua
+++ b/lua/lualine/components/diagnostics/init.lua
@@ -56,10 +56,10 @@ function M:init(options)
   self.lsp_working = false
   vim.lsp.handlers["$/progress"] = function(_, result)
     self.lsp_started = true
-    if result.value.kind == "end" then
-      self.lsp_working = false
-    else
+    if result.value.kind == "report" then
       self.lsp_working = true
+    elseif result.value.kind == "end" then
+      self.lsp_working = false
     end
   end
 end


lua/lualine/components/diagnostics/init.lua
```diff
@@ -48,12 +50,25 @@ function M:init(options)
...
     )
   end
+
+  -- Check if a language server reports working progress.
+  self.lsp_started = false
+  self.lsp_working = false
+  vim.lsp.handlers["$/progress"] = function(_, result)
+    self.lsp_started = true
+    if result.value.kind == "end" then
+      self.lsp_working = false
```
Slotos, line 60: `gopls` will emit a "begin" event when encountering an issue loading workspace, with no followup "end". In order to recover from bad actors, LSP progress handlers need to implement a debounced clear routine.

lafleurdeboum:
Hi @Slotos, thanks for caring ! I don't use to use Go, could you show me how to reproduce the bug, or perhaps point me to a small repo that would reproduce it, so that I can implement this debounce ? I often use luals, vimls, bashls and rust-analyzer, and I never faced that issue (I suppose the issue would imply having both the "working" and either of "ok" or "warning" icons).

Slotos:
In my case it was enough to initiate an edit on a new go buffer in a directory with no `go.mod` in its hierarchy with gopls set up. So something like:

```
:=require('lspconfig').gopls.setup{}
:new test.go
```

I was implementing a primitive progress reporter, and saw gopls message getting stuck in a routine very similar to the one I commented on. When logging progress messages, I saw “begin”, “end”, and “begin” messages arrive in that order.

lafleurdeboum:
Thanks ! I was able to reproduce this issue, and solved it by reacting to "report" signals from `$/progress` instead of "begin". It does solve the issue with `gopls` sending a lone " begin" signal before failing. It's indeed cleaner like that.

lafleurdeboum:
Side note : we won't get the "ok" icon if no language server reported working on the current buffer. As a matter of fact, not collecting any error message doesn't imply that the file was linted, either by a language server or by any other linting process.

Is anyone aware of any means to know that the buffer _was_ indeed linted ?

PR #1147, opened 2023-11-17
fix(#820): Illegal character in filename

joaomsa:
Also ran into this issue with the `jdt://` URLs, as reported in #820. Instead of using JDT-specific workarounds like #1030 and #821, let's fix the root cause.

Given a directory structure like `%dir1/%dir2/%actual_%file'`, the shortening process should eliminate the `%` escaping. It currently shortens it to `%/%/%%actual_%%file'` instead of the correct '%%/%%/%%actual_%%file'. The unescaped `%/` is an illegal char sequence.

To address this, it would make sense to run the escaping after shortening has been completed.

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -86,8 +86,6 @@ M.update_status = function(self)
     data = vim.fn.expand('%:t')
   end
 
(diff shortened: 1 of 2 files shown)


shadmansaleh:
lgtm, thanks for your contribution.

Issue #781, opened 2022-08-04
Bug: Results of wildmode `list:longest` erased almost immediately after pressing Tab


antoineco:
### Summary

When the `wildmode` option is set to `list:longest`, a Tab keystroke completes _till the longest common string_.
Results are normally shown on top (instead) of the statusline, and are persisted on screen until the user keeps typing or presses Tab another time, or Esc.

However, with Lualine enabled, the results are erased almost instantly after being shown on screen, and are replaced with the statusline again.

This does not occur with vim-lightline, for example.

https://user-images.githubusercontent.com/3299086/182959757-e72a5595-ba8d-4b17-9246-ca45d0bc2330.mov

---

### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in exsisting issues or prs.

### How to reproduce the problem

(With the default, empty Lualine config)

```vim
:set wildmode=list:longest  " alternatively: list:longest,list:full
```

```vim
:set wi<Tab>
```

### Expected behaviour

Results are shown on top (instead) of the statusline, and persisted until the user performs one of the actions listed in the issue description.

### Actual behaviour

Results are erased very quickly, and replaced with the statusline again.

### Minimal config to reproduce the issue

```lua
-- init.lua

vim.opt.wildmode = 'list:longest'

require'lualine'.setup()
```

### Aditional information

Does not occur with vim-lightline.

PR #1150, opened 2023-11-20
fix: flickering don't react to mode changes if a mapping is not done

Sam-programs:
This pr makes lualine ignore mode changes when a mapping is not done.
Which removes flicking for mappings that switch to normal mode then back to insert mode command mode then back etc.

### DIFF
--- a/lua/lualine/utils/mode.lua
+++ b/lua/lualine/utils/mode.lua
@@ -4,51 +4,68 @@ local Mode = {}
 
 -- stylua: ignore
 Mode.map = {
-  ['n']      = 'NORMAL',
-  ['no']     = 'O-PENDING',
-  ['nov']    = 'O-PENDING',
-  ['noV']    = 'O-PENDING',
-  ['no\22'] = 'O-PENDING',
-  ['niI']    = 'NORMAL',
-  ['niR']    = 'NORMAL',
-  ['niV']    = 'NORMAL',
-  ['nt']     = 'NORMAL',
-  ['ntT']    = 'NORMAL',
-  ['v']      = 'VISUAL',
-  ['vs']     = 'VISUAL',
-  ['V']      = 'V-LINE',
-  ['Vs']     = 'V-LINE',
-  ['\22']   = 'V-BLOCK',
-  ['\22s']  = 'V-BLOCK',
-  ['s']      = 'SELECT',
-  ['S']      = 'S-LINE',
-  ['\19']   = 'S-BLOCK',
-  ['i']      = 'INSERT',
-  ['ic']     = 'INSERT',
-  ['ix']     = 'INSERT',
-  ['R']      = 'REPLACE',
-  ['Rc']     = 'REPLACE',
-  ['Rx']     = 'REPLACE',
-  ['Rv']     = 'V-REPLACE',
-  ['Rvc']    = 'V-REPLACE',
-  ['Rvx']    = 'V-REPLACE',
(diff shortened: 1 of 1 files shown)


Sam-programs:
I am not sure if the `redrawstatus` is gonna cause issues, It seems to be working fine with v9.4 and nightly neovim.
Tested with:
```lua
local keymap = vim.keymap.set

keymap('v','g',':call <C-z>')
keymap('n','g',':call <C-z>')
keymap('v','t',':call <C-z><cr>')
keymap('n','t',':call <C-z><cr>')

-- These don't update the statusline right away but are there any real use cases of them?
keymap('n','w',':ls<C-v>')
keymap('n','m',':ls<C-r>')
```

shadmansaleh:
If I remember correctly this occers when ':' is used instead of `<cmd>` right? Using `<cmd>` is recommended way to map ex commands as they don't change neovim state to jump to cmdline. the flickering you're seeing is I think due to mode change event triggering lualine refresh. I'm not sure if a component triggering lualine refresh is a good idea . In general it is advised not to use redrawstatus.

Also in the example keybinds you've given I don't understand what it's doing. can you explain?

One more thing can this be fixed by making lualine refresh on `CmdlineChanged` in [here](https://github.com/nvim-lualine/lualine.nvim/blob/3abbbdd5e9aa02f06ffc297d73f73ba21f4e5084/lua/lualine.lua#L30) ?

Sam-programs:
> If I remember correctly this occers when ':' is used instead of <cmd> right? Using <cmd> is recommended way to map ex commands as they don't change neovim state to jump to cmdline. the flickering you're seeing is I think due to mode change event triggering lualine refresh. 

Yes but using `:` allows commands to be repeated by `@:` and many plugins use it.
Flickering also happens when switch between insert and normal mode while a mapping is running not just the command line.
```lua
-- autopair example not easy to see but does flicker
vim.keymap.set('i','t','{<cr>}<esc>==O')
```
Sometimes even when using `:normal`, e.g. 
```lua
vim.keymap.set('i','g','<cmd>normal ==<cr>')
```
Both of these of course don't need the cmdline `redrawstatus`.
>I'm not sure if a component triggering lualine refresh is a good idea 

> One more thing can this be fixed by making lualine refresh on CmdlineChanged in [here](https://github.com/nvim-lualine/lualine.nvim/blob/3abbbdd5e9aa02f06ffc297d73f73ba21f4e5084/lua/lualine.lua#L30) ?

I am not sure where else to put the refresh since a `redrawstatus` is needed on top of the refresh.
> . In general it is advised not to use redrawstatus.
Also in the example keybinds you've given I don't understand what it's doing. can you explain?

I found the `redrawstatus` issue [here](https://github.com/nvim-lualine/lualine.nvim/issues/781) since the menu was disappearing, I used the keybinds to open the completion menu to test if it disappears after the mappings are done. 
Also the `redrawstatus` side effects have been reduced to nothing for most versions.
https://github.com/neovim/neovim/commit/ad1f353fe1aeb54144a34d1a0de8e318bd5113aa.
tags are:v0.9.4 v0.9.3 v0.9.2 v0.9.1 v0.9.0 v0.8.3 v0.8.2 v0.8.1 v0.8.0 stable nightly 

Maybe an nvim 0.7 compat branch can be made?, it would also help with [winbar](https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine.lua#L566).
Or a check for neovim versions in the code like winbar.

PR #1151, opened 2023-11-21
fix: flicking with noice detect actual_win with autocmds instead of using vim.g.actual_win

Sam-programs:
this pr locks window focus detection between CmdlineEnter and CmdlineLeave.
to prevent loss of focus by ui replacement plugins.
ui plugins usually return the focus back to the current window after half a second causing flickering.

https://github.com/nvim-lualine/lualine.nvim/assets/130783534/ad285103-1848-4b25-a2e9-bf578b8b7e42

### DIFF
--- a/lua/lualine/utils/utils.lua
+++ b/lua/lualine/utils/utils.lua
@@ -99,9 +99,32 @@ function M.define_autocmd(event, pattern, cmd, group)
   end
 end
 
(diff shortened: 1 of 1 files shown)


Sam-programs:
nvm this doesn't happen in nightly neovim, only in 0.9.4 as far as am aware.

Issue #1122, opened 2023-10-20
Bug: require('lualine').hide() doesn't hide winbar with Goyo


cprn:
### Self Checks

- [X] I'm using the latest lualine.
- [X] I didn't find the issue in existing issues or PRs.

### How to reproduce the problem

I apologize, it's not an isolated issue - it only happens when I use Goyo with Lualine but I'm asking for help because I can't put my finger on it. I tried to make it as easy to analyze as possible.

1. use provided `init.lua`
1. open any plain text file and call `:Goyo`.

### Expected behaviour
Both statusline and winbar gone.

### Actual behaviour
Only statusline gone.

### Minimal config to reproduce the issue

```lua
-- { 1. install some add-on manager (I'm using Lazy):
local lazypath = vim.fn.stdpath 'data' .. '/lazy/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system {
        'git',
        'clone',
        '--filter=blob:none',
        'https://github.com/folke/lazy.nvim.git',
        '--branch=stable',
        lazypath,
    }
end
vim.opt.rtp:prepend(lazypath)
-- }

-- { 2. install Goyo, Limeligth, Lualine and configure some winbar
require('lazy').setup({
    {'junegunn/limelight.vim'},
    {'junegunn/goyo.vim'},
    {'nvim-lualine/lualine.nvim', dependencies = {'nvim-tree/nvim-web-devicons', 'SmiteshP/nvim-navic'}, opts = {
        options = {theme = 'onedark', disabled_filetypes = {statusline = {'NvimTree'}}},
        winbar = {lualine_c = {{'filetype', icon_only = true}, 'navic'}}
    }},
}, {})
-- }

-- { 3. add Goyo config to autohide lualine
local goyo_group = vim.api.nvim_create_augroup('GoyoGroup', {clear = true})
vim.api.nvim_create_autocmd('User', {
    desc = 'Enable LimeLight on GoYo enter',
    group = goyo_group,
    pattern = 'GoyoEnter',
    command = 'Limelight|lua require("lualine").hide()',
})
vim.api.nvim_create_autocmd('User', {
    desc = 'Disable LimeLight on GoYo leave',
    group = goyo_group,
    pattern = 'GoyoLeave',
    command = 'Limelight!|lua require("lualine").hide {unhide=true}',
})
-- }
```

### Additional information
[...]

PR #1157, opened 2023-11-28
fix: set option to nil if not cached and set by lualine

Sam-programs:
https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine.lua#L567
This line doesn't restore the global 'winbar', because lualine never sets the global 'winbar'.
I am not sure why but Goyo's window uses the global winbar option,and lualine not setting global winbar causes the winbar to stay in Goyo.
fixes(https://github.com/nvim-lualine/lualine.nvim/issues/1122)

### DIFF
--- a/lua/lualine/utils/nvim_opts.lua
+++ b/lua/lualine/utils/nvim_opts.lua
@@ -90,6 +90,8 @@ function M.restore(name, opts)
         restore_to = ''
       end
       vim.api.nvim_set_option(name, restore_to)
+    else
+      vim.o[name] = nil
     end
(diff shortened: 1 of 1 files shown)


shadmansaleh:
I don't think if an option isn't set by us we should touch it when it's being restored. This should probably be fixed on Goyo so it properly hides and unhides lualine. We could probably throw a user even when lualine is being hidden or unhidden. But I haven't seen interest in something like that yet. So probably not necessary for now.

Thanks for looking into it.

PR #1164, opened 2023-12-11
feat: add proper distro icons based on fileformat

peter-teslenko:
Inspired by airline I wrote a similar implementation to display the correct OS logo
https://github.com/ryanoasis/vim-devicons/blob/master/plugin/webdevicons.vim#L565

### DIFF
--- a/lua/lualine/components/fileformat.lua
+++ b/lua/lualine/components/fileformat.lua
@@ -9,21 +9,81 @@ local symbols = {
   mac = '', -- e711
 }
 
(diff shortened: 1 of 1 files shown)


shadmansaleh:
File format has nothing to do with distro or os. It says how line endings are marked in file. whether LF or CR is used or both. If distro was to be a component it should be a separate one. Also, it seems unnecessary tbh. There are so many distros so many OSs there will always be some not present and don't really do much other than decoration.

PR #1172, opened 2023-12-27
fix: use larger ft icon when coloring applied

ribru17:
This PR adds a space after the filetype icon *before* the highlight is applied, meaning that the space after the icon will also get the same foreground highlight as the icon itself, allowing the icon to take its proper full width on most (all?) terminal emulators. Users of `vim-devicons` or users who set the icon highlight to `false` will have the same experience as before. Examples:
Before:
lualinebeforelua
After:
lualineafterlua
Before:
lualinebeforerust
After:
lualineafterrust

### DIFF
--- a/lua/lualine/components/filetype.lua
+++ b/lua/lualine/components/filetype.lua
@@ -40,6 +40,9 @@ function M:apply_icon()
       icon = ''
       icon_highlight_group = 'DevIconDefault'
     end
+    if icon then
+      icon = icon .. ' '
+    end
     if self.options.colored then
       local highlight_color = modules.utils.extract_highlight_colors(icon_highlight_group, 'fg')
       if highlight_color then
@@ -56,7 +59,7 @@ function M:apply_icon()
   else
     ok = vim.fn.exists('*WebDevIconsGetFileTypeSymbol')
     if ok ~= 0 then
-      icon = vim.fn.WebDevIconsGetFileTypeSymbol()
+      icon = vim.fn.WebDevIconsGetFileTypeSymbol() .. ' '
     end
   end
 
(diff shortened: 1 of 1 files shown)


zuqini:
Just wanted to follow up on this PR and ask if there's anything additional we can do to help get it merged in.

I've been looking forward to this fix but the PR seems to have lost some traction. Appreciate the hard work!

ribru17:
@shadmansaleh Do you need any extra work to be done on this?

shadmansaleh:
Looks good to me. 

However, it doesn't make the icons bigger for me. what terminal are you using? Is this specific to vim-devicons? I have nvim-devicons and am using kitty. 

Though, I don't see much downside to adding a space to icons if it makes it better for some people.

ribru17:
Interesting... I use kitty as well and I have to say the effect is much more noticeable on wezterm. Especially with the Rust icon (but this is a separate issue, nvim-devicons users a Rust icon that kitty renders very small for some reason). That being said, for me on kitty it looks like the icons are slightly larger with this change

shadmansaleh:
So kitty already renders the icon in full size and on wezterm it requires an extra space is that what you're saying?

ribru17:
Basically yes; or rather, they both require an extra space behind the icon (unless you set a special setting in kitty), but Wezterm isn't smart enough to recognize that the icon should be full width if the empty space after the icon has a different foreground color (even though it is just an empty space). This PR moves the space so it is applied before the foreground color changes, allowing Wezterm and similar terminals to recognize that the icon should be full-width

vurtune:
I think this space may be optional in config, because some users use mono font, like me, no extra space needed.

Extra space showed in red

Julpikar:
its look too many space if setting with just only icon or icon in right position. for wezterm small rendered icon issue, i have found this issue.

small rendered font: https://github.com/wez/wezterm/issues/342
probably fix: https://wezfurlong.org/wezterm/config/lua/config/allow_square_glyphs_to_overflow_width.html

PR #1192, opened 2024-02-11
Minor change updating deprecated apis in one of the example files,

seancraven:
I referenced the nvim docs just to update the api's used in the example. Please close if this is a waste of time.


seancraven:
Found there was another pr with the exact change.

PR #1194, opened 2024-02-12
feat: Allow cterm highlight number through

### LINKED ISSUE #1176: Feat: Setting ANSI/Raw colors
### Requested feature

I'd like to be able to set ANSI colors, without them being mapped to a hex/RGB value.
Most numbers work correctly but for some if I try `theme={normal={a={bg=2, fg=21}}}`, I observe `ctermfg=12` instead of 21:
```
lualine_a_normal xxx ctermfg=12 ctermbg=2 guifg=#0000ff guibg=#008000
```

### Motivation
Users of https://github.com/tinted-theming/base16-shell set their themes with ANSI color codes and then change the display value for the ANSI colors in their terminal, hence changing all at once (git, ipython, vim, tmux...).

If anything this is an anti-feature, I'd like lualine to do *less* with its input.

AntoinePrv:
Close #1176

### DIFF
--- a/lua/lualine/highlight.lua
+++ b/lua/lualine/highlight.lua
@@ -76,6 +76,19 @@ local function sanitize_color(color)
   end
 end
 
+---converts color_name type colors to cterm format and let cterm color pass through
+---@param color string|number
+---@return string
+local function sanitize_color_for_cterm(color)
+  if type(color) == 'number' then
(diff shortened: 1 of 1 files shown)


shadmansaleh:
lgtm. Thanks for your contribution.

PR #1195, opened 2024-02-14
Wombat theme: Fix B section background

muellerj:
I'm not sure if this is a typo, but having `#808080` as background for sections `B` and `Y` looks weird and not at all like in the screenshots.

Setting it to `#666666` looks better and more true to the screenshots.


muellerj:
Nevermind, I figured out that `#808080` gets overwritten by my `base16.sh`. Closing this.

PR #1198, opened 2024-02-18
fix: fzf extension selected element parsing

### LINKED ISSUE #1197: Bug: fzf extension shows 'nil' instead of selected element on split pickers
In the code for the fzf extension, there was [a section added to lualine](https://github.com/nvim-lualine/lualine.nvim/blob/7d131a8d3ba5016229e8a1d08bf8782acea98852/lua/lualine/extensions/fzf.lua#L40) to show the current selected element. This only works for "floating" fzf pickers and not "split" pickers.

### Self Checks

- [x] I didn't find the issue in existing issues or PRs.

1. Install fzf plugin
2. Enable fzf extension for lualine
3. Open a SPLIT fzf picker and notice the lualine on the bottom has section y as nil

### Expected behaviour

This section should not show 'nil' on fzf split pickers

### Actual behaviour

The section shows 'nil' because of [this line](https://github.com/nvim-lualine/lualine.nvim/blob/7d131a8d3ba5016229e8a1d08bf8782acea98852/lua/lualine/extensions/fzf.lua#L24), where "selected" is no longer in the table returned by `get_info()`.

trevarj:
On non-floating windows the selected element is not supported and showed 'nil'.
This fixes that and also uses fzf's path parser for a cleaner parsing of the
selection.

Closes #1197

### DIFF
--- a/lua/lualine/extensions/fzf.lua
+++ b/lua/lualine/extensions/fzf.lua
@@ -5,10 +5,12 @@ works with both https://github.com/junegunn/fzf.vim and https://github.com/ibhag
 -- fzf-lua must be set-up in split mode
 ]]
 
-local fzf_lua, _ = pcall(require, 'fzf-lua')
+local function has_fzf()
+  return pcall(require, 'fzf-lua')
+end
 
 local function fzf_picker()
-  if not fzf_lua then
+  if not has_fzf() then
     return ''
   end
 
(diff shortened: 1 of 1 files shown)


trevarj:
Relates to https://github.com/ibhagwan/fzf-lua/issues/537

PR #1202, opened 2024-03-07
feat: align output of searchcount component with neovim default

camoz:
Neovim's default formatting for indicating the search result position, from `:h search-commands`:

```
  [1/5]         Cursor is on first of 5 matches.
  [1/>99]       Cursor is on first of more than 99 matches.
  [>99/>99]     Cursor is after 99 match of more than 99 matches.
  [?/??]        Unknown how many matches exists, generating the
                statistics was aborted because of search timeout.
```

Meaning of result.incomplete, from `:h searchcount()`:

```
  0: search was fully completed
  1: recomputing was timed out
  2: max count exceeded
```

### DIFF
--- a/lua/lualine/components/searchcount.lua
+++ b/lua/lualine/components/searchcount.lua
@@ -24,8 +24,17 @@ function M:update_status()
     return ''
   end
 
-  local denominator = math.min(result.total, result.maxcount)
-  return string.format('[%d/%d]', result.current, denominator)
+  if result.incomplete == 0 then
+    return string.format('[%d/%d]', result.current, result.total)
+  elseif result.incomplete == 1 then
+    return '[?/??]'
+  elseif result.incomplete == 2 then
+    if result.current <= result.maxcount then
+      return string.format('[%d/>%d]', result.current, result.maxcount)
+    else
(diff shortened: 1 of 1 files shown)


camoz:
I have tested every case locally (result.incomplete == 0|1|2) it works here. Not sure about the performance impact though.

Also, idk why that blank line was included in the diff :confused:

lua/lualine/components/searchcount.lua
```diff
@@ -24,8 +24,17 @@ function M:update_status()
...
-  return string.format('[%d/%d]', result.current, denominator)
+  if result.incomplete == 0 then
+    return string.format('[%d/%d]', result.current, result.total)
+  elseif result.incomplete == 1 then
+    return '[?/??]'
+  elseif result.incomplete == 2 then
+    if result.current <= result.maxcount then
+      return string.format('[%d/>%d]', result.current, result.maxcount)
+    else
+      return string.format('[>%d/>%d]', result.maxcount, result.maxcount)
```
shadmansaleh, line 35: what do we have in result.current in this case?

camoz:
Been using this for nearly 2 months now without any issues...

PR #1204, opened 2024-03-11
fix(branch): extract correct branch name for term buffers

Willem-J-an:
terminal buffers have a different path that did not work to retrieve the branch name. By extracting the path out of the terminal path, the branch name can still resolve even when in term buffers.


shadmansaleh:
lgtm. thanks

PR #1206, opened 2024-03-12
Allow customizing the file path

patnr:
Usage example:
`some-long/path/to/the/OneDrive - Company/folder/file`
could get replaced by
`D/folder/file`
which is particularly useful if `~/D` symlinks to the OneDrive dir.

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -16,6 +16,7 @@ local default_options = {
   file_status = true,
   newfile_status = false,
   path = 0,
+  custom_shortener = nil,
   shorting_target = 40,
 }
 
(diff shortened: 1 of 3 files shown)


lua/lualine/components/filename.lua
```diff
@@ -90,6 +91,10 @@ M.update_status = function(self)
     data = self.options.symbols.unnamed
   end
 
+  if self.options.custom_shortener then
+    data = self.options.custom_shortener(data, path_separator)
+  end
+
   if self.options.shorting_target ~= 0 then
```
shadmansaleh, line 98: when custom shorter is used we should disable this.

patnr: Are you sure? It can be nice to have both features (without user re-implementing `shorten_path`). And it can always be disabled by setting `shorting_target`. And

shadmansaleh: Alright, can be disabled on the user side easily enough. then it's fine.

shadmansaleh:
Remove the changes from doc/lualine.txt. it'll be auto-generated upon merge

shadmansaleh:
One more thing can't this be already done with existing `fmt` option?

patnr:
Yes, you're right, it absolutely can. Sorry for wasting your time!

Issue #1213, opened 2024-03-19
Bug: User runtime path doesn't have priority on Windows


daephx:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in existing issues or PRs.

### Describe the issue

I adding a custom module to my user config `lua/lualine/themes/<theme>.lua` to modifying the colors used by lualine and my preferred theme. The changes worked without issue on my Linux host, but these settings where not applied to my Windows host after pulling over the changes.

It seems the logic added in this commit nvim-lualine/lualine.nvim@afb8bfb for prioritizing the users config directory when loading themes has some side effects with Neovim's default path structure on Windows.

https://github.com/nvim-lualine/lualine.nvim/blob/af4c3cf17206810880d2a93562e0a4c0d901c684/lua/lualine/utils/loader.lua#L229-L233

I believe the culprit is the use of `vim.startswith` method to compare the path strings.
On Linux, this makes perfect sense as the paths are quite different:

Config: `/home/<user>/.config/nvim`
Data: `/home/<user>/.local/share/nvim`

But on Windows the default locations are as follows:

Config: `C:\Users\<user>\AppData\Local\nvim`
Data: `C:\Users\<user>\AppData\Local\nvim-data`

If it's not apparent as to why `vim.startswith` would see these directories as equivalent.
The function will only match up to a portion of the data path:

`C:\Users\<user>\AppData\Local\nvim`
`C:\Users\<user>\AppData\Local\nvim`-data

Resulting in the match being true for basically any plugin, not allowing the user config directory to take priority!

I replaced this with a simple `string.match` that leveraging the [`sep`]erator variable provided by the `lualine_require` module to ensure that the path string is matched to the appropriate directory, and I have not exerienced any issues with this change.

```lua
local lualine_require = require('lualine_require')
local sep = lualine_require.sep
-- ...
local user_config_path = vim.fn.stdpath('config')
table.sort(files, function(a, b)
  local pattern = table.concat { user_config_path, sep }
  return string.match(a, pattern) or not string.match(b, pattern)
end)
```

### How to reproduce the problem

1. Be on Windows
2. Set your colorscheme to one that is compatible with lualine themes
3. Create a module in the user config directory with the same folder namespace
   Example: `/lua/lualine/themes/tokyonight.lua`
4. Take notice that none of your changes are applied when you reload your colorscheme.
Despite the fact that using the `require` function on the new module will return a table with your custom settings.

PR #1214, opened 2024-03-19
fix: prioritize user config for theme-loader on windows (#1213)

daephx:


daephx:
@shadmansaleh Man I even referenced the commit in my issue and didn't notice lualine_require just duplicates the same logic... Should I patch that one too? I've never used the custom require function so I idk in what circumstances it will have the same problem.

https://github.com/nvim-lualine/lualine.nvim/blob/b5e8bb642138f787a2c1c5aedc2a78cb2cebbd67/lua/lualine_require.lua#L65-L67

PR #1215, opened 2024-03-21
fix: prioritize user config for extensions on windows

daephx:
Fixes the second part of: #1213

The same logic fixed in (#1214) is duplicated in the lualine_require module, which appears to be used internally for loading things like extensions.

On windows, user defined extensions that overwrite the builtin similarly cannot be loaded due to the same string matching issue. In addition, the number of queried runtime files is always less-than or equal to 1 due to the `all` parameter for `nvim_get_runtime_file` being set to false. Making it impossible to sort the list of files and prioritize ones within the users config path `vim.fn.stdpath("config")`.

### DIFF
--- a/lua/lualine_require.lua
+++ b/lua/lualine_require.lua
@@ -54,16 +54,17 @@ function M.require(module)
   end
 
   pattern_path = table.concat { 'lua/', module:gsub('%.', '/'), '.lua' }
-  local paths = vim.api.nvim_get_runtime_file(pattern_path, false)
+  local paths = vim.api.nvim_get_runtime_file(pattern_path, true)
   if #paths <= 0 then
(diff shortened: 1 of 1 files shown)


daephx:
Hi @shadmansaleh,

I wanted to follow up on this PR which has been open for a few months. I understand you might be busy, but I’d appreciate it if you could take a look when you have the chance. Thanks!

Upon reflection, the change from `vim.startswith` to `string.match` was a bit superfluous, as the main issue is just that the path string needed a trailing separator to prevent bad matches with Windows `$ENV:LOCALAPPDATA\nvim-data`. But both results are basically the same.

shadmansaleh:
Thanks again. I didn't see this one.

PR #1228, opened 2024-04-11
feat: Add BOM indicator to encoding component

sergiuser1:
Inspired by vim-airline, it adds `[BOM]` to the encoding if the file uses byte-order mark

### DIFF
--- a/README.md
+++ b/README.md
@@ -670,6 +670,20 @@ sections = {
 }
 ```
 
+#### encoding component options
+
+```lua
+sections = {
+  lualine_a = {
+    {
+      'encoding',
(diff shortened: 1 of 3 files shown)


shadmansaleh:
Does vim/neovim's builtin statusline show this by default? if not can you put it behind an option, set default to false?

sergiuser1:
By default it doesn't have anything about file encoding, I found this about making a custom component: https://stackoverflow.com/a/10416234/11710862

I'll add the option and some tests and will update the PR.

sergiuser1:
@shadmansaleh ping

sergiuser1:
Thanks!

Issue #1077, opened 2023-08-09
Feat: Better integration with plugins like oil


12bchl:
### Requested feature
With oil set as default explorer, when you jump out to current dir, you will be in a remote oil buffer. This means you are outside your git repo and your filename will be of the form oil://home/user/gitFolder/repoFolder/subDir...

This is pretty gross. Unfortunately the oil team did not have any solution.

I really like the branch and filename section types. And I use the diff symbols regularly when editting oil buffers. I'd just like to have my repo section and filename section working sensibly with oil directory buffers.

The filetype is recognized as "oil", so I'm wondering if it is possible to override the branch string and filename string receivers for given filetypes? I think it would be a pretty extensible feature and useful for other things too, maybe helpful with ssh. 

### Motivation
Sane & expected lualine appearance with oil or other plugins that do not operate directly in the apparent cwd.

bryanboateng:
Where you able to find some sort of solution to this problem?

12bchl:
Nope

shadmansaleh:
You can use the `fmt` option to transform result of a component before it gets displayed.

So in this case you can set fmt option on branch and filename component to a function that gets rid of the things that don't need to be show from result of these component in case of oil filetype.

A bit of filetype check and pattern matching should be enough.

the fmt funciton will get the result of the component as it's argument and what ever the function returns will be rendered in lualine.

![image](https://github.com/nvim-lualine/lualine.nvim/assets/13149513/b37564a8-051e-4cca-99a0-29e8aea3765b)

https://github.com/nvim-lualine/lualine.nvim#general-component-options

DovieW:
Wasn't this [PR](https://github.com/nvim-lualine/lualine.nvim/pull/1154) supposed to add this?

My issue is that it says [ no name ]

DovieW:
@kjuq Would you be able to weigh in on why we can't get the path to show?
All plugins are up to date (using lazy.nvim) and I can see the oil.lua file you created in my filesystem.
In my case it always says `[No Name]`

kjuq:
@DovieW Could you send me a minimal configuration, which probably enables only lualine and oil, to reproduce the issue? In my environment, the path is shown properly.

DovieW:
Thanks for responding @kjuq.

I was able to get the old `oil:///` path instead of `[No Name]`. 
It seems that having ANY options defined in the lualine config, causes the `[No Name]` path. The entire `opts` block has to be commented it out.
[...]

PR #1242, opened 2024-05-20
`branch` supports `oil` now

Tronikelis:
Hey, so when using oil file manager the branch is empty when viewing an oil buffer, this is because oil has its own path, now this is a small change to make it work, but I don't really like modifying globally the component like this, maybe this project should have some hooks or apis that the user could override, like `get_buffer_dir` ?

The same problem is with `filename` component and the fix is the same to use the oil api get_current_dir, so thats why I think having a global user editable api for directories would be a great option

https://github.com/nvim-lualine/lualine.nvim/issues/1077

https://github.com/nvim-lualine/lualine.nvim/issues/1077#issuecomment-2119754614

### DIFF
--- a/lua/lualine/components/branch/git_branch.lua
+++ b/lua/lualine/components/branch/git_branch.lua
@@ -74,12 +74,15 @@ function M.find_git_dir(dir_path)
     return git_dir
   end
 
-  local oil_exists, oil = pcall(require, 'oil')
-
   -- get file dir so we can search from that dir
   local file_dir = dir_path or vim.fn.expand('%:p:h')
-  if vim.bo.filetype == 'oil' and oil_exists then
-    file_dir = vim.fn.fnamemodify(oil.get_current_dir(), ':p:h')
+
+  local oil_exists, oil = pcall(require, 'oil')
+  if oil_exists then
+    local ok, dir = pcall(oil.get_current_dir)
+    if ok and dir and dir ~= '' then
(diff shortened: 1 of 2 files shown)


dtvillafana:
Bumping this PR, I would also like these changes.

Tronikelis:
If anyone wants these changes already, I created a separate plugin that patches the lualine components and exports them https://github.com/Tronikelis/lualine-components.nvim

shadmansaleh:
If I understand correctly with this enabled branch component shows git branch based on which directory you have opened in oil right?

This effects all buffers not just the oil buffer? or is it just on the buffer where oil is opened?

Tronikelis:
@shadmansaleh 

> If I understand correctly with this enabled branch component shows git branch based on which directory you have opened in oil right?

yes

> This effects all buffere not just the oil buffer?

It affects all buffers, this is fine because `get_current_dir` returns empty string if current buffer is not oil so we just check for that

Tronikelis:
I first thought that I should check the buffer filetype as you can see in this commit 

[`2941f5b` (#1242)](https://github.com/nvim-lualine/lualine.nvim/pull/1242/commits/2941f5b2cc3660fcb9671bbfe4ff15532a1f74eb)

But I ran into some issues where I presume lualine updates before oil sets the buffer type or something like that and that code just did not work, but I am very new to nvim/lua in general so it could be a skil issue

shadmansaleh (approved):

PR #1243, opened 2024-05-20
fix: Location component was showing wrong column with inlay hints.

davidt:
The location component had been previously changed to use
`vim.api.virtcol` (the screen position of the cursor) instead of
`vim.api.col` (the byte position of the cursor), in order to prevent
showing the wrong column when multibyte characters are present.
Unfortunately, the new inlay hints in neovim 0.10 make heavy use of
virtual text, and therefore the column shown in the location component
is often incorrect.

This change fixes it to use `vim.api.charcol`, which correctly handles
variable-width characters without including virtual text.

### DIFF
--- a/lua/lualine/components/location.lua
+++ b/lua/lualine/components/location.lua
@@ -2,7 +2,7 @@
 -- MIT license, see LICENSE for more details.
 local function location()
   local line = vim.fn.line('.')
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Thanks for the fix

PR #1246, opened 2024-05-23
Deprecated function in example evil_lualine.lua

bryanme5:
When using the example of [evil_line.lua](https://github.com/nvim-lualine/lualine.nvim/blob/master/examples/evil_lualine.lua#L163), show the following error:

remove

In neovim 0.12 vim.lsp.get_active_clients() function will be [deprecated](https://neovim.io/doc/user/deprecated.html#vim.lsp.get_active_clients()), vim.lsp.get_clients() solve the problem and works perfectly

### DIFF
--- a/examples/evil_lualine.lua
+++ b/examples/evil_lualine.lua
@@ -160,7 +160,7 @@ ins_left {
   function()
     local msg = 'No Active Lsp'
     local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
-    local clients = vim.lsp.get_active_clients()
+    local clients = vim.lsp.get_clients()
     if next(clients) == nil then
       return msg
     end


iton0 (changes requested):
vim.api.nvim_buf_get_option is also deprecated with nvim 0.10.
I believe that `local buf_ft = vim.api.nvim_get_option_value('filetype', { buf = 0 })
` should be used instead

examples/evil_lualine.lua
```diff
@@ -160,7 +160,7 @@ ins_left {
   function()
     local msg = 'No Active Lsp'
     local buf_ft = vim.api.nvim_buf_get_option(0, 'filetype')
```
iton0, line 162: vim.api.nvim_buf_get_option is also deprecated with nvim 0.10.
I believe that `local buf_ft = vim.api.nvim_get_option_value('filetype', { buf = 0 })
` should be used instead

bryanme5: For better readability and be more specific defining the buffer, we can use `local buf_ft = vim.api.nvim_get_option_value('filetype', { 'buf': 0 })`, no?
The scope option it's not necessary because by default the use of buf implies {scope} is `local`.
Here is the [documentation](https://neovim.io/doc/user/api.html#nvim_get_option_value())

iton0: Yeah the only adjustment is instead of `'buf' : 0` it should be `buf = 0`

yaotutu (approved):

PR #1260, opened 2024-06-12
Create trouble_v3 extension

fixerer:
Trouble v3 has other filetype (`trouble`) and other way to get mode.

NB: Only works for single windows, since last_mode is overwritten.

### DIFF
--- a/lua/lualine/extensions/trouble_v3.lua
+++ b/lua/lualine/extensions/trouble_v3.lua
@@ -1,0 +1,23 @@
+local M = {}
+
+local function get_trouble_mode()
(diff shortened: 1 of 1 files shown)


shadmansaleh:
There's already a trubble extension you should modify that to add support to this. 

you can have

```lua
M.filetypes = { 'Trouble' , 'trouble'}
```

So that extension gets loaded on both filetypes. Then update `get_trouble_mode` to support both. You don't need to create a new extension with a separate name.

shadmansaleh:
resolved in #1383

PR #1270, opened 2024-07-05
feat: add more match for bufferline icons

Parsifa1:


shadmansaleh:
Are these icons specific to mini.icon? Or do they exist in nvim-web-devicons?

lua/lualine/components/buffers/buffer.lua
```diff
@@ -38,12 +38,94 @@ function Buffer:get_props()
...
+    elseif self.filetype == 'neotest-output-panel' then
+      dev, _ = require('nvim-web-devicons').get_icon_by_filetype('neotest-output-panel')
+    elseif self.filetype == 'neotest-summary' then
+      dev, _ = require('nvim-web-devicons').get_icon_by_filetype('neotest-summary')
+    elseif self.filetype == 'oil' then
+      dev, _ = require('nvim-web-devicons').get_icon_by_filetype('oil')
     elseif self.filetype == 'TelescopePrompt' then
-      dev, _ = require('nvim-web-devicons').get_icon('telescope')
+      dev, _ = require('nvim-web-devicons').get_icon_by_filetype('TelescopePrompt')
     elseif self.filetype == 'fugitive' then
```
Parsifa1, line 125: 事实上原本的这个查询语句并不能正常工作

Parsifa1:
nvim-web-devicons does not provide icons for these (maybe some do), but if nvim-web-devicons cannot be queried, it will return `nil`, which will not have any impact.

lua/lualine/components/buffers/buffer.lua
```diff
@@ -38,12 +38,94 @@ function Buffer:get_props()
     local status, _ = pcall(require, 'nvim-web-devicons')
     if not status then
       dev, _ = '', ''
+    elseif self.filetype == 'minideps-confirm' then
```
shadmansaleh, line 41: Can you put a comment at the start of these and at the end of these. stating the icons in this block exist in mini.icons (with link). and that it's relying on the fact that mini.icons patching nvim-web-devicons interface. Tbh this feels like an anti-pattern that they are patching other modules like this. I'm still not sure about relying on it like this. especially not without proper documentation.

Parsifa1: Maybe you are right, these adjustments can wait until the subsequent development of `nvim-web-devicons` or is `lualine` directly compatible with `mini.icons`?

PR #1285, opened 2024-07-31
Add new filename path option

ptdewey:
Thanks for the great plugin!

I was finding it mildly confusing to tell which file I'm in at any given time in a Rust project I've been working on recently due to the presence of many 'mod.rs' files. However, I prefer my status line to be quite minimal, ideally showing only the filename. In this case, I thought it would be helpful to dynamically add the parent directory to the filename component when multiple buffers are open with files of the same name.
As such, I decided to add a new path option to lualine to accomodate this behavior.

**Example:** 
Directory structure:
```bash
src
├── mod.rs
└── parser
    └── mod.rs
```

Relevant lualine config:
```lua
lualine_c = {
     { "filename", path = 5, padding = 1, },
 },
```
With one of the mod.rs files open, lualine will look like this:
image

When opening the other mod.rs file, lualine will now look like this: (for parser/mod.rs)  
image
And this for src/mod.rs:
image

Deleting either buffer will cause the component to revert to showing only the filename again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,6 +639,7 @@ sections = {
                                -- 2: Absolute path
                                -- 3: Absolute path, with tilde as the home directory
                                -- 4: Filename and parent dir, with tilde as the home directory
+                               -- 5: Filename or filename and parent dir when there are duplicate filenames open, with tilde as the home directory
 
       shorting_target = 40,    -- Shortens path to leave 40 spaces in the window
                                -- for other components. (terrible name, any suggestions?)
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -81,6 +81,26 @@ M.update_status = function(self)
   elseif self.options.path == 4 then
     -- filename and immediate parent
     data = filename_and_parent(vim.fn.expand('%:p:~'), path_separator)
+  elseif self.options.path == 5 then
+    -- check for duplicate filenames in current buffers and add parent dir if a duplicate is found
+    local current_filename = vim.fn.expand('%:t')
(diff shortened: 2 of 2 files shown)


Suryansh-Dey:
We really need this in opened buffers also since implementing it with a fmt callback seems too inefficient because it gets called too many times without a reason to update. While it must be calculated only when a new buffer is listed.  
NvChad tabufline and all modern editors do this to avoid ambiguity.  
Also the eclipse feature i.e. max_char after which the name gets eclipsed.  
I think these are very basic requirements.

Suryansh-Dey:
Efficient workaround till then
```lua
local name_cache = {}
vim.api.nvim_create_autocmd({ "BufAdd", "BufDelete", "BufFilePost" }, {
    callback = function() name_cache = {} end,
})

local function get_unique_name(filename, bufnr)
    if name_cache[bufnr] then return name_cache[bufnr] end

    local path = vim.api.nvim_buf_get_name(bufnr)

    local clash = false
    for _, b in ipairs(vim.api.nvim_list_bufs()) do
        if b ~= bufnr and vim.bo[b].buflisted then
            local other_path = vim.api.nvim_buf_get_name(b)
            if vim.fn.fnamemodify(other_path, ':t') == filename and path ~= other_path then
                clash = true
                break
            end
        end
    end

    local result = clash and (vim.fn.fnamemodify(path, ':p:h:t') .. '/' .. filename) or filename
    name_cache[bufnr] = result
    return result
end
....
opts = {
    sections = {
        lualine_a = {
            {
                'buffers',
                fmt = function(name, context)
                    return get_unique_name(name, context.bufnr)
                end
            }
        },
...
    }
}
```

PR #1288, opened 2024-08-04
extensions:fugitive: add repo name

tmnvanderberg:
Add the repository name to the fugitive extension.

image

In my use case, I  have a bunch of different project repositores that are forks of the same code base. I like to see what repo I'm currently in when I'm using fugitive to make sure.

Not sure how many people really need this, but IMO it's a sensible default, also given the current fugitive extension only shows the branch and there is plenty of space.


lua/lualine/extensions/fugitive.lua
```diff
@@ -7,8 +7,16 @@ local function fugitive_branch()
   return icon .. ' ' .. vim.fn.FugitiveHead()
 end
 
+local function get_git_toplevel_basename()
+  local handle = io.popen("basename `git rev-parse --show-toplevel`")
```
shadmansaleh, line 11: I don't think this works cross-platform.
See if fugitive provides repo name somewhere you can fetch it from there easily in that case.

PR #1296, opened 2024-08-24
Git branch component: Add option to limit the length of the displayed branch name

### LINKED ISSUE #1289: Feat: Configurable length for branch component
If `branch` component's string is too long, it takes away space from other components.

Take for example my config, where I've put the branch component into the tabline, next to the tabnames (displays current active buffer of the tab). I think that identifying which tab you are on is way more important than knowing (the entirety) of the branch name. Note, that there are 10 tabs open, out of which only 2 are visible, due to the branch name being displayed in full.
image
The problem also exists on the default layout: The name of the current buffer is: `this_is_a_long_file_name_for_real`, but it gets shortened in favor of the branch name.
image

### Requested feature

I would like to provide a config option for choosing the maximum length of the `branch` component's text, such as `tabs`'s `max_length` or `filename`'s `shorting_target`. I would gladly implement this feature.

### Motivation

Branch name in most cases conveys less information than the current name of the buffer/tab.
While it's good to have/see on the screen, it's not needed to see the entirety (especially if the branch name is a long string).

Nandax00:
Added a new function that truncates git branch name if the max_length option is set. It also attempts to strip the last incomplete word after the last word separator (i.e. "-", "_").

Demo in tabline with `max_length = 30`:
- First one's straight-forward.
  branch name = `thisisjustonereallylongbranchnamewithoutseparators` =>
  image
- Here, the 30-long substring would contain "inexplica" at the end, but it is stripped.
  branch name = `this-one-contains-an-inexplicably-long-word` =>
  image

I'm using the following tabline settings:
```lua
  tabline = {
    lualine_a = {
      {
        "tabs",
        -- options
        -- ...
      }
    },
    lualine_b = {
      {
        "branch",
        -- NEW OPTION
        max_length = 30,
      },
      "diff",
    },
  },
```

closes #1289.

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,6 +490,21 @@ These are options that are available on specific components.
 For example, you have option on `diagnostics` component to
 specify what your diagnostic sources will be.
 
+#### branch component options
+
+```lua
+sections = {
+  lualine_a = {
+    {
+      'branch',
+      max_length = 0, -- Maximum displayed characters of the branch name.
+                      -- Takes effect, if the value is larger than 0.
+                      -- Recommended setting: 30.
+    }
+  }
+}
+```
+
 #### buffers component options
 
 ```lua
@@ -608,6 +623,20 @@ sections = {
 }
 ```
 
+#### encoding component options
+
+```lua
+sections = {
+  lualine_a = {
+    {
+      'encoding',
+      -- Show '[BOM]' when the file has a byte-order mark
+        show_bomb = false,
+    }
+  }
+}
+```
+
 #### fileformat component options
 
 ```lua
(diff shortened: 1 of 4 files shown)


DavidePatria:
you can use fmt option and format a string as you wish
```
    lualine_b = { {
      'branch',
      -- 'branch', 'diff', 'nvim_diagnostics',
      -- qui viene passato direttamente il componente di interesse
      fmt = function(str) return format(str, 10) end,
      -- separator = { right = ''}
      separator = { left = '', right = '' },
    } }

function format(str, len)
  if string.len(str) > len then
    -- tagliare il nome del ramo troppo lungo in mezzo anziché mostrare solamente l'inizio
    return string.sub(str,1,len/2) .. '…' .. string.sub(str, str:len()-len/2+1, str:len())
  else
    return str
  end
end

```

Nandax00:
Thanks, I definitely missed that part in the docs :D Closing PR and ticket

PR #1298, opened 2024-08-31
feat: filename now supports oil.nvim

Tronikelis:
https://github.com/nvim-lualine/lualine.nvim/issues/1077

This PR adds filename support for oil.nvim, so it's no longer the full path but how the user specified the path to be shown

To be honest this is kind of a selfish PR where I am modifying the vanilla component, and as I said in my previous PR it would be better to somehow extract this logic for the user to write himself, but maybe that's too much effort

### DIFF
--- a/lua/lualine/components/filename.lua
+++ b/lua/lualine/components/filename.lua
@@ -69,21 +69,38 @@ end
 M.update_status = function(self)
   local path_separator = package.config:sub(1, 1)
   local data
+
+  local function set_data(arg)
+    data = vim.fn.expand('%' .. arg)
+
+    if package.loaded.oil and vim.bo.filetype == 'oil' then
+      data = vim.fn.fnamemodify(require('oil').get_current_dir(), arg)
+    end
+  end
+
+  local arg
+
   if self.options.path == 1 then
     -- relative path
-    data = vim.fn.expand('%:~:.')
+    arg = ':~:.'
   elseif self.options.path == 2 then
     -- absolute path
-    data = vim.fn.expand('%:p')
+    arg = ':p'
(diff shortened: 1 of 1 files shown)


cgimenes:
Why isn't this done inside the oil extension?

tronikelis:
> Why isn't this done inside the oil extension?

Is it possible to do this easily? When I implemented this I remember checking out extensions and seeing some blocker that I can't recall.

If possible yeah, this should be an extension, but the only way I see is somehow extending the current filename to add support

cgimenes:
I don't think we would really need the filename component. AFAIK something like this would work as a custom component inside de oil extension: 
```lua
function() {
      -- proper module exists validation
      local oil = require 'oil'
      return oil.get_current_dir() .. oil.get_cursor_entry().name
}
```

tronikelis:
> I don't think we would really need the filename component. AFAIK something like this would work as a custom component inside de oil extension:
> 
> ```lua
> function() {
>       -- proper module exists validation
>       local oil = require 'oil'
>       return oil.get_current_dir() .. oil.get_cursor_entry().name
> }
> ```

The filename component automatically shrinks the path when it does not fit, without it I wouldn't be able to read the path when in a terminal split most of the time, without that it's not even worth adding it imo.

Your solution would show the full path all the time

cgimenes:
I see. So, we could move this shrinking logic to another place for it to be reusable. What do you think?

tronikelis:
> I see. So, we could move this shrinking logic to another place for it to be reusable. What do you think?

That's one solution, but we are not really solving the root problem, the root problem is that getting the current directory should be exposed to the global config somehow, then everything would just work

PR #1300, opened 2024-09-16
fix(loader): look for themes on `package.path`

mrcjkb:
Fixes https://github.com/nvim-neorocks/rocks.nvim/issues/536.

When lualine.nvim is installed as a luarocks package[^1], the loader can't find the themes on the runtimepath, because
Neovim doesn't keep the runtimepath and `package.path` in sync.
This PR introduces a fallback search of the `package.path` if a theme can't be found on the runtimepath.

[^1]: luarocks installs packages with a non-flat hierarchy, with `lua` and `lib` directories in a separate tree.

### DIFF
--- a/lua/lualine/utils/loader.lua
+++ b/lua/lualine/utils/loader.lua
@@ -220,8 +220,12 @@ local function load_theme(theme_name)
   end
   local n_files = #files
   if n_files == 0 then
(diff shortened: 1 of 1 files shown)


shadmansaleh:
Thanks. neorocks looks interesting.

PR #1307, opened 2024-10-13
Adding Kanagawa-dragon theme

mathieurousseau:
Kanagawa-dragon adaptation


gloggers99:
Why is this not merged? This was opened last year and I don't see any problems with it.

PR #1316, opened 2024-11-01
Fixup: fix flickering in lualine

shadmansaleh:
debounce updates for 100ms while setting the options

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,9 +141,9 @@ require('lualine').setup {
     always_divide_middle = true,
     globalstatus = false,
     refresh = {
-      statusline = 1000,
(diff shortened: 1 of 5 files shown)


mikesmithgh:
Hi 👋 I noticed that the autocmd triggers were removed. In my personal config, I have all timers set to 0 and rely only on autocmds to avoid triggering frequent function calls. Is there a way to use autocmds after this PR or do I need to explicitly set it up myself? Thanks!

Keivan-sf:
@mikesmithgh Just wanted to say thanks for sharing this. This way of handling refreshes solved my flickering issues

PR #1326, opened 2024-11-10
feat(filename): add filename_map option

simonmandlik:
This PR adds a possibility to further apply any transform to the `filename`. I find handy especially in temporary buffers (from e.g. codecompanion, git diff, oil...)

### DIFF
--- a/README.md
+++ b/README.md
@@ -644,6 +644,7 @@ sections = {
                                -- 2: Absolute path
                                -- 3: Absolute path, with tilde as the home directory
                                -- 4: Filename and parent dir, with tilde as the home directory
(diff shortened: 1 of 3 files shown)


simonmandlik:
> _Warning_: This pull request is touching the following templated files:
> 
> * doc/lualine.txt - `lualine.txt` is generated from README.md. Make changes there instead.

Should be fixed now

shadmansaleh:
Why not just use already existing `fmt` option ? It allows you to perform arbitrary transformation in any component.

https://github.com/nvim-lualine/lualine.nvim?tab=readme-ov-file#general-component-options

simonmandlik:
Yes, you are right. Should have noticed that 😅

PR #1344, opened 2024-12-08
Fix filetype for Trouble extension

markjaquith:
Trouble creates buffers with a file type of `trouble` (lowercase) not `Trouble`.

This PR fixes the Trouble extension to use the lowercase "trouble" filetype.

See: https://github.com/folke/trouble.nvim/blob/46cf952fc115f4c2b98d4e208ed1e2dce08c9bf6/lua/trouble/view/window.lua#L97

Also I verified it myself with `:set filetype?`
CleanShot 2024-12-08 at 01 30 58@2x

### DIFF
--- a/lua/lualine/extensions/trouble.lua
+++ b/lua/lualine/extensions/trouble.lua
@@ -17,6 +17,6 @@ M.sections = {
   },
 }
 
-M.filetypes = { 'Trouble' }
+M.filetypes = { 'trouble' }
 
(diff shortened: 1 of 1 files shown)


xieyonn:
Change filetype is not enough, See https://github.com/nvim-lualine/lualine.nvim/pull/1383

shadmansaleh:
resolved in #1383

PR #1366, opened 2025-01-14
Fix automatic refresh variable name

matesssaks:
When using global status line and "ignore_focus" option together in some cases status line disappear. Can be replicated with "globalstatus = true" and "ignore_focus = { "lazy", "notify" }" and focusing floating window of Lazy or Notify.

Problem is global status line needs refresh scope "window".  Refresh timers set variable named "kind", but refresh function expects variable "scope", becase it's not set it fallback to scope "tabpage".

Fixed by renaming "kind" to "scope" in refresh timers.


shadmansaleh:
lgtm. thanks. I think I missed it at some point during refactor

PR #1368, opened 2025-01-17
fix: insert r_padding within final highlight

danielwe:
When a component ends with a custom highlight, any `r_padding` should be added inside that highlight, not after, just like how `l_padding` is added inside a custom starting highlight. This PR takes care of that.

This is required to fully address https://github.com/folke/trouble.nvim/issues/569 (see also https://github.com/folke/trouble.nvim/pull/616 for the fix to trouble.nvim's own part in the issue).

**Before:**
Screen Shot 2025-01-16 at 16 53 11 PM

**After:**
Screen Shot 2025-01-16 at 17 18 20 PM

### DIFF
--- a/lua/lualine/component.lua
+++ b/lua/lualine/component.lua
@@ -105,7 +105,13 @@ function M:apply_padding()
     end
   end
   if r_padding then
-    self.status = self.status .. string.rep(' ', r_padding)
+    if self.status:reverse():find('%*%%.*#.*#%%') == 1 then
+      -- When component has changed the highlight at the end
+      -- we will add the padding before the highlight terminates
+      self.status = self.status:sub(1, -3) .. string.rep(' ', r_padding) .. self.status:sub(-2, -1)
+    else
+      self.status = self.status .. string.rep(' ', r_padding)
+    end
   end
 end
 


lua/lualine/component.lua
```diff
@@ -105,7 +105,13 @@ function M:apply_padding()
     end
   end
   if r_padding then
-    self.status = self.status .. string.rep(' ', r_padding)
+    if self.status:reverse():find('%*%%.*#.*#%%') == 1 then
+      -- When component has changed the highlight at the end
+      -- we will add the padding before the highlight terminates
+      self.status = self.status:sub(1, -3) .. string.rep(' ', r_padding) .. self.status:sub(-2, -1)
```
shadmansaleh, line 111: Explain this line please.  Why are you removing only last 2 chars and not the HL group mapping %#group_name# ?

danielwe: This line inserts padding at the _end_ of the `<text>` part of a string of the form `%#Group#<text>%*`, so the only characters to hold out are the last two characters `%*`.

danielwe:
There isn't a separate HL marker at the end, it's just that if a component is wrapped in a single HL marker, `%#Group#<text>%*` and I want to add padding, I should add the padding inside the marker such that it becomes part of the `<text>` component. Currently, this is done for left padding but not right padding. With this PR, left padding and right padding are handled symmetrically.

danielwe:
See the screenshots in the OP: this PR addresses the miscolored space after `view:wait`

Issue #1375, opened 2025-02-12
Feat: provide simple `lsp_status` component


andreihh:
### Requested feature

Lualine should provide a simple `lsp_status` component, which displays the list of active LSPs in the current buffer, and a simple spinner / done symbol to indicate that the LSP is working.

Example screenshots:
![lualine-bashls-no-progress](https://github.com/user-attachments/assets/da44e21a-2e1f-49dd-90aa-6c38f13c2c76)
![lualine-lua_ls-done](https://github.com/user-attachments/assets/16880b41-0577-4d9b-9722-67617f6bce87)
![lualine-lua_ls-progress](https://github.com/user-attachments/assets/e899aeb0-5093-4414-8262-82f61b6fa4b1)

### Motivation

There are various plugins that display verbose LSP status and progress in the status line:
- https://github.com/nvim-lua/lsp-status.nvim (deprecated, unmaintained)
- https://github.com/arkav/lualine-lsp-progress (deprecated, unmaintained)
- https://github.com/linrongbin16/lsp-progress.nvim

The only active plugin, `linrongbin16/lsp-progress.nvim`, displays overly-verbose information that is almost guaranteed to not fit on the status line. Configuration is very clunky, having to effectively rewrite the entire format function just to skip a component.

It's very useful to have the list of active LSPs (most oftenly just one), with a spinner to indicate active work / progress, but without the verbose LSP messages. No plugins provide this out-of-the-box, and configuration to achieve this is overly-complicated.

In line with the LSP diagnostics, Git branch, and other helpful, simple components, Lualine should provide an `lsp_status` component.

Edit: just discovered that https://github.com/echasnovski/mini.statusline supports an LSP status component, which lists the active LSPs (same as this PR but without the progress spinner). So this PR brings feature parity to Lualine.

PR #1376, opened 2025-02-12
feat: add `lsp_status` component to display active LSPs and progress

andreihh:
Closes https://github.com/nvim-lualine/lualine.nvim/issues/1375.

Following an example of a simple LSP progress using notifications: https://github.com/folke/snacks.nvim/blob/main/docs/notifier.md#-examples, we implement a simple `lsp_status` component that:
- Defines an `autocmd` on `LspProgress` to update the spinner state and refreshes Lualine if the state changed
- Displays the list of active LSPs with the corresponding spinner state if supported by the LSP

Not all LSPs support progress (e.g., `bashls` does not). In that case, we simply skip the spinner and display only the LSP name.

LSPs are separated by a configurable delimiter. The progress state symbols are also configurable.

Screenshots:
lualine-bashls-no-progress
lualine-lua_ls-done
lualine-lua_ls-progress

NOTE: I updated the GitHub actions to use Ubuntu 22.04. The 20.04 image is deprecated (see https://github.com/actions/runner-images/issues/11101), and `nvim-nightly` doesn't support it anymore (building `nvim` fails because the `glibc` version is too old).

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,7 @@ sections = {lualine_a = {'mode'}}
 - `selectioncount` (number of selected characters or lines)
 - `tabs` (shows currently available tabs)
 - `windows` (shows currently available windows)
+- `lsp_status` (shows active LSPs in the current buffer and a progress spinner)
 
 #### Custom components
 
@@ -788,6 +789,29 @@ sections = {
         active = 'lualine_{section}_normal',     -- Color for active window.
         inactive = 'lualine_{section}_inactive', -- Color for inactive window.
       },
+    }
+  }
+}
+```
+
+#### lsp status component options
+
+```lua
+sections = {
+  lualine_a = {
+    {
+      'lsp_status',
+      icon = '', -- f013
+      symbols = {
+        -- Standard unicode symbols to cycle through for LSP progress:
+        spinner = { '⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏' },
+        -- Standard unicode symbol for when LSP is done:
+        done = '✓',
+        -- Delimiter inserted between LSP names:
+        separator = ' ',
+      },
+      -- List of LSP names to ignore (e.g., `null-ls`):
+      ignore_lsp = {},
     }
   }
 }
--- a/lua/lualine/components/lsp_status.lua
+++ b/lua/lualine/components/lsp_status.lua
@@ -1,0 +1,103 @@
+local lualine_require = require('lualine_require')
+
+local M = lualine_require.require('lualine.component'):extend()
+
+local default_options = {
+  icon = '', -- f013
+  symbols = {
+    -- Use standard unicode characters for the spinner and done symbols:
+    spinner = { '⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏' },
+    done = '✓',
+    separator = ' ',
+  },
+  -- List of LSP names to ignore (e.g., `null-ls`):
+  ignore_lsp = {},
+}
+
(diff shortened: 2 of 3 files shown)


shadmansaleh:
This is interesting. I'm in favor of adding it.

Can you add a option to ignore a list of lsp servers. Usually I only want to see relevent lsp status for the file. Things like null-ls attaches to everything. It'd be nice to have a way not to show them in statusline.

andreihh:
That's a great idea! I'll add an option for it.

andreihh:
Done! I added an `ignore_lsp` list option.

shadmansaleh (approved):

shadmansaleh:
lgtm. Thanks

andreihh:
Thanks for merging this! I also prepared https://github.com/nvim-lualine/lualine.nvim/pull/1391 - I realized that `vim.loop.hrtime` is available on older nvim versions, so I simplified the spinner logic and tests a bit. There is no change in functionality, only a bit of simplification around backwards-compatibility.

petobens:
Hi @andreihh! Is it possible to add a config option to hide displaying the icon/name once the lsp is done/fully-commenced? i.e want to see a lsp status while the lsp server is actually loading but after that I want it to get out of the way

andreihh:
Hi! You can remove the icon by setting `symbols.done = ""` (override it to the empty string).

lpiepiora:
> Hi! You can remove the icon by setting `symbols.done = ""` (override it to the empty string).

Hi, thanks for the feature, I really like it. I was also looking for removing done & spinner, because I use another plugin for that. I'd still like to use this feature to see LSPs attached to the buffer though.

There is one weird behavior when setting `done` to an empty string ;). When the done symbol is empty, then an extra space is added after the LSP is done. Some LSPs won't show that space first, but will only show it after they are done. 

Screenshot 2025-03-16 at 06 31 30

Also they will add an extra space if they are last in the list (like the `null-ls` in the screenshot below, it first has one space after the `null-ls` word, and when I save the file, there are two).

Screenshot 2025-03-16 at 06 32 59

This makes the attached LSPs float (meaning they shift when done). I think maybe there should be a check, that checks if the set symbol is not empty.

PS. I could commit a fix, if this is something you don't have time for right now.

petobens:
> Hi! You can remove the icon by setting symbols.done = "" (override it to the empty string).

What I meant is that once the lsp loads I want to remove the segment all together. For instance in the GIF below I do want to see the spinner while lua_ls is loading but once if finishes I want the whoe `<icon> lua_ls` segment to disappear i.e I only want to have an indicator in the status of the lsp loading. Is that possible @andreihh 

Peek 2025-03-16 08-20

andreihh:
@petobens I understand what you mean now. I could add an `ignore_idle` option, similar to the `ignore_lsp`, but this would break with LSPs that don't support progress (e.g., `clangd`): the LSP would never show up in the statusline, so you would have no idea if it even attached. It defeats the purpose of this component, so I'd be reluctant to add this option.

@lpiepiora That's a good point, I'll piggy-back on https://github.com/nvim-lualine/lualine.nvim/pull/1391 to add the spaces only if the corresponding symbols are non-empty (or will submit a new pull request if that one gets merged in the meantime).

andreihh:
@lpiepiora Should be fixed in https://github.com/nvim-lualine/lualine.nvim/pull/1391/commits/1d8aed7b92ff342de14e73db2e58a6004262c947. Could you try using my fork and let me know if it works for you?

lpiepiora:
> @lpiepiora Should be fixed in [1d8aed7](https://github.com/nvim-lualine/lualine.nvim/commit/1d8aed7b92ff342de14e73db2e58a6004262c947). Could you try using my fork and let me know if it works for you?

@andreihh thanks, that's much better (at least to my liking ;))

PR #1383, opened 2025-02-20
fix: trouble extension. Works on Trouble.nvim v3 and keep backward compatibility.

xieyonn:
Changes in v3 of [Trouble.nvim](https://github.com/folke/trouble.nvim):
- Filetype changed, `Trouble`  to `trouble`.
- The current window's mode information is stored in the variable: `vim.w[win_id].trouble`.

preview:
image

### DIFF
--- a/lua/lualine/extensions/trouble.lua
+++ b/lua/lualine/extensions/trouble.lua
@@ -1,14 +1,32 @@
 local M = {}
 
-local function get_trouble_mode()
-  local opts = require('trouble.config').options
-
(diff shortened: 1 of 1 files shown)


xieyonn:
Action test case nvim-nightly failed,  because the runner image in action.yaml needs to be updated.

Since someone has already submitted an update, see https://github.com/nvim-lualine/lualine.nvim/pull/1376, so i just leave it.

shadmansaleh (approved):

PR #1394, opened 2025-03-19
feat: allow `ignore_focus` to be set to a `win -> bool` function

noib3:
Use case: I'd like to ignore all windows whose buffer's `buftype` is not empty.

### DIFF
--- a/lua/lualine.lua
+++ b/lua/lualine.lua
@@ -312,6 +312,25 @@ local function status_dispatch(sec_name)
   end
 end
 
+---Determines if a focus event for this window should be ignored.
+---
+---@param user_config table The user's config
+---@param win number The Neovim window handle to check
+---@return boolean Whether focus events for this window should be ignored
+local function should_ignore_focus(user_config, win)
+  local ignore_focus = user_config.options.ignore_focus
+  if type(ignore_focus) == 'table' then
+    -- ignore focus on filetypes listed in options.ignore_focus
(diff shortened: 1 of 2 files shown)


lua/lualine.lua
```diff
@@ -312,6 +312,25 @@ local function status_dispatch(sec_name)
...
+---@return boolean Whether focus events for this window should be ignored
+local function should_ignore_focus(user_config, win)
+  local ignore_focus = user_config.options.ignore_focus
+  if type(ignore_focus) == 'table' then
+    -- ignore focus on filetypes listed in options.ignore_focus
+    local buf = vim.api.nvim_win_get_buf(win)
+    local filetype = vim.api.nvim_buf_get_option(buf, 'filetype')
+    return vim.tbl_contains(ignore_focus, filetype)
+  elseif type(ignore_focus) == 'function' then
+    return ignore_focus(win)
```
shadmansaleh, line 328: Use `nvim_win_call` to call the function in windows context instead of passing it. It's more intuitive for the user that way. And the function just can think about current window.

noib3: https://github.com/nvim-lualine/lualine.nvim/pull/1394/commits/af97df4f5c19ab6047b0576c268beac06d8a6513#diff-921f7a7fb67331a59c9f0c0b903088920375643e4a496e60283145818e7a23dbR328

PR #1409, opened 2025-04-22
Add theme support for tinted-nvim

JamyGolden:
This PR adds support for [tinted-nvim](https://github.com/tinted-theming/tinted-nvim) using the `base16` theme value.

### DIFF
--- a/THEMES.md
+++ b/THEMES.md
@@ -342,6 +342,18 @@ pywal is another special theme. It will load the colors from your current [pywal
 <img width='700' src='https://user-images.githubusercontent.com/41551030/108649258-48907600-74bd-11eb-9bba-8e82b56777d9.png'/>
 </p>
 
+### tinted
+
+This theme will automatically use colors defined by your colorscheme using [tinted-theming/tinted-nvim](https://github.com/tinted-theming/tinted-nvim) or [tinted-theming/tinted-vim](https://github.com/tinted-theming/tinted-vim) plugin.
+The following example is using the `ayu-dark` colorscheme:
+
+<p>
+<img width='700' src='https://user-images.githubusercontent.com/41551030/108648314-357ca680-74bb-11eb-83de-ec94c3a75d99.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/41551030/108648329-3e6d7800-74bb-11eb-901b-41e1f01315f0.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/41551030/108648342-44fbef80-74bb-11eb-8262-a6f615e5fcfb.png'/>
+<img width='700' src='https://user-images.githubusercontent.com/41551030/108648367-52b17500-74bb-11eb-8ec5-dbb59db60d98.png'/>
+</p>
+
 ### Tomorrow
 
 <p>


lua/lualine/themes/base16.lua
```diff
@@ -90,6 +90,34 @@ local function setup_base16_nvim()
   }
 end
 
+local function setup_tinted_nvim()
+  -- Continue to load tinted-nvim
+  local loaded, tinted = pcall(require, 'tinted-colorscheme')
+
+  if not loaded then
+    add_notice(
```
shadmansaleh, line 98: This is problematic as this will show this notice even when base16_nvim is present

shadmansaleh:
Also, you can put the theme under `/lua/lualine/themes/` in the colorschemes repository too. Even if you have the same filename, base16.lua` lualine guarantees it'll load the theme from your repo instead of from itself. That way, when users have your colorscheme installed, your theme gets loaded in lualine. And you have full control over the colors without having to wait for merging updates here.

JamyGolden:
Yeah tinted-vim is the continuation of base16-vim, where as tinted-nvim (new) is a fork of base16-nvim which has been extended. I contacted the base16-nvim maintainer about combining our efforts into 1 repo but they said they didn't have capacity so ended up being a fork.

A lot of nvim users like using a plugin which is built with lua and they tend to pick something other than tinted-vim since it's built with vimscript. We have a lot of people using tinted-vim who don't use nvim, so we ended up having both under the org with different people maintaining the different repos.

JamyGolden:
I've added a `.../themes/tinted.lua` file and added support for tinted-nvim and tinted-vim plugin and added info to the README. I've removed all changes I made to the `.../themes/base16.lua` file.

JamyGolden:
Can you link me to an example of a theme doing this? I'm not quite grasping the setup if it's in the tinted-nvim repo. How is the theme imported and used in lualine if it's external?

shadmansaleh:
> Can you link me to an example of a theme doing this?

Most of the colorschemes actually do that already. For example [onedark](https://github.com/navarasu/onedark.nvim/blob/master/lua/lualine/themes/onedark.lua) [tokyonight](https://github.com/folke/tokyonight.nvim/blob/main/lua/lualine/themes/tokyonight.lua)...

> I'm not quite grasping the setup if it's in the tinted-nvim repo. How is the theme imported and used in lualine if it's external?

You don't need to do anything special. You can just place the same file in same location in your repo. The [loading](https://github.com/nvim-lualine/lualine.nvim/blob/15884cee63a8c205334ab13ab1c891cd4d27101a/lua/lualine/utils/loader.lua#L208-L254) is handled by us. You don't need to worry about that. You basically just need to export the theme in the lualines format from file in lua/lualine/themes/theme-name.lua . You are already doing that in the file here. No need to do anything different.

JamyGolden:
Ah ok interesting! How often does lualine pull those lua theme files from the different repos? I've merged a PR on tinted-nvim: https://github.com/tinted-theming/tinted-nvim/pull/15 

I've updated this PR to just include README info for the `tinted` theme option. (I guess it should be merged after lualine pulls the theme file from tinted-nvim repo)

shadmansaleh:
> How often does lualine pull those lua theme files from the different repos?

We don't : ] 

You're still misunderstanding. Basically, we will load them from your repo at runtime. Lets say someone has both tinted-nvim and lualine installed.  And are using tinted-nvim as their colorscheme. Lualine will load that theme file from tinted-nvims installation. It doesn't need to belong in this repo. Also as long as you name the theme file same as your colorschemes name user doesn't even need to explicitly configure. It get's auto infered when default "auto" theme is used. Basically theme file exisiting in lualines repo and in any other repo is the same to lualine. It actually prioritizes external ones. Say you keep a base64.lua theme in your repo lualine will load that one instead of the one in this repo. We kind of don't even merge new themes nowadays. The ones that are currently there have historically been there.

JamyGolden:
Haha oh ok I see now, all dynamically handled, very nice! Thanks for all the info!

PR #1414, opened 2025-05-10
feat: add extension for avante

thesimonho:
Extension to add information for [avante](https://github.com/yetone/avante.nvim) windows.

Particularly useful when not using a global status line as it simplifies the display of the avante sidebar


shadmansaleh:
lgtm

Issue #1416, opened 2025-05-12
Bug: `lsp_status` shows duplicated lsp


davychhouk:
### Self Checks

- [x] I'm using the latest lualine.
- [x] I didn't find the issue in existing issues or PRs.

### How to reproduce the problem

Open any file with LSP configured.

### Expected behaviour
It should only show each lsp once only.

### Actual behaviour
It shows some lsp 2 times.

<img width="1215" alt="Image" src="https://github.com/user-attachments/assets/b42cf125-e821-4661-8dbf-195fc127aeb7" />

<img width="1498" alt="Image" src="https://github.com/user-attachments/assets/e495ad59-bfdc-45d1-a385-a217c1c0d164" />

### Minimal config to reproduce the issue

```lua
return {
	"nvim-lualine/lualine.nvim",
	dependencies = {
		"williamboman/mason.nvim",
		"nvim-tree/nvim-web-devicons",
	},
	config = function()
		require("lualine").setup({
			sections = {
				lualine_x = { "lsp_status" },
			},
		})
	end,
}
```

PR #1417, opened 2025-05-12
Fix `lsp_status` shows duplicated lsp

davychhouk:
Fixed lsp_status shows duplicated lsp by additional result table insertion check.

### DIFF
--- a/lua/lualine/components/lsp_status.lua
+++ b/lua/lualine/components/lsp_status.lua
@@ -53,6 +53,7 @@ end
 
 function M:update_status()
   local result = {}
+  local processed = {}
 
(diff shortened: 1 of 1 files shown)


davychhouk:
Fix for this issue: https://github.com/nvim-lualine/lualine.nvim/issues/1416

shadmansaleh:
This does solve it. But I don't get why api is returning duplicates.

davychhouk:
> This does solve it. But I don't get why api is returning duplicates.

I believe it has something to do with the latest nvim or lspconfig release.
I only noticed this bug recently after lazy update. It was working fine before that.

PR #1427, opened 2025-06-01
feat(mode/filename/path): add path option 5 to show git repo relative path

am-zest:
show the relative path from the git repo name down to the file name. If not in a git repo, do same as option 4.


am-zest:
sorry, submitted to wrong repo. :/

PR #1448, opened 2025-07-22
fix(avante): recognize "todo" and "code" sections

dtor:
In addition to input, output, and "selected files" windows Avante may be using a window for "todos" and a window to show a snippet of code selected by the user. Recognize them.


shadmansaleh:
lgtm

PR #1475, opened 2025-11-09
fix: use focusable as window filter for refresh()

henry-hsieh:
The original logic uses `win_gettype()` to check the window type. However, not all popup windows do not have winbar. For example, [detour.nvim](https://github.com/carbon-steel/detour.nvim) uses float windows to navigate between positions, or `zen` component of [snacks.nvim](https://github.com/folke/snacks.nvim) uses a float window as zoomed/zen window. When a windows is focusable, it may want to show the winbar.


henry-hsieh:
It seems some popups have the property of focusable. That will make winbar shown in undesired window. Working on an alternative solution.

PR #1480, opened 2025-11-24
fix: make bg color check case insensitive

danielwe:
My statusline was inexplicably missing component separators when using a transparent background. The reason turned out to be that the component had `color_fn_cache.bg = "None"`, while the section had `section_color.bg = "NONE"`. This PR makes these string comparisons case insensitive to ensure that these color specs are considered equal.

### DIFF
--- a/lua/lualine/utils/section.lua
+++ b/lua/lualine/utils/section.lua
@@ -62,14 +62,14 @@ function M.draw_section(section, section_name, is_focused)
       (
         type(section[component_no].options.color) == 'table'
         and section[component_no].options.color.bg
-        and section[component_no].options.color.bg ~= section_color.bg
+        and string.upper(section[component_no].options.color.bg) ~= string.upper(section_color.bg)
       )
       or type(section[component_no].options.color) == 'string'
       or (
         type(section[component_no].options.color) == 'function'
         and section[component_no].color_fn_cache
         and section[component_no].color_fn_cache.bg
-        and section[component_no].color_fn_cache.bg ~= section_color.bg
+        and string.upper(section[component_no].color_fn_cache.bg) ~= string.upper(section_color.bg)
       )
     then
       strip_next_component = true


shadmansaleh:
lgtm

jxlambda:
When updating to 74114f0df6, my statusline ceased to render. I tracked the issue down to the fact that I was passing two background colors that were set to `nil` (which, if I recall correctly, was a workaround to some other bug previously; I think it was that following component background color did not get reset if one wasn't specified).

When specifying the color as "None" (or any variant of that), it still fails to render. Only when I set them to empty strings did they come back.

Another failure mode was when I removed specifying the background color entirely, oddly enough.

danielwe:
This change did assume that the bg field is always a string. I suppose that's not always true? I lack full context/overview here, but I can do a PR to relax that assumption.

jxlambda:
To reproduce, try the following:

```lua
local colors = {
  command      = "#d70000",
  normal       = "#424242",
  insert       = "#00af00",
  replace      = "#d75f00",
  visual       = "#ffd700",

  innerbg      = nil,
  outerbg      = nil,

  black        = "Black",
  white        = "Grey58",
  red          = "Red3",
  green        = "SpringGreen3",
  blue         = "DeepSkyBlue2",
  yellow       = "Gold1",

  darkgray     = "#070707",
  gray         = "#3a3a3a",
  lightgray    = "#424242",
  inactivegray = "#7c6f64",
}

return {
  "nvim-lualine/lualine.nvim",
  
  opts = {
    sections = {
      lualine_a = { "mode" },
      lualine_b = {
        "filename",
        {
          "filetype",
          icon_only = true,
        },
        {
          "fileformat",
          colored = true,
          symbols = {
            unix = "",
            msdos = "",
            mac = ""
          }
        },
        "encoding"
      },
      lualine_c = {
        "branch",
        "diff",
        {
          "diagnostics",

          sources = { "nvim_lsp", "nvim_diagnostic" },
          sections = { "error", "warn", "info", "hint" },
          symbols = { error = "● ", warn = "▲ ", info = "■ ", hint = "◆ " },
          colored = true,
          always_visible = false,
          update_in_insert = false,
        },
      },

      lualine_x = {
        {
          "lsp_status",
          icon = " ",
          symbols = {
            spinner = { "⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏" },
            done = "",
            separator = " ",
          },
          ignore_lsp = {},
        },
        { "searchcount", timeout = 300 },
      },
      lualine_y = {
        { "selectioncount", color = { bg = "blend" } },
      },
      lualine_z = {
        "location",
        "progress",
      }
    },

    inactive_sections = {
      lualine_a = { "mode" },
      lualine_b = { "filename" },
      lualine_c = {},

      lualine_x = {},
      lualine_y = {},
      lualine_z = { "location", "progress" },
    },

    options = {
      icons_enabled = true,
      component_separators = { left = "", right = "" },
      section_separators = { left = "", right = "" },

      ignore_focus = {},
      always_divide_middle = true,
      globalstatus = false,

      refresh = {
        statusline = 1000,
        tabline = 1000,
        winbar = 1000,
      },

      theme = {
        inactive = {
          a = { fg = colors.inactivegray, bg = colors.outerbg },
          b = { fg = colors.inactivegray, bg = colors.outerbg },
          c = { fg = colors.inactivegray, bg = colors.innerbg },
        },
        visual = {
          a = { fg = colors.darkgray, bg = colors.visual },
          b = { fg = colors.lightgray, bg = colors.outerbg },
          c = { fg = colors.lightgray, bg = colors.innerbg },
        },
        replace = {
          a = { fg = colors.darkgray, bg = colors.replace },
          b = { fg = colors.lightgray, bg = colors.outerbg },
          c = { fg = colors.lightgray, bg = colors.innerbg },
        },
        normal = {
          a = { fg = colors.darkgray, bg = colors.normal },
          b = { fg = colors.lightgray, bg = colors.outerbg },
          c = { fg = colors.lightgray, bg = colors.innerbg },
        },
        insert = {
          a = { fg = colors.darkgray, bg = colors.insert },
          b = { fg = colors.lightgray, bg = colors.outerbg },
          c = { fg = colors.lightgray, bg = colors.innerbg },
        },
        command = {
          a = { fg = colors.darkgray, bg = colors.command },
          b = { fg = colors.lightgray, bg = colors.outerbg },
          c = { fg = colors.lightgray, bg = colors.innerbg },
        },
      },
    },
  }
}
```

The intent here is that I have two colored bits on the right-most and left-most ends of the status line, with the left one displaying the current mode and the right one displaying the position in the file, with the rest of the status line having a transparent background. If I recall correctly, then I couldn't achieve that without setting the background colors to `nil`, as the color would not be cleared/reset after the end of a section, and thus would continue until another section explicitly set it to something else.

shadmansaleh:
> This change did assume that the bg field is always a string. I suppose that's not always true? I lack full context/overview here, but I can do a PR to relax that assumption.

That'd be the case. We should do the case-insensitive check only if bothsides are string otherwise do a inequality check as before. A pr would be appreciated.

PR #1486, opened 2025-12-30
fix(tabs): correctly display oil.nvim buffers in tabline

### LINKED ISSUE #1267: Bug: tab label is blank on oil
### Self Checks

- [x] I didn't find the issue in existing issues or PRs.

When 'editing' a dir with oil, the tab label is blank on the following config due to [this line](https://github.com/nvim-lualine/lualine.nvim/blob/0a5a66803c7407767b799067986b4dc3036e1983/lua/lualine/components/tabs/tab.lua#L66)

For example, the bufname `oil:///C/Users/pedro/.local/share/nvim-data/lazy/oil.nvim/doc/` is escaped to ` `.

```
tabline = {
  lualine_a = {},
  lualine_b = {},
  lualine_c = {
    {
      "tabs",
      mode = 2,
      max_length = vim.o.columns,
      use_mode_colors = false,
      tabs_color = {
        active = "lualine_a_inactive",
        inactive = "lualine_c_inactive",
      },
    },
  },
  lualine_x = {},
  lualine_y = {},
  lualine_z = {},
},
```

tea06194:
Fixes [#1267
](https://github.com/nvim-lualine/lualine.nvim/issues/1267)

oil.nvim uses oil:// URIs as buffer names. fnamemodify() does not handle URI
schemes correctly and returns an empty string, causing tabs to show [No Name].

This patch detects oil buffers explicitly and strips the oil:// prefix before
formatting the path, keeping existing tabline behavior intact.

### DIFF
--- a/lua/lualine/components/tabs/tab.lua
+++ b/lua/lualine/components/tabs/tab.lua
@@ -53,6 +53,9 @@ function Tab:label()
   elseif self.buftype == 'terminal' then
     local match = string.match(vim.split(self.file, ' ')[1], 'term:.*:(%a+)')
     return match ~= nil and match or vim.fn.fnamemodify(vim.env.SHELL, ':t')
+  elseif self.filetype == 'oil' and vim.startswith(self.file, 'oil://') then
+    local path = self.file:gsub('^oil://', '')
+    return vim.fn.fnamemodify(path, ':~')
   elseif self.file == '' then
     return '[No Name]'
   end


abeaufays:
Hello ! I believe you have missed the oil extension?
See https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/extensions/oil.lua 
You can use it like this: 
```
    {
        'nvim-lualine/lualine.nvim',
        dependencies = { 'nvim-tree/nvim-web-devicons' },
        config = function()
            require('lualine').setup {
                extensions = { 'oil' },
            }
        end,
    },
```

abeaufays:
In case you were not satisfied with the oil extension, I made a PR about it too: https://github.com/nvim-lualine/lualine.nvim/pull/1488

tea06194:
@abeaufays 
> Hello ! I believe you have missed the oil extension? See https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/extensions/oil.lua You can use it like this:
> 
> ```
>     {
>         'nvim-lualine/lualine.nvim',
>         dependencies = { 'nvim-tree/nvim-web-devicons' },
>         config = function()
>             require('lualine').setup {
>                 extensions = { 'oil' },
>             }
>         end,
>     },
> ```

Thanks for the suggestion!

This change simply extends the existing special-case handling in
`Tab:label()` (fugitive/help/terminal) to cover oil buffers, which expose
URI-based buffer names that `fnamemodify()` cannot handle directly.
This logic is part of the tabline tabs component and does not overlap with
statusline extensions.

PR #1494, opened 2026-02-16
docs: improve styling and formatting

saidelmark:
* Allow autogen pipeline to run on its configuration changes
* Bump both local and CI docgen to panvimdoc v4.0.1
* Regenerate help file

Originally I wanted to just mark all code blocks in the documentation with a language, but couldn't make the 2.7.3 version of panvimdoc work locally. So I switched to the latest, and couldn't make the generated doc look "exactly like before, but with better code blocks".

### DIFF
--- a/doc/lualine.txt
+++ b/doc/lualine.txt
@@ -3,15 +3,25 @@
 ==============================================================================
 Table of Contents                                  *lualine-table-of-contents*
 
-1. lualine.nvim                                         |lualine-lualine.nvim|
-  - Contributing                                        |lualine-contributing|
-  - Performance compared to other plugins|lualine-performance-compared-to-other-plugins|
-  - Installation                                        |lualine-installation|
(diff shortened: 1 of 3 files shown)


doc/lualine.txt
```diff
@@ -539,18 +537,16 @@ General component options              These are options that control behavior
...
-                                       specific components. For example, you
-                                       have option on `diagnostics` component
-                                       to specify what your diagnostic sources
-                                       will be.
+These are options that are available on specific components. For example, you
+have option on `diagnostics` component to specify what your diagnostic sources
+will be.
 
 
-                                           *lualine-buffers-component-options*
```
shadmansaleh, line 551: These tags are acutally quite useful when you want to find specific docs for a component. Also these are referenced in other parts of the doc. So, we need to figure something about this before merge otherwise in current state those links just remain in a dangling.

PR #1499, opened 2026-04-04
Don't assume string in bg check

danielwe:
(Hopefully) fixes issue reported in #1480

I took the liberty of moving `sanitize_color` into the utils module and reusing it here, as this seemed like the most robust and consistent approach.


danielwe:
Maybe this warrants a regression test? I need to move on but can come back later to fix that, otherwise feel free to push to this branch.

shadmansaleh:
Looks good to me and does fix the issue. Thanks for the pr.

PR #1506, opened 2026-05-06
Only render current buffer if listed

oiiiiiiii:
This change makes sure the current buffer is only rendered in the bufferline when in it actually listed, but keeps code changes minimal

### DIFF
--- a/lua/lualine/components/buffers/init.lua
+++ b/lua/lualine/components/buffers/init.lua
@@ -126,7 +126,7 @@ function M:update_status()
   if max_length == 0 then
     max_length = math.floor(2 * vim.o.columns / 3)
   end
-  local total_length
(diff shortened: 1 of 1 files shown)


shadmansaleh:
This is intended. For currentbuffer we should always show it in bufferline. When the buffer is not in focus we can hide it if it's not listed.

oiiiiiiii:
Why is this intended? Then things like telescope etc. also show up, which makes no sense to me. Anyhow, how about an option to have the other behavior if this is intended?

shadmansaleh:
> Then things like telescope etc. also show up, 

Only when you are in thet buffer

> Anyhow, how about an option to have the other behavior if this is intended?

Sure, we can add an option for this

PR #1512, opened 2026-06-04
Component that merges filetype and lsp_status components

theinfinityglitch:
I mixed the filetype and lsp_status into a single component for use cases that the lsp name is not needed and both compoents are used together.

With the normal use of both original components, the padding remains between them, and if the padding is removed, in some cases the filetype gets a weird look. These cases are for example, when the lsp is off or the lsp doesn't return any icon or symbol.

## Component examples

image

image


theinfinityglitch:
Just after I made the pull request I thought that maybe would be even better if instead of a componet that mix these both, we had a way to allow the merge between any component, like a section inside a section

